PR #1, opened 2025-02-26
AI-554: Migrate dependencies from GitHub to PyPi

cyrus2281:

The purpose of this change is to shift dependency management from using GitHub URLs to using the PyPI registry, which helps streamline the build process and improve reliability of package installations.

## How is this accomplished?

- Moved deps to PyPi from github

Ensure that the new dependencies are correctly specified and available on PyPI.


artyom-morozov (approved):
LGTM

PR #12, opened 2025-02-28
AI-558: Improve the configurability of the response_format_prompt

efunneko:

1. Allow a user to put the response_format_prompt in either the gateway or interface configuration and if it isn't there, then insert a default value.
2. Add a bit more information into the orchestrator prompt to indicate how to handle report requests

This should be backwards compatible

### DIFF
--- a/cli/commands/build.py
+++ b/cli/commands/build.py
@@ -160,7 +160,8 @@ def ensure_config_aliases(content, config_aliases):
     updated_content = content
     
     for alias, default_config in config_aliases.items():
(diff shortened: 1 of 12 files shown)


efunneko (commented):
I changed the orchestrator a bit to make the out of the box experience for users a bit nicer. Previously, I had put this into the response_format_prompt, but there were a few places where that was specified and descriptions of how reports should be returned aren't really specific to interface formatting.

cyrus2281 (approved):
LGTM

PR #15, opened 2025-03-03
AI-594: Loosen dependencies

gregmeldrum:

Loosen dependencies for kiota

### How is this accomplished?

 Loosened version requirements in the pyproject.toml file by changing strict version pins to use the "~=" operator for more flexible dependency resolution.

No


RohanSreelesh (approved):
LG!

PR #17, opened 2025-03-04
AI-597: Make postgres instance initialization configurable 

mahtabahmed255:
Make postgres instance initialization configurable
### How is this accomplished?
added a default param.


efunneko (approved):
Looks good

PR #19, opened 2025-03-04
AI-308: Handle litellm errors

Hugo-Pare:
Handling Litellm errors

### How is this accomplished?
Any messages having `handle_error` key set to True will return their payload's content to the orchestrator.

This is a multi-repository change. The change itself does not affect anything yet since message payloads do not include the `handle_error` key.


gregmeldrum (approved):
LG!

PR #21, opened 2025-03-05
AI-607: A few doc changes during the quick start

efunneko:

Previously, the only way given to interact with the system was to use the REST interface. Using the browser interface is a much more interesting experience, so this change suggests that as the first interaction.

Also updated the main arch image to the marketing team's version

### How is this accomplished?

Docs changes.


README.md
```diff
@@ -63,7 +63,11 @@ To get started with Solace Agent Mesh, follow these steps:
 
    _(Use `-eb` to combine build and run steps.)_
 
-5. **Send a Request**:
+5. **Connect to the Web Interface**:
+
+   Open the web interface at [http://localhost:5050](http://localhost:5001) or with the port specified during `init`.
```
gregmeldrum: link text and hyperlink don't match both should use port 5001

gregmeldrum (approved):
Looks good! One minor comment

README.md
efunneko: Fixed!

PR #22, opened 2025-03-06
AI-548: Fix orchestrator asking for email bug

RohanSreelesh:
Fix the issue with user information being "unknown" after opening an agent. This will fix the issues with SAM asking for the users email for tasks such as looking up jira issues etc

### How is this accomplished?
Instead of getting user properties from the data object we get it from the message object which always has it


gregmeldrum (approved):
Looks good!

PR #23, opened 2025-03-07
AI-559: Extend request expiry and lower activity timeout

Hugo-Pare:
Permit a longer duration for streaming messages back to the orchestrator.
Raising Timeout Error if nothing is streamed back within a configurable limit (activity_timeout_s)

### How is this accomplished?
New config added in AI-559 (see solace-ai-connector repo) - catch orchestrator timeouts faster
Extended `request_expiry_ms` value to 5 min, will allow generation of larger files

DO NOT MERGE before the AI-559 branch is merged on solace-ai-connector

### DIFF
--- a/configs/orchestrator.yaml
+++ b/configs/orchestrator.yaml
@@ -85,7 +85,8 @@ flows:
         broker_request_response:
           enabled: true
           broker_config: *broker_connection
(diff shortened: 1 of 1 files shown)


Hugo-Pare:
Changes made to AI-559, no longer need this.

PR #25, opened 2025-03-09
Docs: Fix Gateway Sequence Diagram

TamimiGitHub:

The Gateway sequence diagram has a duplicated Solace Agent Mesh column

### Current Behaviour

https://solacelabs.github.io/solace-agent-mesh/docs/documentation/concepts/gateways

Screenshot 2025-03-09 at 2 51 54 PM

### Expected Behaviour

Screenshot 2025-03-09 at 2 54 56 PM


efunneko (approved):
Thanks Tamimi

PR #26, opened 2025-03-10
AI-548: Better implementation for unknown email bug

RohanSreelesh:
Use a cleaner implementation to not expose unneeded information.

### How is this accomplished?
Instead of using payload we get the user_info from the user_properties dictionary, avoiding passing the whole user_properties in.


cyrus2281 (approved):
LGTM

PR #29, opened 2025-03-11
AI-557: Improve pip and build process by fixing minor details

cyrus2281:

The purpose of this change is to address and fix minor issues related to package installation tools and configurations, ensuring consistency and correctness in the build and plugin commands.

## How is this accomplished?

- The `pip` references were updated to use `pip3` to ensure compatibility with Python 3.
- The installation options for plugins were refined by replacing `uv-pip` with `uv`, simplifying the command interfaces.
- The `Makefile` was updated to use `hatch build` instead of `python3 -m build` for building the package.

Reviewers should focus on ensuring that the changes do not disrupt the existing workflow and that the new commands properly execute the intended actions without errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ build-web-visualizer:
 		npm run build
 
 build-pypi: build-web-visualizer
-	@python3 -m build
+	@hatch build
 
 build: test build-pypi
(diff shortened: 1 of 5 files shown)


gregmeldrum (approved):
LGMT!

PR #30, opened 2025-03-11
AI-612: Simplified init command

RohanSreelesh:
Simplify the init command by removing asking for directories, rest api and web ui. These will now just use the defaults from the options dictionary.

### How is this accomplished?
By changing the init command

The init command should be working as expected


cyrus2281 (approved):
LGTM

PR #31, opened 2025-03-11
AI-615: Fix rare CORS issue due to using localhost 

RohanSreelesh:
Instead of serving web ui on localhost, we use 127.0.0.1, in rare cases we saw some CORS issues due to this.


gregmeldrum (approved):
LG!

PR #37, opened 2025-03-12
AI-559: Handle timeout errors at sam level 

Hugo-Pare:
Catch Timeout errors when handling stream

### How is this accomplished?
Adding an exception for `do_broker_request_response` and streaming the last chunk


cyrus2281 (approved):
LGTM

PR #39, opened 2025-03-14
AI-643: Make examples llm agnostic

RohanSreelesh:
Make examples llm agnostic, so that in the future they can be easily rendered in another format for example markdown for open ai. 

### How is this accomplished?
Before passing in the exampls we convert them based on the given llm_type.

### DIFF
--- a/src/orchestrator/orchestrator_prompt.py
+++ b/src/orchestrator/orchestrator_prompt.py
@@ -414,7 +414,7 @@ you should ask the originator for more information. Include links to the source 
 """
 
 
-def format_examples_by_llm_type( examples: list, llm_type: str = "anthropic") -> list:
+def format_examples_by_llm_type(examples: list, llm_type: str = "anthropic") -> list:
     """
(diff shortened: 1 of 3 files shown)


src/orchestrator/orchestrator_prompt.py
```diff
@@ -413,3 +411,101 @@ def ContextQueryPrompt(query: str, context: str) -> HumanMessage:
...
+    Returns:
+        list: List of examples formatted for the specified LLM
+    """
+    formatted_examples = []
+
+    if llm_type == "anthropic":
+        for example in examples:
+            formatted_example = format_example_for_anthropic(example)
+            formatted_examples.append(formatted_example)
+
```
gregmeldrum, line 436: To be defensive we should have an else that outputs an unsupported llm type error message.

gregmeldrum (approved):
Looks good, please see the comment on adding an else.

cyrus2281 (approved):

PR #40, opened 2025-03-14
Update documentation for web UI to use 127.0.0.1 instead of localhost

RohanSreelesh:
This is being done because we recently switched the Web UI host to 127.0.0.1 from localhost, this was done because in rare cases there would be CORS errors. This is the documentation update for that change. Reference: #31


gregmeldrum (approved):
LG!

cyrus2281 (approved):

PR #41, opened 2025-03-14
AI-645: Make sure the request and response are for the same stimulus

gregmeldrum:

This change adds validation to ensure that responses from the LLM service match their corresponding requests, preventing mismatched processing of request/response pairs in the streaming workflow.

## How is this accomplished?

The implementation adds a correlation check between request and response messages using their "stimulus_uuid" property. If they don't match, the code raises an error instead of processing potentially misaligned content. Helper methods were added to extract user properties and perform correlation checks.

Reviewers should verify that the error handling is appropriate when mismatched messages are detected and ensure this change won't interrupt legitimate request flows.


cyrus2281 (approved):
LG

PR #43, opened 2025-03-17
AI-654: Make the number of planning llms and embedding models configurable.

gregmeldrum:

This pull request updates the version number to match main and makes the number of service instances configurable through environment variables, enabling easier scaling without code changes.

## How is this accomplished?

1. Bumped the patch version from 0.1.2 to 0.1.3 (to match the version in main)
2. Made the number of instances configurable by replacing hardcoded values with environment variables in service_embedding.yaml and service_llm.yaml, with fallbacks to the original values

The default values remain unchanged (1 instance), so this change shouldn't affect existing deployments unless the new environment variables are explicitly set.


cli/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "0.1.2"
+__version__ = "0.1.3"
```
gregmeldrum, line 1: To match main

cyrus2281 (approved):

PR #44, opened 2025-03-17
AI-646: Fixing small bugs raised from feedback or status channel

Hugo-Pare:
Make Solace Chat output more resilient

### How is this accomplished?
Handling edge cases
Decoding byte files

### DIFF
--- a/src/services/file_service/file_service.py
+++ b/src/services/file_service/file_service.py
@@ -392,8 +392,6 @@ class FileService(AutoExpiry, metaclass=AutoExpirySingletonMeta):
             url = url[5:-6].strip()
         if url.endswith('"') or url.endswith("'") or url.endswith(","):
             url = url[:-1]
-        if url.endswith(">"):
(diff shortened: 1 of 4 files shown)


src/services/file_service/file_service.py
```diff
@@ -392,6 +392,8 @@ def _clean_url(url: str) -> str:
             url = url[5:-6].strip()
         if url.endswith('"') or url.endswith("'") or url.endswith(","):
             url = url[:-1]
+        if url.endswith(">"):
+            url = url[:-1]
```
Hugo-Pare, line 396: A few other occurrences of URLs not being cleaned properly. Will need to add more exceptions or to find a more consistent way to clean URLs.

cyrus2281: [Q] A query parameter can have that, what's the reason for this?

cyrus2281 (approved):

PR #52, opened 2025-03-25
AI-674: Streaming Response cut due to wrong formatting 

Hugo-Pare:
- Fixing the streaming output being cut due to wrong formatting
- Uploading images with correct encoding outside of HTML files
- Combining long HTML reports correctly

### How is this accomplished?
- Adjusting to file tags not including any tag prefixes
- Giving an explicit example of how to combine multiple HTML files into one
- Modifying the Regex to escape the last '>' character in a URL

PR includes changes for AI-674 & AI-547

### DIFF
--- a/src/common/utils.py
+++ b/src/common/utils.py
@@ -332,9 +332,7 @@ def parse_orchestrator_response(response, last_chunk=False, tag_prefix=""):
             if id_match:
                 parsed_data["current_subject_starting_id"] = id_match.group(1)
 
-        elif f"<{tp}file" in line or "<file" in line:
-            if "<file" in line:
-                tp = ""
+        elif f"<{tp}file" in line:
             in_file = True
             # We can't guarantee the order of the attributes, so we need to parse them separately
             name_match = re.search(r'name\s*=\s*[\'"]([^\'"]+)[\'"]', line)
@@ -347,7 +345,7 @@ def parse_orchestrator_response(response, last_chunk=False, tag_prefix=""):
             }
             file_start_index = line.index(f"<{tp}file")
             file_line = line[file_start_index:]
-            if f"</{tp}file>" in line or "</file>" in line:
+            if f"</{tp}file>" in line:
                 file_end_index = line.index(f"</{tp}file>")
                 file_line = line[: file_end_index + len(f"</{tp}file>")]
                 file_content = [file_line]
@@ -359,9 +357,7 @@ def parse_orchestrator_response(response, last_chunk=False, tag_prefix=""):
             else:
                 file_content.append(file_line)
 
-        elif f"</{tp}file>" in line or "</file>" in line:
-            if "</file>" in line:
-                tp = ""
+        elif f"</{tp}file>" in line:
             if in_file:
                 if current_text:
                     add_content_entry(
--- a/src/common/utils.py
+++ b/src/common/utils.py
@@ -332,7 +332,9 @@ def parse_orchestrator_response(response, last_chunk=False, tag_prefix=""):
             if id_match:
                 parsed_data["current_subject_starting_id"] = id_match.group(1)
 
-        elif f"<{tp}file" in line:
+        elif f"<{tp}file" in line or "<file" in line:
+            if "<file" in line:
+                tp = ""
             in_file = True
             # We can't guarantee the order of the attributes, so we need to parse them separately
             name_match = re.search(r'name\s*=\s*[\'"]([^\'"]+)[\'"]', line)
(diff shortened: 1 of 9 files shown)


efunneko (changes requested):
I am a bit concerned with just looking for '<file' in the response. It is possible that the orchestrator might generate HTML files with '<file...' in them and this would trigger on it.

Can you look at the SAM repo and look for anywhere in the prompt where we might be telling the orchestrator that it should respond with files in <file> tags? This might have snuck into the interface or gateway configuration and it may be confusing the orchestrator

src/common/utils.py
```diff
@@ -332,7 +332,9 @@ def parse_orchestrator_response(response, last_chunk=False, tag_prefix=""):
             if id_match:
                 parsed_data["current_subject_starting_id"] = id_match.group(1)
 
-        elif f"<{tp}file" in line:
+        elif f"<{tp}file" in line or "<file" in line:
+            if "<file" in line:
```
efunneko: This concerns me that we might trigger on legitimate <file> tags in generated content

src/services/file_service/file_service_constants.py
```diff
@@ -40,7 +40,7 @@
 Keys to be treated as tags in the file block, and not file attributes.
 """
 
-FS_URL_REGEX = r"""<url>\s*({protocol}:\/\/.+?)\s*<\/url>|{protocol}:\/\/[^\s\?]+(\s|$)|{protocol}:\/\/[^\n\?]+\?.*?(?=\n|$)|\"({protocol}:\/\/[^\"]+?)(\"|\n)|'({protocol}:\/\/[^']+?)('|\n)""".format(
+FS_URL_REGEX = r"""<url>\s*({protocol}:\/\/.+?)\s*<\/url>|{protocol}:\/\/[^\s\?]+(\s|$)|{protocol}:\/\/[^\n\?]+\?.*?(?=\n|$|>)|\"({protocol}:\/\/[^\"]+?)(\"|\n)|'({protocol}:\/\/[^']+?)('|\n)""".format(
```
cyrus2281, line 43: You should not check only for a closing tag, it should be open and close both present

Hugo-Pare: The opening tag is already being properly escaped. This regex change is an additional fail-safe measure because the `<amfs[...]>` bug should be resolved by the updated prompt.

Screenshot of the original issue:
extra char at end of url

src/common/utils.py
Hugo-Pare: I have found another solution: modifying the `reponse_format_prompt` (in configs) to explicitly mention a tag prefix when returning a file or image.

src/services/file_service/file_service_constants.py
cyrus2281: I'm just worried it would cause other issues

But it probably will be okay, I'm just looking at cases where we would have html tag in the url

But in this example
```
amfs://47232d19-33d0-4505-acb7-51d6eed0611f_results_506429.csv?resolve=true&jq=.[] | ["<tr><td>", .name, "</td><td>", .status, "</td></tr>"] | join("")
```

The URL is not ending in html tags, so we should be okay

templates/web-default-config.yaml
```diff
@@ -4,4 +4,5 @@
     max_file_size: 2000 # 2GB
 
 - response_format_prompt: &response_format_prompt >
-    Return all responses in markdown format and if the response contains a file or image, return it with the <file> tags and not as a link.
+    Return all responses in markdown format and if the response contains a file or image, use the <{tag_prefix}file> tags with the assigned tag prefix.
```
cyrus2281: Is `{tag_prefix}` replaced before being sent to the LLM?

cyrus2281 (approved):
Looks Good.
Added on questions

templates/rest-api-default-config.yaml
```diff
@@ -24,5 +24,6 @@
     local_dev : ${WEBUI_LOCAL_DEV}
 
 - response_format_prompt: &response_format_prompt >
-    Return all responses in markdown format and if the response contains a file or image, return it with the <file> tags and not as a link.
+    Return all responses in markdown format and if the response contains a file or image, use the <{tag_prefix}file> tags with the assigned tag prefix.
+    Do not return files as links or with generic <file> tags.
```
cyrus2281, line 28: Further pre-processing is required to replace this with actual tag

cyrus2281 (changes requested):

cyrus2281 (approved):
LGTM

PR #53, opened 2025-03-25
AI-681: Handle timeout in actions

Hugo-Pare:
Handle Timeout errors in two actions: plantuml and web_request

### How is this accomplished?
Catching properly the errors


cyrus2281 (approved):
LGTM

PR #55, opened 2025-03-26
DATAGO-98396: relax min python requirement to 3.10.16

artyom-morozov:
- set min and max versions correctly 3.10.16 and 3.13


cyrus2281 (approved):
LGTM

PR #58, opened 2025-03-31
AI-694: Updated the history to include the agents called to generate a response

cyrus2281:

This change aims to enhance the history logging functionality by including detailed information about the agents and actions involved in generating a response. This provides more transparency and context to users and developers about the interactions that occur in the backend.

## How is this accomplished?

- Updated the history to include the agents called to generate a response
- Modified configuration to enforce alternate message roles in the history service
- Enhanced the logic to append agent actions metadata to history entries
- Adjusted tests to support the changes in history handling

Reviewers should ensure that the changes do not break existing functionality and check if the new metadata integration is functioning as expected.


cyrus2281:
I will need to do further testing to ensure the system will not reply with `message_metadata`

gregmeldrum (approved):
LG!

PR #60, opened 2025-04-02
AI-541: Enhance gateway configuration with customizable YAML 

cyrus2281:

The purpose of this change is to enhance the gateway configuration capabilities. By allowing customization of the gateway.yaml and defining gateway configurations for plugins, it improves flexibility for users to tailor integrations according to specific needs and expand the historical logging functionality. Additionally, updating the history to include the agents called to generate a response provides better tracking and transparency.

## How is this accomplished?

- Added customizing gateway.yaml
- Allowed defining gateway config for plugins
- Updated history to include the agents called for generating responses

Ensure the customized configurations for gateways and plugins meet expected function and integration requirements.

### DIFF
--- a/cli/commands/add/gateway.py
+++ b/cli/commands/add/gateway.py
@@ -169,7 +169,8 @@ def add_interface_command(name):
     Creates a new gateway interface with the provided name.
     """
     config = Config.get_plugin_config()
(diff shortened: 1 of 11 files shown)


cyrus2281:
Need to recreate PR off `dev`

PR #62, opened 2025-04-03
AI-660: Agent registry must support expiration and config updates

efunneko:

1. When agent registration messages have not be received for 60s, the agent must be removed from the list of available agents
2. If an agent of the same name registers with different configuration, then the new config must be used. This will happen on upgrades and during dev work

### How is this accomplished?

1. The agent age out method and tracking all already existed. We just needed to hook a call into it in a place where it would be called periodically. This was done in the action response timer handler
2. The orchestrator config file had a very long timeout (60k seconds). This was changed to 60s, which means 4 registrations would need to be missed before removing the agent
3. Previously, we would not update agent configuration when registering the agent. Now we always take the config from the last registration, which will allow agents to be upgraded and changed

### DIFF
--- a/configs/orchestrator.yaml
+++ b/configs/orchestrator.yaml
@@ -50,7 +50,7 @@ flows:
         component_base_path: .
         component_module: src.orchestrator.components.orchestrator_register_component
         component_config:
-          agent_ttl_ms: 60000000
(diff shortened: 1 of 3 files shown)


cyrus2281 (approved):
LGTM

PR #67, opened 2025-04-04
AI-663: Add ability to overwrite env vars in .env during sam init

gregmeldrum:

This PR adds the capability to overwrite environment variables in .env files during the sam init command

## How is this accomplished?

1. Added capability to overwrite environment variables in .env files to prevent duplicating environment variables


cyrus2281 (approved):
LGTM

PR #75, opened 2025-04-04
AI-688: Make embedding service optional

gregmeldrum:

This PR makes the embedding service optional rather than mandatory. It allows users to disable the embedding service if they don't need vector embeddings, simplifying setup and reducing external dependencies.

## How is this accomplished?

- Added a new configuration option `embedding_service_enabled` that defaults to `False`
- Modified the initialization process to only ask for embedding configuration if the service is enabled
- Updated the build process to skip embedding service configuration when disabled
- Added comprehensive tests to ensure the feature works correctly

- Backward compatibility is maintained - existing configurations will continue to work
- The embedding service is now disabled by default in the template configuration
- We've added unit tests to verify the behavior when the embedding service is enabled or disabled


cli/commands/init/ai_provider_step.py
```diff
@@ -39,38 +39,55 @@ def ai_provider_step(options, default_options, none_interactive, abort):
         default_options["llm_model_name"],
         none_interactive,
     )
-
-    ask_if_not_provided(
+    # Ask if the user wants to enable the embedding service
```
gregmeldrum, line 42: Adding the ability to disable the embedding service in the init step.

Alternatively, we could leave always omit the embedding service and make the user manually add it if they wanted to.

cli/commands/init/create_config_file_step.py
```diff
@@ -21,6 +21,14 @@ def create_config_file_step(options, default_options, none_interactive, abort):
     builtin_agents = options.get("built_in_agent", [])
     for agent in builtin_agents_ref:
         agent["enabled"] = agent["name"] in builtin_agents or agent["name"] == "global"
+        
+    # Set up the built-in services
+    builtin_services_ref = (
```
gregmeldrum, line 26: Adding a new section in the `solace-agent-mesh.yaml` file under "built_in" called "services". Here you can enable / disable the embedding service:

```
solace_agent_mesh:
  built_in:
    services:
    - name: embedding
      enabled: true

File: cli/commands/build.py [hunk 3]
```diff
@@ -378,7 +378,22 @@ def build_solace_agent_mesh(config, build_config_dir, abort, parsers):
         if f.endswith(".yaml")
         and not any(f.startswith(prefix) for prefix in skip_prefixes)
     ]
+    
+    # Check if embedding service is enabled
+    embedding_enabled = True  # Default to True for backward compatibility
```
gregmeldrum, line 383: For backward compatibility, if the embedding service is not adding, we add it automatically. This will make sure that existing installations (Solace Chat) still work like before.

cli/commands/init/create_config_file_step.py
cyrus2281: Can we follow the pattern from `runtime`?

```yaml
solace_agent_mesh: 
  built_in: 
     services:
       embedding:
          enabled: true

```

cyrus2281: Unless you think this approach makes more sense for the built time

cyrus2281 (commented):
A question, so the `do_embedding_request` from base agent class is still there regardless if this is enabled or not?

cli/commands/init/create_config_file_step.py
gregmeldrum: I like that better, thanks, I'll update the code.

cyrus2281 (approved):
Looks Good,
Thanks

PR #77, opened 2025-04-07
AI-657: Fix localhost CORS bug with web ui

RohanSreelesh:
This is the documentation update that goes along with this PR https://github.com/SolaceLabs/solace-ai-connector-web/pull/16. It also updates the default host for the web ui to be `localhost`

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ To get started with Solace Agent Mesh, follow these steps:
 
 5. **Connect to the Web Interface**:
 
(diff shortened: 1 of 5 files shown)


cyrus2281 (commented):
What if it's hosted? inside the container it would still be localhost, but in the brower it would have the container public URL

RohanSreelesh:
> What if it's hosted? inside the container it would still be localhost, but in the brower it would have the container public URL

@cyrus2281 I updated the advanced configuration docs to reflect how to configure for this case

cyrus2281 (approved):
LGTM

PR #80, opened 2025-04-08
AI-709: Use .env file by default and introduce new option -u

RohanSreelesh:
This changes the `sam run` command to utilize the `.env` file by default. For deployment purposes where users might prefer using system environment variables over the .env file, there is a new option `-u` or `--use-system-env` which will prefer the system environment over the .env file.
 In cases where `sam run` is called but there is no `.env` file we already automatically use the system environment, I have made no changes to this behavior.
 
This PR also contains all the documentation changes for the new option + removing references to the `-e` option

### How is this accomplished?
Changing the run command.

Reviewers should ensure documentation is updated properly and the new option works as expected

### DIFF
--- a/docs/docs/documentation/user-guide/custom-gateways.md
+++ b/docs/docs/documentation/user-guide/custom-gateways.md
@@ -542,7 +542,7 @@ class DirWatcherOutput(DirWatcherBase):
 To build and run the gateway, use the following command:
 
 ```bash
-solace-agent-mesh run
+solace-agent-mesh run -b
 ```
 
 This command builds and runs the gateway in one step. Alternatively, you can execute these steps separately.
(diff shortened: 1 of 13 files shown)


gregmeldrum (approved):
Nice! Thanks for this.

docs/docs/documentation/user-guide/custom-gateways.md
```diff
@@ -542,7 +542,7 @@ class DirWatcherOutput(DirWatcherBase):
 To build and run the gateway, use the following command:
 
 ```bash
-solace-agent-mesh run -eb
+solace-agent-mesh run
```
cyrus2281: `-b` should still be here

cyrus2281 (approved):
Looks Good, thanks
Added one comment

PR #82, opened 2025-04-08
FIX: Fix action history while reverting system prompt changes

cyrus2281:

This PR reverts changes made to the system prompt while retaining the fix for action history. The goal is to revert to the previous prompt structure while keeping improvements to how action history is handled.

## How is this accomplished?

- Reverting the system prompt and only keeping action history fix

Ensure the action history fix is maintained while the system prompt structure reverts to its previous form.


efunneko (approved):
Thanks

PR #83, opened 2025-04-09
AI-746: Adjust the prompt and strip some text from results to combat hallucinations

efunneko:

There have been cases with Sonnet 3.7 where it will invoke actions and then
respond with hallucinated data immediately afterwards. This addresses this

### How is this accomplished?

The prompt has been enhanced to more clearly indicate that no response is expected after actions are invoked.

Additionally, the parser strips any text that does follow any invoke actions to really make sure that if it does happen, the text is not sent back to the user.

Think of the big picture of this. Are we breaking anything by not including the text after actions are invoked?

### DIFF
--- a/src/common/utils.py
+++ b/src/common/utils.py
@@ -324,6 +324,7 @@ def parse_orchestrator_response(response, last_chunk=False, tag_prefix=""):
     current_param_value = []
     open_tags = []
     current_text = []
+    seen_invoke_action = False
 
     for line in response.split("\n"):
 
@@ -381,6 +382,7 @@ def parse_orchestrator_response(response, last_chunk=False, tag_prefix=""):
             if in_invoke_action:
                 parsed_data["errors"].append("Nested <invoke_action> tags")
             in_invoke_action = True
+            seen_invoke_action = True
             open_tags.append("invoke_action")
             current_action = {
                 "agent": None,
@@ -464,7 +466,11 @@ def parse_orchestrator_response(response, last_chunk=False, tag_prefix=""):
                 current_param_value.append(line.strip())
 
         else:
-            current_text.append(line)
+            # NOTE that we are intentionally ignoring all output text that occurs
+            # after any <invoke_action> tag. It has been told to never do this and
+            # if it does, then there is a good chance it is hallucinating responses
+            if not seen_invoke_action:
+                current_text.append(line)
 
     if open_tags:
         parsed_data["errors"].append(f"Unclosed tags: {', '.join(open_tags)}")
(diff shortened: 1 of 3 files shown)


src/orchestrator/orchestrator_prompt.py
```diff
@@ -236,12 +231,16 @@ def SystemPrompt(info: Dict[str, Any], action_examples: List[str]) -> str:
...
+    2. Invoke agents' actions to perform these tasks
+    3. After invoking an action with the invoke action directive, finish the response and wait for the action to complete.
+    4. The action will be run and the results will then be returned on a later step. NEVER guess or fill in the response without waiting for the action's response. 
+    5. Prioritize using available actions to fulfill the stimulus whenever possible.
+    6. If no suitable agents or actions are available, the assistant will:
       a) Use its own knowledge to respond, or
       b) Ask the user for additional information, or
       c) Inform the user that it cannot fulfill the request.
   - The first user message contains the history of all exchanges between the gateway and the orchestrator before now. Note that this history list has removed all the agent's action invocation outputs and reasoning.
+  - Do not use the history as a guide for how to respond. That history is only present to provide some context to the coversation. The format and data have been modified and does not show all the assistant's directives.
```
gregmeldrum, line 242: Do we use "directives" anywhere else in the prompt? I'm wondering if we need more precise terminology so the LLM will understand.

gregmeldrum (approved):
Looks good! One question about terminology.

src/common/utils.py
```diff
@@ -464,7 +466,11 @@ def parse_orchestrator_response(response, last_chunk=False, tag_prefix=""):
                 current_param_value.append(line.strip())
 
         else:
-            current_text.append(line)
+            # NOTE that we are intentionally ignoring all output text that occurs
+            # after any <invoke_action> tag. It has been told to never do this and
+            # if it does, then there is a good chance it is hallucinating responses
+            if not seen_invoke_action:
+                current_text.append(line)
```
cyrus2281, line 475: What if the output is

```
< invoke_action>
first action call
</invoke_action>

some random test,

< invoke_action>
second action
</invoke_action>

more random text
```

Wouldnt' in this case, only `more random text` be removed?

cyrus2281 (approved):
Added one comment

The rest looks good! Pushing to beta now

src/common/utils.py
efunneko: For this case, I think both would be removed because no text is stored as soon as the invoke has been seen.

It won't remove it for the storage of history, though. That is a much simpler replacement of text

src/orchestrator/orchestrator_prompt.py
efunneko: Good point. I doubt that we do.

efunneko: I looked at this and tried to find another way to say it that would be more consistent and didn't find one. Ultimately, this is the formula I came up with that didn't create any hallucinations in 100 attempts, so I am reluctant to change anything more.

PR #84, opened 2025-04-12
AI-696: Adjust the orchestrator to work well with Gemini 2.5

efunneko:

Make the Gemini 2.5 model work well with SAM

### How is this accomplished?

There were a few things that needed changing:

1. In the file directives of the orchestrator prompt, it was possible that the model would add t###_ to the internal tags. We now are okay if that happens
2. There were a number of places where parameters would have their whitespace stripped. This was wrong and has been removed
3. Be strict about needing <t### _reasoning> in the response. Gemini occasionally gives a crazy response, which we should error on
4. Handle action parameters being surrounded by CDATA or having XML entities in them.
5. Fix plotly graph examples to not have double curly braces
6. Look for errors in the parsing and reinvoke the orchestrator based on that
7. Explicitly tell the orchestrator to not put ```xml around the response directives

### DIFF
--- a/src/agents/global/actions/plotly_graph.py
+++ b/src/agents/global/actions/plotly_graph.py
@@ -47,7 +47,7 @@ class PlotlyGraph(Action):
                         ],
                         "reasoning": [
                             "- user has requested a random bar graph",
-                            "- invoke the plotly action from the global agent to generate a bar graph with random data"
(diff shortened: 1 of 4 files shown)


gregmeldrum (approved):
Looks good 🧹

PR #89, opened 2025-04-16
AI-613: New initialization GUI

RohanSreelesh:
Introduces the init GUI.

Project breakdown:
There are two flows, the quick flow and the advanced flow. The quick flow contains only the LLM provider setup and review. The advanced flow contains everything in the CLI flow. 

Code breakdown:
There is one main file called [InitializationFlow.tsx](https://github.com/SolaceLabs/solace-agent-mesh/compare/dev...rohan/AI-613/init-gui?expand=1#diff-00f00a2cd562a88ced2967bfdb880823f250687208b946c7eded72ef39af8cd4). this file basically assembles all the steps and passes in the initial data to these components. The initial data is provided by the `default_options` dictionary which used to reside in the CLI folder but now has been moved to `common.py` in the config_portal folder. Most of the steps are similar to the ones we have in the CLI with a few additions such as the initial path selection step and the success screen step. 

Devops consideration:
For this to work as a release we would have to add a step to `npm run build` similar to how we do it for the web visualizer.

Ensure both init flows work as expected.


config_portal/frontend/README.md
```diff
@@ -0,0 +1,51 @@
+# templates/spa
```
gregmeldrum: Do we need this file? Looks like boilerplate.

gregmeldrum (approved):
Look good! Just wondering if we should get rid of that README.

config_portal/frontend/README.md
RohanSreelesh: Removed now

PR #92, opened 2025-04-17
DATAGO-99407: add custom build hook for config_portal build

artyom-morozov:

- install npm deps and build the UI on `hatch build`


RohanSreelesh (approved):
LGTM! thanks for this

PR #94, opened 2025-04-21
Release update

cyrus2281:


tests/test_orchestrator_streaming_output.py
```diff
@@ -14,10 +14,11 @@
 
 FileService(file_manager_config)
 
+
 def strip_lines(text):
     return "\n".join([line.strip() for line in text.strip().split("\n")]).strip()
 
-
+@unittest.skip("Entire test class skipped due to hanging tests")
```
efunneko: This should be removed now that I have merged in the fix for this

efunneko (changes requested):
One small change to re-enable some tests

tests/test_orchestrator_streaming_output.py
RohanSreelesh: It has been removed now in #96

efunneko (approved):
Please go ahead!

PR #95, opened 2025-04-21
AI-781: Fixed some test errors caused by changes for Gemini

efunneko:

Address some test errors that were introduced when Gemini 2.5 changes were added

### How is this accomplished?

Relaxed the requirement for a reasoning tag to be present for some tests and changed some tests to have reasoning tags. Note that the test infrastructure currently expects a single response from the streaming output component for every input. With recent changes, it is quite possible that this does not happen - which is okay, but causes the tests to stall.


cyrus2281 (approved):
LGTM

PR #98, opened 2025-04-22
(Merge dev into main) Update readme and docs to say 3.10.16 instead of 3.10 (#97)

RohanSreelesh:


cyrus2281 (commented):
@efunneko I wonder if we should bring the 3.10.16 to something lower to be more inclusive. 3.10.16 is the latest 3.10 python, so all other versions of python 3.10 are not supported

cyrus2281 (approved):
LGTM

PR #101, opened 2025-04-22
AI-787 : Update docs for init gui and add --use-web-based-init flag

RohanSreelesh:
Update docs for init gui and add --use-web-based-init flag. Added a new step for web init to fit it in better with the existing flow.It also has a small refactor for the [BuiltinAgentSetup.tsx](https://github.com/SolaceLabs/solace-agent-mesh/compare/dev...rohan/init-gui-improvements?expand=1#diff-9a980f5f7bfd72ce264eb8523c9ca6d730ad5efc1ab304c14f2d7068e402adfd) component.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -175,3 +175,5 @@ wip/
 solace-agent-mesh.yaml
 playground.py
 tmp
+
+config_portal/frontend/static/*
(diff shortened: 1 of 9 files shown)


cyrus2281 (approved):
LGTM

PR #105, opened 2025-04-24
AI-790: Modified response format prompt

Hugo-Pare:
Resolve bug where images would something be returned in markdown formatting causing unwanted output.

### How is this accomplished?
Improve response format prompt to make sure files are returned in the desired format


cyrus2281 (commented):
How about https://github.com/SolaceLabs/solace-agent-mesh/blob/hugo/ai-790/streaming-response-file-bug/templates/gateway-config-template.yaml

PR #106, opened 2025-04-24
AI-791 : Init GUI should not add provider/ if already provided by the user

RohanSreelesh:
This came up in testing where if the user provides the model name as `openai/<model_name>`in the init GUI then when they submit another `openai/` is added to this causing the model name to be `openai/openai/<model_name>`

### DIFF
--- a/config_portal/frontend/app/common/providerModels.ts
+++ b/config_portal/frontend/app/common/providerModels.ts
@@ -193,3 +193,16 @@ export async function fetchModelsFromCustomEndpoint(
     return [];
   }
 }
+
+  // Format model name for litellm
+  export const formatModelName = (modelName: string, provider: string): string => {
+  
+    // If model name already includes a provider prefix (contains '/'), return as is
+    if (modelName.includes('/')) {
+      return modelName;
+    }
+        
+    // Get the correct provider prefix
+    const providerPrefix = PROVIDER_PREFIX_MAP[provider] || provider;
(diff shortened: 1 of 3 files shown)


config_portal/frontend/app/common/providerModels.ts
```diff
@@ -192,4 +192,17 @@ export async function fetchModelsFromCustomEndpoint(
...
+  export const formatModelName = (modelName: string, provider: string): string => {
+  
+    // If model name already includes a provider prefix (contains '/'), return as is
+    if (modelName.includes('/')) {
+      return modelName;
+    }
+        
+    // Get the correct provider prefix
+    const providerPrefix = PROVIDER_PREFIX_MAP[provider] || provider;
+    return `${providerPrefix}/${modelName}`;
```
cyrus2281, line 207: But the drop down already says `Openai compatible model`, so we should either change the title to something like `Custom model provider` or only ask for the model name from the user and hard-code the `openai` as it's given from the `openai compabitle` portion

RohanSreelesh: Right now it does hardcode `openai` if they choose `open ai compatible`. The issue is that if they pass in `openai/<modelname>` the llm test will pass but at the end when they submit, the code adds another `openai/ `to the model name making it `openai/openai/<modelname>`. Basically this change is to handle the case where they provide the `<provider>/`for whatever reason (One reason could be they have used the cli before and are used to this)

PR #115, opened 2025-05-06
FIX: Fix memory issues

cyrus2281:

This change fixes a potential memory error that could occur when retrieving a non-existent session from the SQL history provider.

## How is this accomplished?

- Added slight fix for memory by modifying the `get_session` method to check if `row` exists before accessing its data

This is a small defensive coding fix to prevent accessing attributes on a None object when no session is found.


gregmeldrum (approved):
LGTM

PR #117, opened 2025-05-07
AI-731: Fix timeout error stuck in infinite retry loop 

Hugo-Pare:
Proper handling of Timeout Errors

### How is this accomplished?
Implement try/catch logic and re-invoking orchestrator


cyrus2281 (approved):
LGTM

PR #121, opened 2025-05-13
DATAGO-100687: Fix Web Visualizer Crashing Errors

mo-radwan1:

Fixes a crash in the SAM Web Visualizer. The visualizer would become unusable, preventing users from seeing agent capabilities.

### How is this accomplished?
The crash was caused by trying to display objects directly within lists (like params, examples, required_scopes) on the CapabilitiesPage.jsx.
This change updates the page to use JSON.stringify() on these objects, converting them to strings before display. This stops the rendering error and the crash.

N/A


RohanSreelesh (approved):
LGTM

PR #123, opened 2025-05-14
AI-815: Verify broker connection in init gui

RohanSreelesh:
Verify broker connection in init gui, so that users know if they entered the wrong credentials.

verify_broker_demo


cyrus2281 (approved):
LG

PR #125, opened 2025-05-14
AI-712: Refactor Dockerfile and documentation for Solace Agent Mesh

cyrus2281:

This PR improves the Docker implementation for the Solace Agent Mesh by enhancing security, usability, and documentation. The changes make the Docker image more production-ready and provide better guidance for containerized deployments.

## How is this accomplished?

- Refactor Dockerfile to implement security best practices (non-root user)
- Set correct entry point to use the installed CLI package
- Update documentation to explain Docker usage options throughout the application
- Add Docker alternatives in the quick start guide and installation instructions

The new ENTRYPOINT configuration assumes the package is correctly installed in the container's PATH. Make sure this works as expected with the current package setup.


Dockerfile
```diff
@@ -1,23 +1,26 @@
...
 
-ENV PYTHONUNBUFFERED=1
+# Create a non-root user
+RUN groupadd -r samapp && useradd -r -g samapp samapp
+RUN chown -R samapp:samapp /app /tmp
 
+# Switch to non-root user
+USER samapp
 
 LABEL org.opencontainers.image.source https://github.com/SolaceLabs/solace-agent-mesh
```
Hugo-Pare: Prefer using `LABEL <key>=<value>` format instead of legacy `LABEL <key> <value>` format

Hugo-Pare (approved):
Left one comment.
LG

PR #127, opened 2025-05-15
fix(init GUI): Include logo as base64

RohanSreelesh:
Include the solace community logo as base64 for now as the server is shut down by the time they navigate to this page so it cannot be served by the backend


Hugo-Pare (approved):
GTG

PR #132, opened 2025-05-20
Update README.md

Mrc0113:
Add download badge and star call to action to the top of the README


Hugo-Pare (approved):
LG

PR #133, opened 2025-05-22
DATAGO-102726: Add identity to meta properties passed into actions

gregmeldrum:
Allow actions to use identity

### How is this accomplished?
Pull the identity from the user_properties and pass it into the action

No

### DIFF
--- a/src/agents/base_agent_component.py
+++ b/src/agents/base_agent_component.py
@@ -130,8 +130,11 @@ class BaseAgentComponent(LLMServiceComponentBase, ABC):
                 )
             else:
                 resolved_params = data.get("action_params", {}).copy()
+                user_properties = message.get_user_properties() or {}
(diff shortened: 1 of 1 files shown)


src/agents/base_agent_component.py
```diff
@@ -109,86 +109,86 @@ def get_agent_summary(self):
...
+                action_name,
+                json.dumps(data),
+            )
+            return ActionResponse(
+                message="Unauthorized: You don't have permission to perform this action.",
+            )
+        try:
+            meta = {
+                "session_id": session_id,
+                "identity": identity,
```
gregmeldrum, line 164: This is basically the change (adding identity to the meta object) but SonarQube complained about cognitive complexity so I had to refactor the invoke method.

cyrus2281 (approved):
LG

PR #151, opened 2025-07-24
DATAGO-106770: Stop auto fit-view in flowchart once there is user interaction

ziyanwan:

The current auto-fit behavior of the workflow diagram causes the UI to snap to the center on every new render after adding a new element. This can be frustrating for users who are trying to view other parts of the diagram.

This change stops auto fit-view in flowchart once there is user interaction. 

### How is this accomplished?

Add a new state `hasUserInteracted` that is set to `true` if user has interaction (click on node or edge, zoom or pan), and reset to `false` if `taskId` changes. Only call `fitview`() if `hasUserInteracted` is false.

`onMoveStart` can be triggered by initial `fitview`(), so adding `event.isTrusted` condition is necessary for real user action.

### DIFF
--- a/client/webui/frontend/src/lib/components/activities/FlowChartPanel.tsx
+++ b/client/webui/frontend/src/lib/components/activities/FlowChartPanel.tsx
@@ -7,6 +7,7 @@ import { FiPause, FiPlay } from "react-icons/fi";
 
 import { PopoverManual } from "@/lib/components/ui";
 import { useTaskContext } from "@/lib/hooks";
+import { useChatContext } from "@/lib/hooks";
 import type { VisualizerStep } from "@/lib/types";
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/components/activities/FlowChartPanel.tsx
```diff
@@ -510,6 +523,7 @@ const FlowRenderer: React.FC<FlowChartPanelProps> = ({ processedSteps, isRightPa
                 nodesDraggable={false}
                 elementsSelectable={false}
                 nodesConnectable={false}
+                minZoom={0.2}
```
ziyanwan, line 526: Default minZoom is 0.5. Change to 0.2 to give user a larger space.

PR #156, opened 2025-07-28
DATAGO-107349: Part 2 of src refactor, fix direct src references in code and llm.txt

gregmeldrum:

Fix code that uses direct references to the src directory to use src.solace_agent_mesh instead

### How is this accomplished?

Update code references


cyrus2281 (approved):
LGTM

PR #166, opened 2025-07-29
DATAGO-107340: Better handling of sub-agent call timeouts

efunneko:

Fix an issue where if a sub-agent call timed out we would not properly clean up the correlation data for that call and would then later process a response that might show up. 

This added two peer results into the session history causing corruption.

### How is this accomplished?

There was some refactoring to clean up the handling of peer task state as well as issuing a cancel request to the peer agent when the timeout happened to try to stop any further processing.


gregmeldrum (approved):
LG!

PR #168, opened 2025-07-29
DATAGO-107231: Improve artifact creation

RohanSreelesh:
Added more instructions for when to create text based artifacts, it should avoid cases where it goes and creates a comprehensive artifact for a simple query


src/solace_agent_mesh/agent/adk/callbacks.py
```diff
@@ -785,6 +785,26 @@ def _generate_fenced_artifact_instruction() -> str:
 
 The system will automatically save the content and give you a confirmation in the next turn."""
 
+def _generate_artifact_creation_instruction() -> str:
+    return """\
```
cyrus2281: Extra slash

```suggestion
    return """
```

PR #169, opened 2025-07-29
DATAGO-107598: Remove html2text from repo

gregmeldrum:
Removing html2text from the repo since it triggers a whitesource violation.


RohanSreelesh (commented):
LG

PR #183, opened 2025-07-30
DATAGO-107794: new chat session warning dialog

lgh-solace:
Issue:
* When a new chat session is created, a user's chat history and files will be cleared - there's no warning of this behaviour so it could happen unexpected.
* This PR adds a warning dialog for this behaviour. 

image


client/webui/frontend/src/lib/components/chat/ChatSessionDialog.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import { Button, Dialog, DialogContent, DialogDescription, DialogHeader, DialogTitle } from "@/lib/components/ui";
+import { useChatContext } from "@/lib/hooks";
+
+interface ChatSessionDialogProps {
+    isOpen: boolean;
+    onClose: () => void;
+}
+
+export const ChatSessionDialog: React.FC<ChatSessionDialogProps> = ({ isOpen, onClose }) => {
```
lgh-solace, line 11: The new dialog.

client/webui/frontend/src/lib/providers/ChatProvider.tsx
```diff
@@ -752,8 +752,6 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
             });
 
             if (response.status === 202) {
-                addNotification(`Cancellation signal sent for task ${currentTaskId}. Waiting for confirmation...`);
```
lgh-solace, line 755: Removing an extraneous notification. There were two notifications around cancellation. One for the sending the signal and another for completion of cancellation. We only need the latter.

RohanSreelesh (approved):
LG

PR #187, opened 2025-07-31
DATAGO-107284: Add prisma

artyom-morozov:
Add prisma checks 
Add Whitesource cleanup workflow
Push ECR image on push

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,15 +52,9 @@ jobs:
       - name: Get Version from __init__.py
         id: get_version
         run: |
-          version=$(python -c "
-          import re
-          with open('cli/__init__.py', 'r') as f:
-              content = f.read()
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -48,19 +47,131 @@ jobs:
         id: get_commit_hash
         run: |
           echo "commit_hash=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
+          echo "short_sha=${GITHUB_SHA:0:10}" >> $GITHUB_OUTPUT
+
+      - name: Get Version from __init__.py
```
shooshmand-sol: could we use `$(hatch version)` here?

shooshmand-sol (approved):
👍

.github/workflows/ci.yaml
artyom-morozov: Added

PR #189, opened 2025-07-31
DATAGO-107222: Reset timeout for responsive task

alimosaed:
**What is the purpose of this change?**
The timeout on a peer agent call is a safety net for something going badly wrong. If we know that the peer agent is still working successfully because we are getting status responses back for the request, we should reset the timeout

**How is this accomplished?**
Reset the timeout in cache for each task that returns the status or response.

**Anything reviews should focus on/be aware of?**


src/solace_agent_mesh/agent/sac/component.py
```diff
@@ -656,6 +656,42 @@ async def _claim_peer_sub_task_completion(
...
 
+    async def reset_peer_timeout(self, sub_task_id: str):
+        """
+        Resets the timeout for a given peer sub-task.
+        """
+        log_id = f"{self.log_identifier}[ResetTimeout:{sub_task_id}]"
+        log.debug("%s Resetting timeout for peer sub-task.", log_id)
+
+        # Get the original logical task ID from the cache without removing it
+        logical_task_id = self.cache_service.get_data(sub_task_id)
```
enavitan, line 619: quick question on this one, from the ipml it appears we will use either InMemStorage or SQLAlchemyStorage - while for InMemStorage its ok to be blocking for SQL one I assume we should offload this call to asyncio.to_thread ?

please discard above if my understanding is incorrect and we are not hitting persistent storage

alimosaed: I do not think we store sub-task information in persistent database at this moment. If this is in our roadmap, we should refactor this module to cover persistent storage because we are using cache in other methods of this module.

PR #194, opened 2025-08-01
DATAGO-107656: Not all files being returned to chat

RohanSreelesh:
test

### DIFF
--- a/client/webui/frontend/src/lib/providers/ChatProvider.tsx
+++ b/client/webui/frontend/src/lib/providers/ChatProvider.tsx
@@ -431,7 +431,6 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
 
             const isEndOfThisTurn = (isFinalStatusUpdate && !isEmptyLlmResponseSignal) || isFinalResponseEvent || isErrorResponse;
 
-            console.log(`%cHANDLER: Calling setMessages. isEndOfThisTurn=${isEndOfThisTurn}`, "color: orange;");
             setMessages(prevMessages => {
                 let newMessages = [...prevMessages];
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/providers/ChatProvider.tsx
```diff
@@ -883,7 +883,7 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
             console.log(`ChatProvider Effect: currentTaskId is null or apiPrefix missing. Ensuring EventSource is closed.`);
             closeCurrentEventSource();
         }
-    }, [currentTaskId, apiPrefix, handleSseOpen, handleSseError, handleSseMessage, closeCurrentEventSource]);
+    }, [currentTaskId, apiPrefix]);
```
lgh-solace, line 886: Just curious how this change fixes the issue?

RohanSreelesh: 1. State changes (like isResponding) caused the component to re-render.                                                                                                                               
 2. This re-render created new versions of the event handler functions (handleSseMessage, etc.).                                                                                                       
 3. Because those functions were in the useEffect's dependency array, React saw them as "changed" and re-ran the effect.                                                                               
 4. Re-running the effect closed the old connection and opened a new one, causing the cycle.  

Basically the opening and closing of connection was happening on each chunk, for text this was not noticeable but if a lot of files were returned the closing and opening would cause some of the files to not show up. I confirmed by adding a `console.log` before the even source creation and cleanup.

RohanSreelesh: Just to add I was actually surprised we never had missing text chunks

PR #196, opened 2025-08-01
DATAGO-107894:test-coverage-server

zhenyu369:


tests/integration/scenarios_programmatic/test_server.py
```diff
@@ -0,0 +1,165 @@
+import asyncio
+from collections.abc import AsyncIterable
+
+from solace_agent_mesh.common.types import (
```
Hugo-Pare, line 5: It should be important to import SAM from the same place. Either from the package or the source code.

tests/integration/scenarios_programmatic/test_server.py
```diff
@@ -0,0 +1,165 @@
...
+    TaskStatus,
+    TaskStatusUpdateEvent,
+)
+
+
+def test_A2A_server():
+
+    from src.solace_agent_mesh.common.server.server import A2AServer
+    from src.solace_agent_mesh.common.server.task_manager import InMemoryTaskManager
+    from src.solace_agent_mesh.common.types import (
```
Hugo-Pare, line 21: Here is import from the `src/` folder

Hugo-Pare (approved):
LG, just a comment

cyrus2281 (approved):

PR #199, opened 2025-08-01
DATAGO-107836: Large message payload guard

gregmeldrum:


src/solace_agent_mesh/common/utils/message_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        json_str = json.dumps(payload, separators=(",", ":"), ensure_ascii=False)
+        return len(json_str.encode("utf-8"))
+    except (TypeError, ValueError) as e:
+        # Graceful fallback if JSON serialization fails
+        log.warning(
+            f"Failed to serialize payload to JSON for size calculation: {e}. "
+            f"Using string representation fallback."
+        )
+        try:
+            return len(str(payload).encode("utf-8"))
```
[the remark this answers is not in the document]
gregmeldrum: We've already failed the json serialization so how would this work?

src/solace_agent_mesh/agent/sac/component.py
```diff
@@ -104,7 +106,15 @@
...
-    "config_parameters": [],
+    "config_parameters": [
+        {
+            "name": "max_message_size_bytes",
+            "required": False,
+            "type": "integer",
+            "default": 10_000_000,
+            "description": "Maximum allowed message size in bytes before rejecting publication to prevent broker disconnections. Default: 10MB",
+        }
+    ],
```
efunneko: I think this config should move to the app.py config schema with all the other configs.

Configs on the app are also available on the component through get_config()

src/solace_agent_mesh/common/utils/message_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    Returns:
+        The size of the payload in bytes.
+
+    Note:
+        If JSON serialization fails, falls back to string representation
+        for size estimation.
+    """
+    try:
+        # Use JSON serialization + UTF-8 encoding to match Solace AI Connector
+        json_str = json.dumps(payload, separators=(",", ":"), ensure_ascii=False)
```
efunneko, line 35: I am a bit concerned about how expensive this might be. I know that if we don't do this, it would be impossible to get an accurate size, but I wonder if we really need that level of accuracy. 

For now, we should probably just go ahead with it since this optimization is not our primary concern. Perhaps in the future if we are needing to streamline performance, we will have to do it differently.

src/solace_agent_mesh/common/utils/message_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        - int: The actual size of the payload in bytes
+
+    Note:
+        Logs an error if size exceeds the limit.
+    """
+    actual_size = calculate_message_size(payload)
+
+    # Check if size exceeds the limit
+    if actual_size > max_size_bytes:
+        log.error(
```
efunneko: [Q] Are we going to a get multiple errors in the log if we do it here and at least one or more times in the higher layers? Funny, coming from me who is responsible for so many useless logs right now...

efunneko (changes requested):
A couple of comment and one change request

src/solace_agent_mesh/common/utils/message_utils.py
gregmeldrum: Removed this log since there is another error log at the next layer up.

src/solace_agent_mesh/agent/sac/component.py
gregmeldrum: moved to app.py

efunneko (approved):
LG - thanks!

PR #204, opened 2025-08-05
DATAGO-108103: Enhance file processing in run command to support directory discovery and YAML validation

cyrus2281:

This change enhances the file processing capabilities of the run command to be more flexible and robust when handling configuration files by supporting directory inputs and implementing proper YAML file validation.

## How is this accomplished?

- Enhance file processing in run command to support directory discovery and YAML validation
- Allow directory inputs to be accepted alongside individual files
- Add recursive scanning of directories for YAML files (.yaml, .yml extensions)
- Add validation to ensure only proper YAML files are processed
- Maintain consistent filtering logic for files with underscore prefixes or "shared_config" naming

The change maintains backward compatibility with existing behavior while adding new functionality. Verify that directory scanning works properly with nested YAML files and that the skip logic is correctly applied.


gregmeldrum (approved):
LG! Thanks for this.

PR #205, opened 2025-08-05
DATAGO-108108: Update plugin links, add Dockerfile instructions, and enhance agent configurations with logging and shared settings

cyrus2281:

This PR adds sample configurations and fixes documentation links for plugins, making it easier for users to get started with Solace Agent Mesh. It includes basic and core sample configurations for agents, gateways, logging, and shared settings, along with Dockerfile updates to include these samples in the container image.

## How is this accomplished?

- Fix plugin link references in agents, gateways, and quick start documentation
- Add COPY instruction for sample SAM applications in Dockerfile
- Add core and basic agent configurations, logging, and shared settings

Please verify that the sample configurations follow best practices and the directory structure makes sense for new users. Also confirm that the Dockerfile COPY instruction properly includes all necessary sample files.


gregmeldrum (approved):
LGTM!

PR #207, opened 2025-08-05
DATAGO-108136: Fix the extraction of CSV data from a JSON field

efunneko:
There were various issues with the extraction of text from a JSON field and then converting that text to a csv file.


src/solace_agent_mesh/common/utils/embeds/converter.py
```diff
@@ -93,6 +93,196 @@ def _parse_string_to_list_of_dicts(
...
+
+    lines = text.strip().split("\n")
+
+    # Must have at least 2 lines (header + data)
+    if len(lines) < 2:
+        return False
+
+    # Check that lines contain commas (basic CSV indicator)
+    for line in lines:
+        if "," not in line.strip():
```
gregmeldrum, line 121: You can have csv without commas:

```
Name
Billy
Joe
```

gregmeldrum (approved):
Look good! Just 1 comment about csv validation.

PR #216, opened 2025-08-07
DATAGO-108430: Fix log level stdio configuration

Hugo-Pare:
Re-configuring the logging after solace ai connector startup

### How is this accomplished?
Calling the reconfigure logging once the `run` command is called

This change goes hand to hand with another change in the solace-ai-connector repo
Does not rely on the changes to solace-ai-connector


RohanSreelesh (approved):
LG

PR #220, opened 2025-08-08
DATAGO-108589: preventing editing but supporting selection to allow copying

lgh-solace:
In the artifact previews, there was an issue with the previous solution of using `contentEditable="true"` to allow copying (via cmd-a, for example). It allowed copying, but also supported the content being editable (without edits being persistable).

There has been some confusion so replacing that support with a custom hook that can be used anywhere selecting is required.

### DIFF
--- a/client/webui/frontend/src/lib/hooks/useCopy.ts
+++ b/client/webui/frontend/src/lib/hooks/useCopy.ts
@@ -8,7 +8,7 @@ interface UseCopyReturn<T extends HTMLElement> {
 }
 
 /**
- * A custom hook that provides a ref and keydown handler for selecting content for copying.
+ * A custom hook that provides a ref and keydown handler for selecting content to copy it.
  */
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/components/chat/preview/Renderers/MarkdownRenderer.tsx
```diff
@@ -2,11 +2,19 @@ import React from "react";
 
 import { MarkdownHTMLConverter } from "@/lib/components";
 import type { BaseRendererProps } from ".";
+import { useCopy } from "../../../../hooks/useCopy";
 
 export const MarkdownRenderer: React.FC<BaseRendererProps> = ({ content }) => {
+    const { ref, handleKeyDown } = useCopy<HTMLDivElement>();
```
lgh-solace, line 8: Removing `contentEditable="true"` for a reusable hook that can be used where the code should be selectable, but not editable.

PR #230, opened 2025-08-12
DATAGO-108718: Greatly improve MCP result handling

efunneko:
This PR does the following things:

* Do 'intelligent' processing of MCP results. It now handles:
  * Audio, Image and Resource types in the response content list
  * For Text items, it will try to determine if it is json, yaml, csv or markdown and save the item as an artifact of that type
* Cleanup a problem where an MCP tool could have the same name as an internal tool. Ensure that this is handled gracefully
* Add a test MCP server for the integration tests
  * Can test both stdio and http transports
* Add a bunch of tests for all of this


tests/integration/conftest.py
```diff
@@ -250,12 +252,6 @@ def create_agent_config(
         }
 
     test_agent_tools = [
-        {
-            "tool_type": "python",
-            "component_module": "solace_agent_mesh.agent.tools.test_tools",
-            "function_name": "time_delay",
-            "component_base_path": ".",
-        },
```
Hugo-Pare, line 258: Keeping the `time_delay` tool in the conftest file should be important since it is used to create 'aritificial' timeouts in other tests (see `test_cancellation_flows.py`)

tests/integration/scenarios_declarative/test_mcp_processing.py
Hugo-Pare, line 1: Empty file, should be removed

Hugo-Pare (approved):
LG, left two comments

tests/integration/conftest.py
efunneko: This tool is already included as part of the group:

        {"tool_type": "builtin-group", "group_name": "test"},

As part of some fixing in this branch I tightened up checking for duplicate tool names and it caught that time_delay was a duplicate

PR #233, opened 2025-08-13
DATAGO-106076: Support s3 for artifact storage

mo-radwan1:
What is the purpose of this change?

  Add S3-compatible storage support for artifact persistence, enabling users to store agent artifacts in AWS S3 or S3-compatible
  services like MinIO instead of local filesystem or GCS.

  How is this accomplished?

  - Implement S3ArtifactService using boto3 with structured key format: {scope}/{user_id}/{session_id}/{filename}/{version}
  - Add S3 configuration options to CLI init and add-agent commands
  - Support multiple authentication methods: env vars, AWS credentials file, config file, and IAM roles
  - Include frontend UI for S3 configuration in config portal

  Anything reviewers should focus on/be aware of?

  - Install requires pip install solace-agent-mesh[s3] to include boto3 dependency
  - Minimal IAM permissions documented in S3ArtifactService class docstring
  - TLS verification can be disabled via verify_ssl: false for self-signed certificates
  - Defaults to AWS S3 endpoint when endpoint_url not specified


cli/commands/add_cmd/agent_cmd.py
```diff
@@ -66,6 +69,26 @@ def _write_agent_yaml_from_data(
...
+                )
+                if endpoint_url_val:
+                    custom_artifact_lines.append(f'endpoint_url: "{endpoint_url_val}"')
+
+                region_val = config_options.get(
+                    "artifact_service_region",
+                    AGENT_DEFAULTS.get("artifact_service_region", "us-east-1"),
+                )
+                if region_val:
+                    custom_artifact_lines.append(f'region: "{region_val}"')
```
sgamelin, line 91: How does this work with environment variables? Are environment variables referenced directly in the configuration file?

cli/commands/init_cmd/__init__.py
```diff
@@ -30,6 +30,9 @@ def _get_flat_orchestrator_defaults():
     flat_defaults["artifact_service_scope"] = O_DEFAULTS["artifact_service"][
         "artifact_scope"
     ]
+    flat_defaults["artifact_service_bucket_name"] = O_DEFAULTS["artifact_service"].get("bucket_name", "")
+    flat_defaults["artifact_service_endpoint_url"] = O_DEFAULTS["artifact_service"].get("endpoint_url", "")
```
sgamelin, line 34: Should the app fail to startup if these are empty (i.e., not assuming defaults)? How would we know if these are misconfigured?

cli/commands/init_cmd/env_step.py
```diff
@@ -168,6 +171,27 @@ def create_env_file(project_root: Path, options: dict, skip_interactive: bool) -
...
+            False,
+            "S3_ENDPOINT_URL",
+        ),
+        (
+            "s3_region",
+            "S3_REGION",
+            "Enter S3 Region (for S3 artifact service)",
+            False,
+            "S3_REGION",
+        ),
```
sgamelin, line 194: Are we supporting authenticated bucket access (i.e. access key ID + secret access key)?

config_portal/frontend/app/components/steps/gateway/GatewayArtifactServiceStep.tsx
```diff
@@ -106,6 +121,56 @@ const GatewayArtifactServiceStep: React.FC<GatewayArtifactServiceStepProps> = ({
               />
             </FormField>
           )}
+
+          {showS3Config && (
```
sgamelin, line 125: `showS3Config` -> if we end up controlling runtime configurations in our homecloud control plane, rather than at the application level, will this (e.g. `showS3Config`, and related functionality if it exists) be disabled for cloud deployments?

pyproject.toml
```diff
@@ -63,6 +63,7 @@ dependencies = [
 [project.optional-dependencies]
 database = ["sqlalchemy>=2.0"]
 gcs = ["google-cloud-storage>=2.0"]
+s3 = ["boto3>=1.0"]
```
sgamelin: I'm not too familiar with this syntax, but should we specify a specific version to ensure repeatable (idempotent) builds?

src/solace_agent_mesh/agent/adk/services.py
```diff
@@ -184,6 +185,97 @@ def initialize_artifact_service(component) -> BaseArtifactService:
...
+            if endpoint_url:
+                s3_config["endpoint_url"] = endpoint_url
+                
+            region = config.get("region")
+            if region:
+                s3_config["region_name"] = region
+                
+            # AWS credentials can be provided via environment variables or config
+            aws_access_key_id = config.get("aws_access_key_id") or os.environ.get("AWS_ACCESS_KEY_ID")
+            aws_secret_access_key = config.get("aws_secret_access_key") or os.environ.get("AWS_SECRET_ACCESS_KEY")
```
sgamelin, line 276: the other s3 variables are generic in name (`S3_*`), should these also be generic in name (e.g. `S3_ACCESS_KEY_ID`, etc.), for consistency?

src/solace_agent_mesh/agent/adk/artifacts/s3_artifact_service.py
```diff
@@ -0,0 +1,435 @@
...
+            raise ValueError("bucket_name cannot be empty for S3ArtifactService")
+        if not scope_identifier:
+            raise ValueError("scope_identifier cannot be empty for S3ArtifactService")
+
+        self.bucket_name = bucket_name
+        self.scope_identifier = scope_identifier
+
+        if s3_client is None:
+            try:
+                self.s3 = boto3.client("s3", **kwargs)
```
sgamelin, line 79: Is there a way to skip TLS verification? This will eventually be required for customers using HTTPS S3 endpoints and self-signed certs (i.e., certificates not under the global trust store).

src/solace_agent_mesh/agent/adk/artifacts/s3_artifact_service.py
sgamelin, line 1: Can you please list the specific S3 permissions required by this service? This will be useful when we configure policies on the bucket during cloud persistence layer provisioning.

cli/commands/init_cmd/env_step.py
sgamelin: I see that it's referenced under `src/solace_agent_mesh/agent/adk/services.py` with `AWS_*` prefixes (please see that comment as well). Should we expose these here to make those options visible?

cli/commands/add_cmd/agent_cmd.py
```diff
@@ -375,6 +398,28 @@ def create_agent_config(
...
+                collected_options,
+                "artifact_service_bucket_name",
+                "S3 bucket name",
+                AGENT_DEFAULTS.get("artifact_service_bucket_name", ""),
+                skip_interactive,
+            )
+            collected_options["artifact_service_endpoint_url"] = ask_if_not_provided(
+                collected_options,
+                "artifact_service_endpoint_url",
+                "S3 endpoint URL (leave empty for AWS S3)",
```
j-setiawan, line 412: I think it's better to default to AWS' URL and keep it required. It's effectively the same thing (i.e. if you don't specify it, you get AWS' endpoint) but with more consistent behaviour (ex. "If I leave this blank, am I getting a default or it really means 'nothing'?").

cli/commands/add_cmd/agent_cmd.py
mo-radwan1: Yeah, these env variables will be referenced directly in the config file

cli/commands/init_cmd/__init__.py
mo-radwan1: This is a good point. We should add validation to fail early if required S3 configs are missing when S3 is selected as the artifact service.

cli/commands/init_cmd/env_step.py
mo-radwan1: Yeah it is supported via `AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY` it is read directly from the `~/.aws/credentials` but I could also update it to take them as env variables

config_portal/frontend/app/components/steps/gateway/GatewayArtifactServiceStep.tsx
mo-radwan1: This shouldnt impact cloud deployment since its the init files for local / deslf hosted

src/solace_agent_mesh/agent/adk/services.py
mo-radwan1: Mmmm yeah we can do that because of how .aws/credentials stores that and wanted to follow the same format but yeah being consistence it good

src/solace_agent_mesh/agent/adk/artifacts/s3_artifact_service.py
mo-radwan1: Let me take a look

cli/commands/init_cmd/env_step.py
mo-radwan1: Current implementation is that we use env variables but fall back to ~/.aws/credentials

src/solace_agent_mesh/agent/adk/services.py
mo-radwan1: AWS_* naming since that's the standard boto3 convention

examples/a2a_mcp_example.yaml
```diff
@@ -190,3 +197,51 @@ apps:
...
+  #       defaultInputModes: ["text"]
+  #       defaultOutputModes: ["text", "file"]
+  #       skills: []
+  
+  #     # --- Discovery & Communication ---
+  #     agent_card_publishing: { interval_seconds: 10 }
+  #     agent_discovery: { enabled: false }
+  #     inter_agent_communication:
+  #       allow_list: []
+  #       request_timeout_seconds: 30
```
efunneko: [Q] It is good to have an example for this, but I doubt people would look in the a2a_mcp_example to find it. It might make more sense to put it in its own file with a more meaningful name?

PR #236, opened 2025-08-13
fix(DATAGO-109069): Docker builds fail when installing playright chromium deps

johnvincentcorpuz:
Playwright deps is not natively supported in python3.11-slim images: 
```
 #16 [10/14] RUN playwright install-deps chromium
#16 0.574 BEWARE: your OS is not officially supported by Playwright; installing dependencies for ubuntu20.04-x64 as a fallback.
#16 0.574 Installing dependencies...
```

The suggested workaround is to install the packages required by playwright manually instead of using playwright install-deps chromium, or finding a base image that's compatible. 

Since we want to keep this image as light as possible, we stuck with python slim

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,25 +9,28 @@ RUN apt-get update && \
     git \
     curl \
     ffmpeg \
-    libnss3 \
-    libxss1 \
     libasound2 \
-    libxtst6 \
-    libgtk-3-0 \
-    libgbm1 \
-    libxrandr2 \
-    libu2f-udev \
+    libatk-bridge2.0-0 \
+    libatk1.0-0 \
+    libatspi2.0-0 \
+    libcairo2 \
+    libcups2 \
+    libdbus-1-3 \
     libdrm2 \
+    libgbm1 \
+    libglib2.0-0 \
+    libnspr4 \
+    libnss3 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -3,9 +3,34 @@ FROM python:3.11-slim
 
 ENV PYTHONUNBUFFERED=1
 
-# Install dependencies
+# Install all dependencies including Node.js and Playwright browser dependencies
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends git curl ffmpeg && \
+    apt-get install -y --no-install-recommends \
```
johnvincentcorpuz, line 8: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/registry/nativeDeps.ts#L1034-L1062 <- list of deps required by playwright instead of `playwright install-deps chromium`

ziyanwan (approved):

Dockerfile
lgh-solace: Just a question - is this fragile?

johnvincentcorpuz: @lgh-solace 
If playwright changed there packaging deps it could be. 
The other trade off is to use a full debian/ubuntu image.

PR #252, opened 2025-08-21
fix artifact panel + embed resolution for persistence

RohanSreelesh:


client/webui/frontend/src/lib/hooks/useArtifacts.ts
```diff
@@ -14,9 +14,10 @@ interface UseArtifactsReturn {
 
 /**
  * Custom hook to fetch and manage artifact data
+ * @param sessionId - The session ID to fetch artifacts for
  * @returns Object containing artifacts data, loading state, error state, and refetch function
  */
-export const useArtifacts = (): UseArtifactsReturn => {
+export const useArtifacts = (sessionId?: string): UseArtifactsReturn => {
```
lgh-solace, line 20: Could we get the sessionId from ChatContext instead of passing it around? It may be simpler?

```
const { sessionId } = useChatContext();
```

RohanSreelesh: That was my initial thought too but the `useArtifacts` hook is being called inside the `ChatProvider` component itself, So if we try to use `useChatContext` inside useArtifacts it creates a circular dependency.

lgh-solace: I wonder if we should have an ArtifactProvider... anyway, we can always refactor later!

PR #255, opened 2025-08-27
DATAGO-109993: Replace react-icons with react-feather for licence compliance

ziyanwan:
`react-icons` bundles multiple icon libraries. One of these libraries (typicons) is licensed under CC-BY-SA-3.0, a content copyleft license that is incompatible with many enterprise software licensing policy. Currently, only feather icons are used, which is MIT licensed. This change switch `react-icons` to `lucide-react`.


client/webui/frontend/package.json
```diff
@@ -71,9 +71,9 @@
     "lucide-react": "^0.511.0",
     "marked": "^15.0.12",
     "radix-ui": "^1.4.2",
-    "react": "^19.1.0",
-    "react-dom": "^19.1.0",
-    "react-icons": "^5.5.0",
+    "react": ">=19.0.0",
```
lgh-solace, line 74: Intentional react change?

client/webui/frontend/src/lib/components/activities/VisualizerStepCard.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 
-import { FiCheckCircle, FiFileText, FiHardDrive, FiLink, FiMessageSquare, FiShare2, FiTerminal, FiUser, FiXCircle, FiZap } from "react-icons/fi";
+import { CheckCircle, FileText, HardDrive, Link, MessageSquare, Share2, Terminal, User, XCircle, Zap } from "react-feather";
```
lgh-solace: Did you look to potentially swap these out for lucide icons? Then we could remove the dependency entirely?

lgh-solace (approved):
Ok! Perhaps we can look to replace the dependency entirely in another iteration?

client/webui/frontend/package.json
ziyanwan: Not intentional. 
It is a result of Peer dependencies reconciliation when doing npm install. 
We have "react": ">=19.0.0" in peerDependencies.

client/webui/frontend/src/lib/components/activities/VisualizerStepCard.tsx
ziyanwan: Good idea. Let me see if I can do it in this PR.

PR #256, opened 2025-08-27
DATAGO-110002: Mikhail/DATAGO-110002 streamable-http support enabled

mgevantmakher:
- streamable-http support enabled
- mcp doc updated


efunneko:
Ideally, we add to the test environment MCP server to also support streamable HTTP (it uses the fastmcp 2.0 module, so it should have the support). Then we can have a few tests for streamable

RZuchniak:
- Updated test infrastructure to enable running both transports
- Added integration test to ensure streamable-http connections are working
- Small formatting fix (slipped through when rebasing)

client/webui/frontend/src/lib/providers/ChatProvider.tsx
```diff
@@ -635,17 +635,17 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
...
-                    isUser: false,
-                    isComplete: true,
-                    role: "agent",
-                    metadata: {
-                        sessionId: "", // Empty - will be populated when session is created
-                        lastProcessedEventSequence: 0,
-                    },
-                },
-            ]
+                  {
```
efunneko, line 638: Q: I am curious where this change came from. Seems very unrelated...

efunneko (changes requested):
You should look at config_tool_types.py. I think you may need to adjust the McpToolConfig model definition

client/webui/frontend/src/lib/providers/ChatProvider.tsx
RZuchniak: We recently added a prettier pre-commit script and I think it triggered and made these changes. I can try to get rid of them by merging with main again.

PR #257, opened 2025-08-29
chore(DATAGO-110120): fixing package-lock.json

lgh-solace:
Fixing `npm ci`.

BEFORE:
Screenshot 2025-08-29 at 12 28 38 PM

AFTER:
image


.hooks/pre-commit
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+
+# Check if any files in the frontend directory have been staged
+FRONTEND_CHANGES=$(git diff --cached --name-only | grep "^client/webui/frontend/")
```
lgh-solace, line 4: Adding code to check if there's a client frontend change and running a pre-commit check before allowing the commit. Note: other checks can be added as required for BE, etc.

PR #271, opened 2025-09-04
chore(DATAGO-110389): improving handling of decoding errors 

lgh-solace:
If the BE returned a message with string content that was not base64 encoded, the FE would fall over in the attempt to decode it. Now, it will be rendered as is. This can likely be improved and BE changes are coming to improve the content returned, but this change will prevent the UI from falling over.

### DIFF
--- a/client/webui/frontend/src/lib/components/chat/ChatMessage.tsx
+++ b/client/webui/frontend/src/lib/components/chat/ChatMessage.tsx
@@ -55,7 +55,6 @@ const MessageContent: React.FC<{ message: MessageFE }> = ({ message }) => {
         modifiedText = modifiedText.replace(item.originalMatch, "");
 
         if (item.type === "file") {
-            // This is our new case for non-renderable data URIs
             const fileAttachment: FileAttachment = {
                 name: item.filename || "downloaded_file",
                 content: item.content,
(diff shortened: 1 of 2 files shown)


client/webui/frontend/src/lib/components/chat/ChatMessage.tsx
```diff
@@ -66,24 +65,15 @@ const MessageContent: React.FC<{ message: MessageFE }> = ({ message }) => {
...
-                    finalContent = decodeBase64Content(item.content);
-                } catch (e) {
-                    console.error("Failed to decode base64 content for embedded item:", e);
-                    setRenderError("Failed to decode content for preview.");
-                    // maybe skip this item
-                    return;
-                }
+        } else if (!RENDER_TYPES_WITH_RAW_CONTENT.includes(item.type)) {
+            const finalContent = decodeBase64Content(item.content);
+            if (finalContent) {
```
lgh-solace, line 70: `decodeBase64Content` no longer throws an error so removing the try catch.

client/webui/frontend/src/lib/components/chat/preview/previewUtils.ts
```diff
@@ -234,7 +232,7 @@ export function decodeBase64Content(content: string): string {
             return atob(content);
         } catch (atobError) {
             console.error("Failed to decode base64 content with atob fallback:", atobError);
-            throw new Error("Invalid base64 string");
+            return content;
```
lgh-solace, line 235: Return the original content - it's a string that is not base64 encoded.

PR #272, opened 2025-09-04
Use builder phase to reduce size of final image

j-setiawan:
Also add support for ARM builds by installing `build-essential`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,66 +1,56 @@
-
-FROM python:3.11-slim
+FROM python:3.11-slim AS builder
 
 ENV PYTHONUNBUFFERED=1
+WORKDIR /app
 
 # Install all dependencies including Node.js and Playwright browser dependencies
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
+    build-essential \
     git \
-    curl \
-    ffmpeg \
-    libasound2 \
-    libatk-bridge2.0-0 \
-    libatk1.0-0 \
-    libatspi2.0-0 \
-    libcairo2 \
-    libcups2 \
-    libdbus-1-3 \
-    libdrm2 \
-    libgbm1 \
-    libglib2.0-0 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,67 +1,64 @@
...
-    libnss3 \
-    libpango-1.0-0 \
-    libx11-6 \
-    libxcb1 \
-    libxcomposite1 \
-    libxdamage1 \
-    libxext6 \
-    libxfixes3 \
-    libxkbcommon0 \
-    libxrandr2 \
```
mo-radwan1, line 32: I noticed that we moved some of these down but do we not need this full list ? 

we are only doing an install on these 
RUN apt-get update && \
    apt-get install -y --no-install-recommends \
    ffmpeg \
    git \
    curl && \
    curl -sL https://deb.nodesource.com/setup_24.x | bash - && \
    apt-get install -y --no-install-recommends nodejs && \
    python3.11 -m pip install playwright && \
    playwright install-deps chromium && \
    apt-get purge -y --auto-remove && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

j-setiawan: ` playwright install-deps chromium` installs all those deps without us having to hardcode them.

Dockerfile
```diff
@@ -1,67 +1,64 @@
...
 ENV PYTHONUNBUFFERED=1
+WORKDIR /app
 
 # Install all dependencies including Node.js and Playwright browser dependencies
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
+    build-essential \
     git \
-    curl \
-    ffmpeg \
```
cyrus2281, line 11: needs to be brought back

carolmorneau: Removing `ffmped` here should be fine if it's not needed at build time. The library is brought in at line 30 for the runtime image.

PR #281, opened 2025-09-09
feat(DATAGO-110888): Add dynamic tools into the 'python' tool type

efunneko:
This PR adds the concept of dynamic tools that can be part of the 'python' types of tools. For this, your tool module will implement a DynamicTool or a DynamicToolProvider so that the tool can have a programmatically created name, signature and description. The provider is capable of creating many tools for this.

See docs/docs/documentation/user-guide/creating-python-tools.md for a full description


src/solace_agent_mesh/agent/adk/setup.py
```diff
@@ -90,48 +126,117 @@ def _check_and_register_tool_name(name: str, source: str):
...
-                    specific_tool_config = tool_config.get("tool_config")
-                    tool_callable = ADKToolWrapper(
-                        func,
-                        specific_tool_config,
-                        function_name,
-                        origin="python",
-                        raw_string_args=tool_config.get("raw_string_args", []),
-                    )
+                        if tool_name:
+                            function_name = tool_name
```
alimosaed, line 158: [Q] is this your intention to overwrite the function name with the tool name?

src/solace_agent_mesh/agent/tools/dynamic_tool.py
```diff
@@ -0,0 +1,313 @@
...
+        str: adk_types.Type.STRING,
+        int: adk_types.Type.INTEGER,
+        float: adk_types.Type.NUMBER,
+        bool: adk_types.Type.BOOLEAN,
+        list: adk_types.Type.ARRAY,
+        dict: adk_types.Type.OBJECT,
+    }
+
+    for param in sig.parameters.values():
+        if param.name in ("tool_context", "tool_config", "kwargs", "self", "cls"):
```
alimosaed, line 195: [Q] cannot we set a proper tag for the tool's schema? This exclusion may be problematic in the future when we add a specific tag for some other settings and we forget to exclude them here.

efunneko: I am not sure exactly what you are suggesting. Can you give more information?

alimosaed: In this line if param.name in ("tool_context", "tool_config", "kwargs", "self", "cls"): we exclude these five parameters from the function signature and collect the remaining parameters as the schema. I was wondering if we want to add a specific tool parameter (e.g., tool_config_X) in the future that handles a specific feature, we have to update this line of code to exclude tool_config_X.
I was thinking that we add a specific filed in config (e.g., dynamic_params) rather than excluding the above parameters.

PR #294, opened 2025-09-15
feat(DATAGO-111607): Add per-tool lifecycle functions

efunneko:
This adds 'init' and 'cleanup' hooks to SAM so that python tools can have lifecycle functions. For raw python tools, the functions must be declared in the agent config. For dynamic python functions, they are part of the dynamic tool and will be called automatically.


cyrus2281 (approved):
LGTM

PR #296, opened 2025-09-15
feat(DATAGO-111231): Add enterprise migration support to WebUI backend

mo-radwan1:
- Add auto-detection for enterprise migration package
- Run community migrations first, then enterprise migrations
- Graceful fallback when enterprise package not available
- Clean up duplicate migration logic and improve error handling


Hugo-Pare (approved):
LG

PR #298, opened 2025-09-16
fix(DATAGO-111703): Buffer task events that arrive in the gateway before the frontend creates an SSE connection

efunneko:
It seems to be possible sometimes for an agent to send events back to the user before the SSE connection is established for that task. We need to buffer those events until the SSE connection arrives


alimosaed (approved):
LG

ghaithdallaali (approved):

PR #301, opened 2025-09-16
DATAGO-111683: Set default value for gateway_max_message_size_bytes

Hugo-Pare:


src/solace_agent_mesh/gateway/base/app.py
```diff
@@ -111,6 +111,17 @@ def __init__(self, app_info: Dict[str, Any], **kwargs):
             code_config_app_block, yaml_app_config_block
         )
 
+        try:
+            # Validate the configuration against the base model
+            validated_config = BaseGatewayAppConfig.model_validate_and_clean(
+                resolved_app_config_block
+            )
+            # Use the validated model's dict representation
```
cyrus2281, line 119: Doesn't this happen automatically?

Hugo-Pare: It was not for gateways other than Web UI

cyrus2281: Ok, since now that we have it in the base, should we remove it from the webui then?

Hugo-Pare: No, the WebUI has extra parameters that need to be validated too

PR #308, opened 2025-09-16
feat(DATAGO-106919): add tool to convert pdf to markdown

g-raman:
# Description
Add existing conversion tool for converting file types into markdown into tool registry

### DIFF
--- a/src/solace_agent_mesh/agent/tools/builtin_artifact_tools.py
+++ b/src/solace_agent_mesh/agent/tools/builtin_artifact_tools.py
@@ -36,7 +36,7 @@ from ...common.utils.embeds import (
 )
 from ...agent.utils.context_helpers import get_original_session_id
 from ...agent.adk.models.lite_llm import LiteLlm
-from google.adk.models import LlmRequest
+from google.adk.models.llm_request import LlmRequest
 from google.adk.models.registry import LLMRegistry
 from ...common.utils.mime_helpers import is_text_based_file
 


solace-mdupls:
@g-raman are you still looking for a review for this? Can you include more detail in the description please? I'm curious as to root cause here. The bug states

> I've done it many times in the past, but I notice it's failing every time now.

So what broke? Is this relating to a refactoring that omitted the tool definition?

g-raman:
> @g-raman are you still looking for a review for this? Can you include more detail in the description please? I'm curious as to root cause here. The bug states
> 
> > I've done it many times in the past, but I notice it's failing every time now.
> 
> So what broke? Is this relating to a refactoring that omitted the tool definition?



> @g-raman are you still looking for a review for this? Can you include more detail in the description please? I'm curious as to root cause here. The bug states
> 
> > I've done it many times in the past, but I notice it's failing every time now.
> 
> So what broke? Is this relating to a refactoring that omitted the tool definition?

Actually an outdated PR that needs to be updated. I'm still investigating the issue but there seems to be an issue with the actual MarkItDown library. It seems to return an empty string for some PDFs.

src/solace_agent_mesh/agent/tools/builtin_artifact_tools.py
cyrus2281, line 1: We already have a built-in tool for this, what's the purpose of this PR?

g-raman: i haven't updated this PR in some time, these changes were deemed unnecessary already.

The issues seems to be with the markitdown library, and I'm working towards a fix

PR #309, opened 2025-09-17
chore(DATAGO-111804): Update to epoch time and convert it in the UI

mo-radwan1:
This PR updates the tables to use epoch time and then convert it in the UI to show the user the proper time based on the time zone. Also converted from using the term created at and updated at to created Time and updatedTime

### DIFF
--- a/examples/agents/orchestrator_example.yaml
+++ b/examples/agents/orchestrator_example.yaml
@@ -40,7 +40,11 @@ apps:
         - You must then review the list of artifacts and return the ones that are important for the user by using the `signal_artifact_for_return` tool.
         - Provide regular progress updates using `status_update` embed directives, especially before initiating any tool call.
 
-      session_service: *default_session_service # Points to shared_config.yaml
+      # session_service: *default_session_service # Points to shared_config.yaml
+      session_service:
+        type: "sql"
+        database_url: "${ORCHESTRATOR_DATABASE_URL}"
+        default_behavior: "PERSISTENT"    
       artifact_service: *default_artifact_service # Points to shared_config.yaml
       artifact_handling_mode: "reference" # Embed artifacts created by *this* agent
       enable_embed_resolution: true # Enable embed feature and instruction injection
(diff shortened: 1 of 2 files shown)


client/webui/frontend/src/lib/utils/format.ts
```diff
@@ -14,11 +15,14 @@ export const formatBytes = (bytes: number, decimals = 2): string => {
 
 /**
  * Helper function to format date (relative time)
- * @param dateString 
+ * @param dateString
  */
 export const formatRelativeTime = (dateString: string): string => {
+    if (!dateString) return "Invalid date";
```
lgh-solace: Maybe change "Invalid date" to "N/A"? That would be more typical in cloud EP.

client/webui/frontend/src/lib/components/chat/SessionList.tsx
```diff
@@ -4,12 +4,13 @@ import { Trash2, Check, X, Pencil, MessageCircle } from "lucide-react";
...
 import { useChatContext, useConfigContext } from "@/lib/hooks";
 import { authenticatedFetch } from "@/lib/utils/api";
+import { formatTimestamp } from "@/lib/utils/format";
 import { Button } from "@/lib/components/ui/button";
 
 interface Session {
     id: string;
-    created_at: string;
-    updated_at: string;
+    created_time: string;
```
lgh-solace: It's not "standard" to be dealing with snake case in javascript - would this be a good opportunity to change the response to using camel case in the response?

client/webui/frontend/src/lib/utils/format.ts
```diff
@@ -33,18 +37,20 @@ export const formatRelativeTime = (dateString: string): string => {
...
         console.error("Error formatting date:", e);
-        return dateString;
+        return "Invalid date";
     }
 };
 
 /**
  * Helper function to format ISO timestamp
- * @param isoString 
+ * @param isoString
```
lgh-solace, line 46: Is this still an isoString or timestamp? Just wondering if we should update the param name?

client/webui/frontend/src/lib/components/chat/SessionList.tsx
mo-radwan1: Yeah, will fix that

client/webui/frontend/src/lib/utils/format.ts
mo-radwan1: Its iso string Yeah let me update that

PR #314, opened 2025-09-18
fix(DATAGO-111615): change orchestrator to use persistence by default

RZuchniak:
- Changed cli to use 'sql' for orchestrator session_type
- Added cli option to add default sql url to .env (true by default)

### DIFF
--- a/config_portal/frontend/app/components/AddAgentFlow.tsx
+++ b/config_portal/frontend/app/components/AddAgentFlow.tsx
@@ -22,6 +22,7 @@ export interface AgentFormData {
 
   session_service_type?: string;
   session_service_behavior?: string;
+  database_url?: string;
 
   artifact_service_type?: string;
   artifact_service_base_path?: string;


RohanSreelesh:
Please update the web based init as well to add this option. And [documentation](https://solacelabs.github.io/solace-agent-mesh/docs/documentation/getting-started/configurations#artifact-service)

docs/docs/documentation/getting-started/configurations/configurations.md
```diff
@@ -155,7 +155,7 @@ The parameters are described in the following table:
 
 | Parameter | Options | Description | Default |
 | :--- | :--- | :--- | :--- |
-| `type` | `memory`, `vertex_rag` | Configuration for ADK Session Service | `memory` |
+| `type` | `memory`, `sql`, `vertex_rag` | Configuration for ADK Session Service | `memory` |
```
RZuchniak, line 162: I'm not too sure what to write as the default here. The `shared_config.yaml` still has memory as default, but we make sql the default for both webui and orchestrator

RohanSreelesh: I think we can keep that to say the default is memory but add a tip block informing the user that the Orchestrator and webui default to sql to allow persistent sessions. Example of a tip block: https://solacelabs.github.io/solace-agent-mesh/docs/documentation/getting-started/quick-start

docs/docs/documentation/getting-started/configurations.md
```diff
@@ -155,9 +155,13 @@ The parameters are described in the following table:
 
 | Parameter | Options | Description | Default |
 | :--- | :--- | :--- | :--- |
-| `type` | `memory`, `vertex_rag` | Configuration for ADK Session Service | `memory` |
+| `type` | `memory`, `sql`, `vertex_rag` | Configuration for ADK Session Service | `memory` |
 | `default_behavior` | `PERSISTENT`, `RUN_BASED` | The default behavior of keeping the session history | `PERSISTENT` |
 
+:::tip
+Although the default session service type is `memory`, both Orchestrator and Webui use `sql` as their session service to allow for persistent sessions.
```
RohanSreelesh:
```suggestion
Although the default session service type is `memory`, both Orchestrator Agent and Web UI gateway use `sql` as their session service to allow for persistent sessions.
```

RohanSreelesh: Small doc suggestion

RohanSreelesh (changes requested):
Sorry missed this in my first pass but need to update the add agent GUI as well. 
image

PR #315, opened 2025-09-18
fix: add user_profile

alimosaed:
**Goal:**
- Revert passing _kwargs_ parameter to the identity provider
- Instead, include the user profile in the auth_claims parameter


alimosaed:
Seems that this change has been addressed recently. @efunneko will decide to merge/close this PR.

cyrus2281 (approved):
LG

PR #332, opened 2025-09-23
fix(DATAGO-112495): Fix large upload problem

efunneko:
We had an issue with uploading files larger than 1 MB

### DIFF
--- a/.github/actions/push-docker/action.yml
+++ b/.github/actions/push-docker/action.yml
@@ -48,6 +48,7 @@ runs:
         context: "."
         file: "Dockerfile"
         push: true
+        platforms: linux/amd64,linux/arm64
         tags: ${{ steps.docker_tags.outputs.TAGS_LIST }}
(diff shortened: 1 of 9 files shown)


src/solace_agent_mesh/gateway/http_sse/routers/sessions.py
```diff
@@ -15,10 +18,45 @@
...
+    """Creates a new session on-demand and returns its ID."""
+    user_id = user.get("id")
+    log.info("User %s requesting new session", user_id)
+    try:
+        new_session_id = session_manager.create_new_session_id(request)
+        log.info("Created new session ID: %s for user %s", new_session_id, user_id)
+
+        # Attempt to create the session record in the DB.
+        # The service will handle the check for whether persistence is enabled.
+        session_service.create_session(
```
alimosaed, line 42: Do not you need to check if the session with new_session_id is created before sending the response?

PR #334, opened 2025-09-25
fix(DATAGO-111329): Failed migration should fail the application from starting up 

mo-radwan1:


src/solace_agent_mesh/gateway/http_sse/app.py
```diff
@@ -188,31 +185,6 @@ def __init__(self, app_info: Dict[str, Any], **kwargs):
         )
         super().__init__(app_info, **kwargs)
 
-        try:
```
alimosaed, line 191: Why do you remove this logic while there is not a corresponding login in your changes?

mo-radwan1: Its duplicated logic. We do the migrations in two places

alimosaed (approved):
LG

PR #342, opened 2025-09-26
feat(DATAGO-UNKNOWN): Lazy load all commands

m9p909:
# What
- reduces time for sam --help by 3 seconds, tested using time command
- Speeds up sam run by an unknown amount. Not sure how to test this but it feels close to 3 seconds as well
- Manual tests indicate that this works. 
# How
Adds a Click command wrapper that calls module = __import__(self.module_path, fromlist=[self.command_name]) to import the module and call through to the function when the command wrapper happens. 
In theory there should be no changes to the underlying behaviour except that imports will happen at runtime instead of at start time and the heavy imports don't need to be imported. Like pandas etc. 

Also removed the context settings from the group annotation because -h --help are default

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -1,50 +1,24 @@
 import click
 import os
 import sys
-import importlib
+
+from cli.lazy_group import LazyGroup
 
 SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
 sys.path.append(os.path.dirname(SCRIPT_DIR))
 
 from cli import __version__
 
-
-class LazyCommand(click.Command):
-    """A command that imports its callback lazily when invoked"""
-
-    def __init__(self, name, module_path, command_name, **kwargs):
-        self.module_path = module_path
-        self.command_name = command_name
-        self._original_command = None
-        super().__init__(name, callback=self._lazy_callback, **kwargs)
-
-    def _lazy_callback(self, *args, **kwargs):
-        """Load the actual command on first invocation"""
-        if self._original_command is None:
-            try:
-                module = importlib.import_module(self.module_path)
-                self._original_command = getattr(module, self.command_name)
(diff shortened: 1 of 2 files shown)


cli/main.py
```diff
@@ -6,15 +6,44 @@
...
+        self.module_path = module_path
+        self.command_name = command_name
+        self._original_command = None
+        super().__init__(name, callback=self._lazy_callback, **kwargs)
+
+    def _lazy_callback(self, *args, **kwargs):
+        """Load the actual command on first invocation"""
+        if self._original_command is None:
+            try:
+                module = __import__(self.module_path, fromlist=[self.command_name])
```
RohanSreelesh: Could you address this warning from python for `__import__`:

> Import a module.
> 
> Because this function is meant for use by the Python
> interpreter and not for general use, it is better to use
> importlib.import_module() to programmatically import a module.

ref: https://docs.python.org/3/library/importlib.html#importlib.__import__

RohanSreelesh (changes requested):

RohanSreelesh:
Seems like this functionality has broken the ability to pass in arguments. Can reproduce with `sam run configs` and `sam add agent --gui`. Both give the following output:
```Usage: sam add [OPTIONS]
Try 'sam add --help' for help.

Error: No such option: --gui
```
```
Usage: sam run [OPTIONS]
Try 'sam run --help' for help.

Error: Got unexpected extra argument (configs/)
```

m9p909:
I'll try some other techniques and see if I can get it working.

m9p909:
Updated to use the click documentation way of lazy loading commands. Should be more reliable. Can't run the cli with python ./cli/main.py need to use python -m cli.main

m9p909:
It kind of works now. It works well on some commands like "run" but its not enough to be worth it.

PR #343, opened 2025-09-26
fix(DATAGO-111614): Fix trailing '/' in namespace

RZuchniak:
- S3 Artifact Storage now strips any trailing '/' in the namespace
- Fixed an issue with SAM erroring on startup because of the `S3_ENDPOINT_URL` variable
- ~Fixed a problem with the `Boto3` client receiving some extra and mislabeled arguments causing it to fail to initialize~

### DIFF
--- a/src/solace_agent_mesh/agent/adk/services.py
+++ b/src/solace_agent_mesh/agent/adk/services.py
@@ -270,22 +270,24 @@ def initialize_artifact_service(component) -> BaseArtifactService:
 
         try:
             from .artifacts.s3_artifact_service import S3ArtifactService
-            
-            s3_config = {}
 
-            if "region_name" not in s3_config and "region" in config:
-                s3_config["region_name"] = config["region"]
-            
-            for key, value in config.items():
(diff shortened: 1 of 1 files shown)


src/solace_agent_mesh/agent/adk/services.py
```diff
@@ -272,9 +272,12 @@ def initialize_artifact_service(component) -> BaseArtifactService:
             from .artifacts.s3_artifact_service import S3ArtifactService
             
             s3_config = {}
+
+            if "region_name" not in s3_config and "region" in config:
+                s3_config["region_name"] = config["region"]
             
             for key, value in config.items():
-                if key not in ["type", "bucket_name", "artifact_scope"]:
+                if key not in ["type", "bucket_name", "artifact_scope", "base_path", "artifact_scope_value", "region"]:
```
mo-radwan1: Heads up I pushed a fix for it here https://github.com/SolaceLabs/solace-agent-mesh/pull/354 you can remove this change

cyrus2281 (approved):
LGTM

PR #346, opened 2025-09-29
DATAGO-110479: update the maximum capacity of loading artifacts

alimosaed:
**Goal:**
- Increased the maximum and default capacity of loading artifacts to 200K and 100K respectively.
- Added 'message-to-llm' data field to lead the LLM

### DIFF
--- a/src/solace_agent_mesh/agent/sac/app.py
+++ b/src/solace_agent_mesh/agent/sac/app.py
@@ -23,7 +23,7 @@ from ...common.a2a import (
     get_agent_status_subscription_topic,
     get_sam_events_subscription_topic,
 )
-from ...common.constants import DEFAULT_COMMUNICATION_TIMEOUT
+from ...common.constants import DEFAULT_COMMUNICATION_TIMEOUT, TEXT_ARTIFACT_CONTEXT_MAX_LENGTH_CAPACITY
 from ...agent.sac.component import SamAgentComponent
 from ...agent.utils.artifact_helpers import DEFAULT_SCHEMA_MAX_KEYS
 from ...common.utils.pydantic_utils import SamConfigBase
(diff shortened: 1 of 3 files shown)


src/solace_agent_mesh/agent/sac/app.py
```diff
@@ -353,9 +353,9 @@ class SamAgentAppConfig(SamConfigBase):
         description="Configuration for the agent's custom cleanup function.",
     )
     text_artifact_content_max_length: int = Field(
-        default=1000,
+        default=100000,
```
efunneko: Since we are adding one constant here (which is good), both the max and default should be constants in the constants file

src/solace_agent_mesh/agent/utils/artifact_helpers.py
```diff
@@ -819,7 +820,7 @@ async def load_artifact_content_or_metadata(
 
     if max_content_length is None and component:
         max_content_length = component.get_config(
-            "text_artifact_content_max_length", 1000
+            "text_artifact_content_max_length", 100000
```
efunneko: You should check the 'get_config' call. It should always return a value - either the one that is set or the default value. Then we don't need this logic here - just raise a ValueError if it is None

src/solace_agent_mesh/agent/utils/artifact_helpers.py
```diff
@@ -983,8 +985,12 @@ async def load_artifact_content_or_metadata(
                 if is_text:
                     try:
                         content_str = data_bytes.decode(encoding, errors=error_handling)
+                        message_to_llm = ""
                         if len(content_str) > max_content_length:
                             truncated_content = content_str[:max_content_length] + "..."
+                            message_to_llm = f"""This artifact content has been truncated to {max_content_length} characters. 
+                                                The artifact is larger ({len(content_str)} characters).
+                                                Please request again with larger max size up to {TEXT_ARTIFACT_CONTEXT_MAX_LENGTH_CAPACITY} for the full artifact"""
```
efunneko: We should tailor this answer to recognize if the requested size is actually the max size allowed. You pointed this out in slack that the LLM might request again with a larger value only to have us reduce it to the max allowed

src/solace_agent_mesh/common/constants.py
```diff
@@ -1,2 +1,3 @@
 
 DEFAULT_COMMUNICATION_TIMEOUT = 600  # 10 minutes
+TEXT_ARTIFACT_CONTEXT_MAX_LENGTH_CAPACITY = 200000  # maximum number of characters that can be loaded from a text artifact
```
efunneko: Let's put the default value here too

PR #352, opened 2025-10-01
fix(DATAGO-113280): auto add files after precommit hook run

g-raman:
# Description
The prettier command was added to the pre-commit hook to ensure consistent formatting. However, it didn't add the formatting changes to the current commit.

This PR adds the `lint-staged` library to handle that and also formats files that weren't properly formatted before

### DIFF
--- a/client/webui/frontend/package.json
+++ b/client/webui/frontend/package.json
@@ -100,6 +100,7 @@
         "eslint-plugin-react-hooks": "^5.2.0",
         "eslint-plugin-react-refresh": "^0.4.19",
         "globals": "^16.0.0",
+        "lint-staged": "^16.2.3",
         "prettier-plugin-tailwindcss": "^0.6.13",
         "tw-animate-css": "^1.3.3",
         "typescript": "~5.8.3",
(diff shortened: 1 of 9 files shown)


client/webui/frontend/package.json
```diff
@@ -47,9 +47,18 @@
         "build-types": "tsc --emitDeclarationOnly --outDir dist --project tsconfig.lib.json",
         "build-package": "npm run build-lib && npm run build-types",
         "lint": "eslint .",
-        "precommit": "eslint . && tsc -b && prettier --write . --log-level silent",
+        "precommit": "tsc -b && lint-staged",
         "preview": "vite preview"
     },
+    "lint-staged": {
+        "*.{js,jsx,ts,tsx}": [
+            "eslint --fix",
```
ziyanwan, line 55: auto-fix with eslint is newly introduced in this PR, would you please do some test?
what happens when there are errors that cannot be auto-fixed by ESLint? 
Will it fail the commit and report the error to user?

Please paste an screenshot of auto-fix success and auto-fix failure. Thank you.

g-raman: @ziyanwan i've added the screenshots

PR #354, opened 2025-10-01
fix: filter S3 config to only valid boto3 parameters

mo-radwan1:
Changed S3 artifact service initialization to use an inclusion list of valid boto3.client() parameters instead of an exclusion list. This prevents errors when extra config properties (like base_path) are present in the configuration.

Valid boto3 parameters: endpoint_url, aws_access_key_id, aws_secret_access_key, aws_session_token, region_name, config


RohanSreelesh (approved):
LG

PR #355, opened 2025-10-02
chore(DATAGO-112813): Add thumbs up/down user feedback

efunneko:
This PR adds:

- ability for the user to provide feedback for a task with a thumbs up or down, along with a message
- storage of this info in the persistent database that stores session info
- changed the gathering of 'stim' files to be stored in that same database
- provide an endpoint to download stim files given the task id
- add a data retention configuration to automatically clear out tasks/events and feedback after a set period of time
- allow feedback to be published to a configurable topic
- add download button in the workflow to download it as a .stim file


src/solace_agent_mesh/gateway/http_sse/repository/entities/feedback.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Feedback(BaseModel):
+    """Feedback domain entity."""
+
+    id: str
+    session_id: str
+    task_id: str
+    user_id: str
+    rating: str
```
mo-radwan1, line 15: Should rating become an enum with just up and down

src/solace_agent_mesh/gateway/http_sse/dependencies.py
```diff
@@ -425,6 +504,32 @@ def get_session_business_service(
...
+        session_service = SessionService(session_repository, message_repository, component)
+        
+        yield session_service, db
+        db.commit()
+    except Exception:
+        db.rollback()
+        raise
+    finally:
+        db.close()
+
```
mo-radwan1, line 535: Why does the session service need its own non http context ? Are we calling it internally to change anything ?

efunneko: Good question. Thankfully, this is actually dead code now so I removed it

PR #357, opened 2025-10-02
DATAGO-111097 - Agent must not be able to impersonate another agent as part of the call flow

enavitan:
This PR introduces two key enhancements:

1  A new optional construct, agent_identity, has been introduced to manage identity keys. 
2 An interceptor has been added to trigger enterprise-level component initialization. This ensures that additional enterprise features are seamlessly enabled during agent startup without impacting the community edition.

correlated PR is submitted in enterprise.

to test:
```
hatch run pytest tests/unit/agent/sac 
```


src/solace_agent_mesh/agent/sac/component.py
```diff
@@ -275,8 +275,25 @@ def __init__(self, **kwargs):
...
+                from solace_agent_mesh_enterprise.init_enterprise_component import (
+                    init_enterprise_component_features,
+                )
+
+                init_enterprise_component_features(self)
+            except ImportError:
+                # Community edition
+                # Contact Solace support for enterprise features
+                pass
+
```
cyrus2281, line 289: we're already initializing sam enterprise in the cli run command

cyrus2281: src/solace_agent_mesh/common/utils/initializer.py

enavitan: yes originally i thought this is the place which intercepts and executes additional logic per agent, however this one here enables enterprise features before agents go thru init flow, the code above will ensure to execute enterprise init logic per agent.

enavitan:
created alternative PR based on the feature branch.

PR #362, opened 2025-10-06
chore(DATAGO_112813): Rework how we save Chat Messages from the UI

efunneko:
- Remove the old chat_messages table
- Introduce chat_tasks table
- Save the current message bubble state and other task metadata from the frontend to the chat_tasks table
- Load from that table when switching sessions


src/solace_agent_mesh/gateway/http_sse/alembic/versions/20250928_079e06e9b448_add_tasks_task_events_and_feedback_.py
```diff
@@ -1,93 +0,0 @@
...
-    sa.Column('created_time', sa.BigInteger(), nullable=False),
-    sa.PrimaryKeyConstraint('id')
-    )
-    op.create_index(op.f('ix_feedback_task_id'), 'feedback', ['task_id'], unique=False)
-    op.create_index(op.f('ix_feedback_user_id'), 'feedback', ['user_id'], unique=False)
-    op.create_index('ix_feedback_created_time', 'feedback', ['created_time'], unique=False)
-    op.create_table('tasks',
-    sa.Column('id', sa.String(), nullable=False),
-    sa.Column('user_id', sa.String(), nullable=False),
-    sa.Column('start_time', sa.BigInteger(), nullable=False),
```
mo-radwan1: I am not certain about the world of alembic, but I know in spring and liquibase deleting migrations will cause errors in databases that have already run them. We would need to write a migration to either migrate the data from the old table or if we don't care about it, we can just delete the old table

mo-radwan1: Oh I see these are from your feedback branch and not main, ignore my comment

client/webui/frontend/src/lib/components/chat/ChatSessionDialog.tsx
```diff
@@ -14,7 +14,7 @@ export const ChatSessionDialog: React.FC<ChatSessionDialogProps> = ({ isOpen, on
 
     return (
         <Dialog open={isOpen} onOpenChange={onClose}>
-            <DialogContent>
+            <DialogContent showCloseButton={false}>
```
lgh-solace: Not a big deal, but we want to push this into the DialogContent component because they aren't supposed to show for any dialog.

client/webui/frontend/src/lib/components/ui/button.tsx
```diff
@@ -6,7 +6,7 @@ import { cn } from "@/lib/utils";
 
 import { Tooltip, TooltipContent, TooltipTrigger } from "./tooltip";
 
-const commonTextStyles = "text-[var(--color-primary-wMain)] hover:text-[var(--color-primary-text-w60)] dark:text-[var(--color-primary-w20)] dark:hover:text-[var(--color-primary-text-w10)]";
+const commonTextStyles = "text-[var(--color-brand-wMain)] hover:text-[var(--color-primary-text-w60)] dark:text-[var(--color-brand-wMain)] dark:hover:text-[var(--color-primary-text-w10)]";
```
lgh-solace: This change seems odd? Changing the text on the buttons to brand from primary would go against the currently supported palette...  if coming from Jessie, she may have been thinking of the current Solace palette which does use brand on buttons - but the new palette (the one we're using) uses primary.

efunneko: She gave me an example of what she wanted to see with Solace colours on the cancel, but she didn't specifically say to do that. I can revert to primary easily. I guess if we wanted primary to be Solace branded we should just change what colour it is?

src/solace_agent_mesh/gateway/http_sse/alembic/versions/20250928_079e06e9b448_add_tasks_task_events_and_feedback_.py
efunneko: Yup - I figured having multiple scripts for no reason was just messy

client/webui/frontend/src/lib/components/chat/ChatSessionDialog.tsx
efunneko: I set the default to 'false' instead of true

PR #365, opened 2025-10-06
feat(DATAGO-113298): Add component parameter to identity service and local file identity service initialization

cyrus2281:

This change adds the ability to pass a component parameter to identity service implementations, allowing identity service providers to access component resources and configuration when needed. This enhances the flexibility and integration capabilities of identity services within the agent mesh architecture.

## How is this accomplished?

- feat: add component parameter to identity service and local file identity service initialization
- Updated the BaseIdentityService class to accept an optional component parameter
- Modified the create_identity_service factory function to pass the component to service implementations
- Updated the LocalFileIdentityService implementation to support the component parameter
- Added test cases to verify the proper handling of the component parameter

Verify that existing identity service implementations will continue to function correctly with this change as the component parameter is optional.

### DIFF
--- a/src/solace_agent_mesh/common/services/identity_service.py
+++ b/src/solace_agent_mesh/common/services/identity_service.py
@@ -8,13 +8,14 @@ from typing import Any, Dict, List, Optional
 import importlib.metadata as metadata
 
 from solace_ai_connector.common.log import log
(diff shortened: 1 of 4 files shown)


efunneko (approved):
Looks good

PR #371, opened 2025-10-07
DATAGO-112660: agent de registration

alimosaed:
**Goal:**
De-register agents at runtime when they become unavailable.

**Changes:**
This PR introduces three mechanisms for agent de-registration:

**- Discovery Agent De-registration:**
The discovery agent monitors all registered agents by checking their last-seen agent cards within a defined TTL window.
If an agent’s last-seen timestamp exceeds the TTL threshold, it is automatically de-registered.
Once de-registered, the discovery agent no longer recognizes or interacts with that agent.

**- Gateway De-registration:**
Gateways perform a similar check on the last-seen agent cards.
Agents that remain inactive beyond the TTL window are removed from the gateway’s registry.

**- De-registration API:**
A dedicated API is introduced to allow explicit agent de-registration when needed.

### DIFF
--- a/src/solace_agent_mesh/agent/sac/component.py
+++ b/src/solace_agent_mesh/agent/sac/component.py
@@ -3198,6 +3198,16 @@ class SamAgentComponent(SamComponentBase):
             health_check_interval,
             ttl_seconds
         )
+
+        # Validate configuration values
+        if ttl_seconds <= 0 or health_check_interval <= 0 or ttl_seconds < health_check_interval:
+            log.error(
+                "%s agent_health_check_ttl_seconds (%d) and agent_health_check_interval_seconds (%d) must be positive and TTL must be greater than interval.",
+                self.log_identifier,
+                ttl_seconds,
+                health_check_interval
+            )
+            raise ValueError(f"Invalid health check configuration. agent_health_check_ttl_seconds ({ttl_seconds}) and agent_health_check_interval_seconds ({health_check_interval}) must be positive and TTL must be greater than interval.")
         
         # Get all agent names from the registry
         agent_names = self.agent_registry.get_agent_names()
--- a/src/solace_agent_mesh/gateway/http_sse/component.py
+++ b/src/solace_agent_mesh/gateway/http_sse/component.py
@@ -1777,14 +1777,15 @@ class WebUIBackendComponent(BaseGatewayComponent):
             ttl_seconds
         )
         
-        if ttl_seconds <= 0 or health_check_interval <= 0:
+        # Validate configuration values
+        if ttl_seconds <= 0 or health_check_interval <= 0 or ttl_seconds < health_check_interval:
             log.error(
-                "%s agent_health_check_ttl_seconds and agent_health_check_interval_seconds must be positive (TTL=%d, interval=%d).",
+                "%s agent_health_check_ttl_seconds (%d) and agent_health_check_interval_seconds (%d) must be positive and TTL must be greater than interval.",
                 self.log_identifier,
                 ttl_seconds,
                 health_check_interval
             )
-            raise ValueError(f"Invalid health check configuration. agent_health_check_ttl_seconds and agent_health_check_interval_seconds must be positive (TTL={ttl_seconds}, interval={health_check_interval}).")
+            raise ValueError(f"Invalid health check configuration. agent_health_check_ttl_seconds ({ttl_seconds}) and agent_health_check_interval_seconds ({health_check_interval}) must be positive and TTL must be greater than interval.")
         
         # Get all agent names from the registry
         agent_names = self.agent_registry.get_agent_names()


src/solace_agent_mesh/agent/sac/component.py
```diff
@@ -3118,6 +3146,122 @@ def get_gateway_id(self) -> str:
...
+        # De-register unresponsive agents
+        for agent_name in agents_to_deregister:
+            self._deregister_agent(agent_name)
+            
+        log.info(
+            "%s Agent health check completed. Total agents: %d, Unresponsive: %d, De-registered: %d",
+            self.log_identifier,
+            total_agents,
+            unresponsive_agents,
+            len(agents_to_deregister)
```
solace-mdupls: Won't these be the same value? The increment and `append` always coincide in the lines above. Same in the other file.

alimosaed: Good point. Fixed.

src/solace_agent_mesh/agent/sac/component.py
```diff
@@ -127,6 +126,10 @@ def __init__(self, **kwargs):
         super().__init__(info, **kwargs)
         self.agent_name = self.get_config("agent_name")
         log.info("%s Initializing A2A ADK Host Component...", self.log_identifier)
+        
+        # Initialize the agent registry for health tracking
+        from ...common.agent_registry import AgentRegistry
```
cyrus2281: can we have the import at the top of the file?

src/solace_agent_mesh/agent/sac/component.py
```diff
@@ -3118,6 +3146,119 @@ def get_gateway_id(self) -> str:
...
+                    time_since_last_seen,
+                    ttl_seconds
+                )
+                agents_to_deregister.append(agent_name)
+            
+        # De-register unresponsive agents
+        for agent_name in agents_to_deregister:
+            self._deregister_agent(agent_name)
+            
+        log.info(
```
cyrus2281: this should be debug

src/solace_agent_mesh/common/agent_registry.py
```diff
@@ -35,8 +45,79 @@ def get_agent_names(self) -> List[str]:
...
+        Checks if an agent's TTL has expired.
+        
+        Args:
+            agent_name: The name of the agent to check
+            ttl_seconds: The TTL in seconds
+            
+        Returns:
+            A tuple of (is_expired, seconds_since_last_seen)
+        """
+        from solace_ai_connector.common.log import log
```
cyrus2281: move import to top

src/solace_agent_mesh/gateway/http_sse/routers/agent_cards.py
```diff
@@ -39,3 +40,39 @@ async def get_discovered_agent_cards(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
             detail="Internal server error retrieving agent list.",
         )
+        
+@router.delete("/agentCards/{agent_name}", response_model=Dict[str, str])
```
mo-radwan1: Do we need this API ? What does it mean to manually deregister an agent ? Wouldnt the agent come back to the registry once it sends a heart beat ?

alimosaed: It is unnecessary. Deleted the API.

src/solace_agent_mesh/agent/sac/app.py
```diff
@@ -79,6 +79,12 @@ class AgentDiscoveryConfig(SamConfigBase):
     enabled: bool = Field(
         default=True, description="Enable discovery and instruction injection."
     )
+    health_check_ttl_seconds: int = Field(
+        default=HEALTH_CHECK_TTL_SECONDS, description="Time-to-live in seconds after which an unresponsive agent is de-registered."
+    )
+    health_check_interval_seconds: int = Field(
+        default=HEALTH_CHECK_INTERVAL_SECONDS, description="Interval in seconds between health checks."
+    )
```
mo-radwan1, line 87: I don't think we can make both of these configurable. We need one property. Otherwise, the customer could easily start running out of sync. Since they have to go hand in hand. I think ttl should be 3 x health check interval or something like that

alimosaed: @mo-radwan1 Your solution reduces flexibility. Suppose the TTL is big (e.g., 5 minutes), your solution can add unexpected latency. However, I agree that user should get an error when the above parameters are not configured properly. I compared these two parameters to ensure that TTL is always lower than Interval. BTW, it is supposed that these parameters use the default values (TTL=60 sec, Interval=10 sec).

mo-radwan1: I think the TTL should be a parameter that is a multiplication of the interval. It can remain a user defined, but it should be an x number of intervals to wait otherwise it will cause issues / confusion to setting up this parameter and user's won't be able to properly set it up

PR #374, opened 2025-10-08
chore(DATAGO-113861): add empty state component

g-raman:
# Description
Add an empty state component to represent error states

# Example
image


client/webui/frontend/src/assets/illustrations/ErrorDisplay.tsx
```diff
@@ -0,0 +1,62 @@
+interface ErrorDisplayProps {
```
lgh-solace: Could you add an illustration for empty? Perhaps this one? 
https://github.com/SolaceLabs/maas-icons/blob/main/illustrations/couldNotFindIllustration.svg

client/webui/frontend/src/lib/components/common/EmptyState.tsx
```diff
@@ -0,0 +1,41 @@
...
+    text: string;
+    variant: ButtonVariant;
+    onClick: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+}
+
+interface EmptyStateProps {
+    title: string;
+    subtitle?: string;
+    variant?: "error" | "empty";
+    buttons?: ButtonWithCallback[];
```
lgh-solace, line 21: Could you add a prop here to pass in an image?  Any passed in image should take precedence over a default "variant" image.

client/webui/frontend/src/assets/illustrations/ErrorDisplay.tsx
```diff
@@ -0,0 +1,62 @@
...
+    width: number;
+    height: number;
+}
+
+function ErrorDisplay({ width = 24, height = 24 }: ErrorDisplayProps) {
```
lgh-solace: I would like to replace the hard coded hex colours here with variables from the theme... but we might need to play around a bit to see what looks good.  

Can you try these and send me a screenshot off line to see what it looks like?
#FDEDED => var(--color-error-w10)
#86939E => var(--color-border)
#FFFFFF => var(--color-background-w10)

client/webui/frontend/src/lib/components/common/EmptyState.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <div className="flex h-full w-full flex-col items-center justify-center gap-3">
+            {variant === "error" ? <ErrorDisplay width={150} height={150} /> : null}
+
+            <p className="text-2xl">{title}</p>
+            {subtitle ? <p className="text-base">{subtitle}</p> : null}
+
+            <div className="flex gap-2">
+                {buttons &&
+                    buttons.map(({ text, variant, onClick }) => (
+                        <Button variant={variant} onClick={onClick}>
```
lgh-solace: Could you add `testid={text}` and `title={text}` to the buttons?

PR #385, opened 2025-10-10
fix(DATAGO-113595): Add errors to LLM context

RZuchniak:
- If an error is encountered when uploading a file, the error is passed to the llm for a better response
image


cyrus2281 (approved):
LGTM

PR #387, opened 2025-10-10
fix(DATAGO-113763): Remove test selector

Hugo-Pare:
Marking all failing tests as `XFAIL`, they need to be updated/fixed ASAP.

By merging this PR, assuring no more regression.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ gcs = ["google-cloud-storage>=2.0"]
 s3 = ["boto3>=1.35.0,<2.0"]
 vertex = ["google-cloud-aiplatform>=1.38"]
 employee_tools = ["holidays>=0.24"]
-postgresql = ["psycopg2-binary>=2.9.0"]
+postgresql = ["psycopg2-binary==2.9.10"]
 test = [
     "pytest-asyncio",
     "pytest>=8.0.0",


pyproject.toml
```diff
@@ -69,46 +69,32 @@ s3 = ["boto3>=1.35.0,<2.0"]
...
-    "solace-agent-mesh[database,cli,gcs,vertex,employee_tools,postgresql,s3]",
-    "sam-test-infrastructure @ {root:uri}/tests/sam-test-infrastructure"
-]
-
-[tool.hatch.envs.hatch-test]
-parallel = false
-installer = "pip"
-dependencies = [
-    "pytest_asyncio",
+test = [
```
enavitan, line 72: @Hugo-Pare - could you please inculde in README how would one run tests, i know it is somewhat documented on wiki somewhere but having it in git repo would make it a lil more easy to locate/without the need to navigate to external resources + docs tend to easily get outdated if smth changed in the code 
WDYT?

tests/integration/scenarios_declarative/test_declarative_runner.py
```diff
@@ -1569,6 +1576,9 @@ async def mock_save_fail(*args, **kwargs):
     if scenario_id in SKIPPED_FAILING_EMBED_TESTS:
         pytest.skip(f"Skipping failing embed test '{scenario_id}' until fixed.")
 
+    if scenario_id in SKIPPED_PERSISTENCE_TESTS:
```
enavitan, line 1579: i assume there is a follow up jira to fix  these?

Hugo-Pare: Yes, this will be addressed later. I just want the fix out, so we don't experience more regression issues.

PR #388, opened 2025-10-10
chore(DATAGO-114299): use empty state component

g-raman:
# No agents found
Screenshot 2025-10-10 at 11 54 14 AM

# No search results
Screenshot 2025-10-10 at 12 05 19 PM

# Config error
Screenshot 2025-10-10 at 1 14 53 PM

# Error loading agents
Screenshot 2025-10-10 at 2 25 01 PM

# Agents page after margin adjusting
Screenshot 2025-10-10 at 1 20 33 PM

### DIFF
--- a/client/webui/frontend/src/lib/components/pages/AgentMeshPage.tsx
+++ b/client/webui/frontend/src/lib/components/pages/AgentMeshPage.tsx
@@ -21,10 +21,6 @@ export function AgentMeshPage() {
         );
     }
 
-    if (agentsError) {
-        return <EmptyState variant="error" title="Error loading agents" subtitle={agentsError} />;
-    }
-
     const renderLayoutContent = () => {
         if (currentLayout === LayoutType.CARDS) {
             return <AgentMeshCards agents={agents} />;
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/components/agents/AgentMeshCards.tsx
```diff
@@ -20,21 +20,15 @@ export const AgentMeshCards: React.FC<AgentMeshCardsProps> = ({ agents }) => {
...
+                    <input type="text" data-testid="agent-search-input" placeholder="Search..." value={searchQuery} onChange={e => setSearchQuery(e.target.value)} className="bg-background mb-4 rounded-md border px-3 py-2" />
+
                     {filteredAgents.length === 0 && searchQuery ? (
-                        <div className="flex h-[calc(100vh-250px)] flex-col items-center justify-center gap-6">
-                            No agents match your search.
-                            <Button variant="outline" title="Clear Search" onClick={() => setSearchQuery("")}>
-                                Clear Search
-                            </Button>
-                        </div>
+                        <EmptyState title="No agents match your search" variant="noImage" buttons={[{ text: "Clear Search", variant: "outline", onClick: () => setSearchQuery("") }]} />
```
lgh-solace: Could you change the button variant? Just remove it or change it to "default" (same thing). I realize you're keeping it as it was, but I think the default matches the library better, yes?

client/webui/frontend/src/lib/components/pages/AgentMeshPage.tsx
```diff
@@ -22,13 +22,7 @@ export function AgentMeshPage() {
...
 
     if (agentsError) {
-        return (
-            <div className="space-y-6">
-                <div className="flex h-96 items-center justify-center">
-                    <MessageBanner variant="error" message={`Error loading agents. ${agentsError}`} />
-                </div>
-            </div>
-        );
+        return <EmptyState variant="error" title="Error loading agents" subtitle={agentsError} />;
```
lgh-solace: Could you move this down to around line 54 - Just thinking it would be nice if we showed the header when this error occurs. Maybe we can even have a `Refresh Agents` button in this empty error view. See what it looks like?

PR #396, opened 2025-10-14
fix(DATAGO-114441): double scrollbar

g-raman:
# Description
Remove double scrollbar

# Large screen
Screenshot 2025-10-14 at 3 47 57 PM

# Small Screen
Screenshot 2025-10-14 at 3 48 27 PM

### DIFF
--- a/client/webui/frontend/src/lib/components/agents/AgentMeshCards.tsx
+++ b/client/webui/frontend/src/lib/components/agents/AgentMeshCards.tsx
@@ -24,7 +24,7 @@ export const AgentMeshCards: React.FC<AgentMeshCardsProps> = ({ agents }) => {
             {agents.length === 0 ? (
                 <EmptyState variant="noImage" title="No agents found" subtitle="No agents discovered in the current namespace." />
             ) : (
-                <div className="h-full w-full px-12 pt-12">
+                <div className="h-full w-full pt-12 pl-12">
                     <input type="text" data-testid="agent-search-input" placeholder="Search..." value={searchQuery} onChange={e => setSearchQuery(e.target.value)} className="bg-background mb-4 rounded-md border px-3 py-2" />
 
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/components/pages/AgentMeshPage.tsx
```diff
@@ -43,7 +43,7 @@ export function AgentMeshPage() {
             ) : agentsError ? (
                 <EmptyState variant="error" title="Error loading agents" subtitle={agentsError} />
             ) : (
-                <div className={`relative flex-1 p-4 ${currentLayout === LayoutType.CARDS ? "" : "bg-[var(--muted)] dark:bg-[var(--color-bg-wMain)]"}`}>
+                <div className={`relative flex-1 overflow-hidden p-4 ${currentLayout === LayoutType.CARDS ? "" : "bg-[var(--muted)] dark:bg-[var(--color-bg-wMain)]"}`}>
```
lgh-solace: The scrollbar should appear on the side of the view not inside. Can you take a look at improving the fix? You might need to go back to the original code that had margins. Or, you could try having pr-0.

g-raman: i think the scrollbar appeared in the inside container so that the top header stays put. the outline wouldn't be lined up if you could scroll the outside container.

like this:
Screenshot 2025-10-14 at 1 32 35 PM

lgh-solace: Take a look at what the solution was originally before the padding change. I believe the header lined up and the scrollbar appeared in the right location. Thanks!

g-raman: okay, i just went back to how the setup was before, but just removed the hardcoded values

PR #397, opened 2025-10-14
fix(DATAGO-110644): unnecessary re rendering

g-raman:
# Description
Remove unnecessary re-rendering

https://github.com/user-attachments/assets/98f1fc18-95e9-4ac9-87a7-42c2131cf504


client/webui/frontend/src/lib/hooks/useArtifacts.ts
```diff
@@ -49,7 +49,7 @@ export const useArtifacts = (sessionId?: string): UseArtifactsReturn => {
         } finally {
             setIsLoading(false);
         }
-    }, [apiPrefix, sessionId]);
+    }, [apiPrefix, sessionId, setArtifacts, setIsLoading, setError]);
```
cyrus2281: Set functions should never be a part of the dependency array

PR #399, opened 2025-10-15
fix(DATAGO-114489): only prompt confirmation for new chat when persistence is not enabled

g-raman:
# Description
Skip confirmation when starting new chat if persistence is enabled

### DIFF
--- a/client/webui/frontend/src/lib/components/chat/ChatSessionDialog.tsx
+++ b/client/webui/frontend/src/lib/components/chat/ChatSessionDialog.tsx
@@ -1,6 +1,6 @@
 import { useChatContext, useConfigContext } from "@/lib/hooks";
 import { Edit } from "lucide-react";
-import { Dialog, DialogContent, DialogDescription, DialogHeader, DialogTitle, DialogClose, DialogTrigger } from "@/lib/components/ui/dialog";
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/components/chat/ChatSessionDialog.tsx
```diff
@@ -1,36 +1,46 @@
...
+            <Edit className="size-5" />
+            {buttonText}
+        </Button>
+    ) : (
+        <Dialog>
+            <DialogTrigger asChild>
+                <Button variant="ghost" tooltip="Start New Chat Session">
+                    <Edit className="size-5" />
+                    {buttonText}
+                </Button>
```
solace-mdupls: Replace the button with a pure function that gets consumed in both places. Prevents divergence - the button should look the same in all cases, where the only difference is an onClick handler when the dialog is needed.

solace-mdupls (approved):
Let @lgh-solace take a look as well. LGTM

lgh-solace (commented):
Looks good! Please just wait before merging to see if anyone wants to reconsider the removal of this dialog. Thanks!

lgh-solace (approved):

PR #401, opened 2025-10-15
fix(DATAGO-111857): Update temporary_queue configuration to use app_info settings

cyrus2281:

This change allows the `temporary_queue` configuration to be controlled through app_info settings instead of being hardcoded to `True`. This provides more flexibility for queue configuration in both agent and gateway components.

## How is this accomplished?

- fix: Update temporary_queue configuration to use app_info settings
  - Modified the agent and gateway classes to respect the `temporary_queue` setting from the app_info configuration
  - Updated the code to use the value from app_info if present, otherwise default to `True` for backward compatibility

Check if there are any existing deployments that might be affected by this change in queue behavior, though backward compatibility is maintained with the default value.

### DIFF
--- a/docs/docs/documentation/installing-and-configuring/configurations.md
+++ b/docs/docs/documentation/installing-and-configuring/configurations.md
@@ -106,9 +106,9 @@ If you need to configure multiple brokers, you can do so by adding additional en
 :::
 
 :::info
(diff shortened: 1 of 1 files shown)


efunneko (approved):
LG

PR #402, opened 2025-10-15
fix(DATAGO-112612): Cleanup startup/shutdown logs

RZuchniak:
- Initial Claude Code attempt to refactor logs
- Targeted at agent/adk folder

### DIFF
--- a/src/solace_agent_mesh/agent/sac/component.py
+++ b/src/solace_agent_mesh/agent/sac/component.py
@@ -128,7 +128,6 @@ class SamAgentComponent(SamComponentBase):
 
         super().__init__(info, **kwargs)
         self.agent_name = self.get_config("agent_name")
-        log.info("%s Initializing A2A ADK Host Component...", self.log_identifier)
         try:
             self.namespace = self.get_config("namespace")
             if not self.namespace:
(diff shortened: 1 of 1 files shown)


src/solace_agent_mesh/agent/adk/artifacts/filesystem_artifact_service.py
```diff
@@ -45,12 +45,12 @@ def __init__(self, base_path: str):
...
+                "Initialized FilesystemArtifactService: type=filesystem base_path=%s scope_support=session,user",
                 self.base_path,
             )
         except OSError as e:
             logger.error(
-                "Failed to create base directory '%s': %s",
+                "Failed to create artifact base directory at '%s': %s - Check directory permissions and disk space",
                 self.base_path,
                 e,
             )
```
carolmorneau, line 56: We'll need to make the guide clearer around exception handling. This should be updated to something like: 
```
logger.error("Failed to create artifact base directory at '%s': Check directory permissions and disk space",
             self.base_path, exc_info=True)
```

carolmorneau: or better
```
logger.exception("Failed to create artifact base directory at '%s': Check directory permissions and disk space", self.base_path)
```
as `logger.exception()` adds the stracktrace automatically

RZuchniak: I believe it used `exc_info=True` in a lot of places, but it also overwrote some of the `logger.exception()` lines in doing so.

src/solace_agent_mesh/agent/adk/embed_resolving_mcp_toolset.py
```diff
@@ -121,6 +121,7 @@ async def _resolve_embeds_recursively(
                         "%s Failed to resolve embed in string: %s",
                         log_identifier,
                         e,
+                        exc_info=True,
```
carolmorneau, line 124: on the right track but the `e` parameter and it's associated `%s` should be removed OR should be converted to `log.exception()`.

PR #420, opened 2025-10-21
chore(DATAGO-111098): Invoke the trust manager in enterprise for identity signing 

efunneko:
- Adds hooks into SAM community to invoke enterprise if present for:
  - distributing trust information
  - signing user information

- Additionally, small refactor of handling multiple timers within SAM components

### DIFF
--- a/src/solace_agent_mesh/common/a2a/protocol.py
+++ b/src/solace_agent_mesh/common/a2a/protocol.py
@@ -200,17 +200,29 @@ def get_sam_events_subscription_topic(namespace: str, category: str) -> str:
 def get_trust_card_topic(namespace: str, component_type: str, component_id: str) -> str:
     """
     Returns the topic for publishing a Trust Card.
-    
+
+    IMPORTANT: The component_id parameter MUST be the exact broker client-username
+    that the component uses to authenticate with the Solace broker. This is critical
+    for trust verification - trust cards are validated against the actual broker
+    authentication identity.
+
     Args:
         namespace: SAM namespace
         component_type: Type of component ("gateway", "agent", etc.)
-        component_id: Unique component identifier (must match client-username)
-    
+        component_id: MUST be the broker client-username (from broker_username config).
+                     DO NOT use arbitrary IDs like agent_name or gateway_id unless they
+                     match the broker_username exactly.
+
     Returns:
         Topic string: {namespace}/a2a/v1/trust/{component_type}/{component_id}
-    
+
     Raises:
         ValueError: If any parameter is empty
+
+    Security Note:
+        Trust card verification relies on matching the topic component_id with the
+        authenticated broker client-username. Using a different value breaks the
+        security model and trust chain verification.
     """
     if not namespace:
         raise ValueError("Namespace cannot be empty.")
--- a/src/solace_agent_mesh/common/sac/sam_component_base.py
+++ b/src/solace_agent_mesh/common/sac/sam_component_base.py
@@ -324,6 +324,10 @@ class SamComponentBase(ComponentBase, abc.ABC):
         """Late initialization hook called after the component is fully set up."""
 
         # Setup the Trust Manager if present (enterprise feature)
+        # NOTE: The Trust Manager should use component.get_broker_username() to retrieve
+        # the actual broker client-username for trust card topic construction. This is
+        # critical because trust cards MUST be published on topics that match the actual
+        # authentication identity (client-username) used to connect to the broker.
         try:
             from solace_agent_mesh_enterprise.common.trust import (
                 initialize_trust_manager,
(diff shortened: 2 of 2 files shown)


src/solace_agent_mesh/agent/sac/app.py
```diff
@@ -23,14 +23,28 @@
...
 from ...common.utils.pydantic_utils import SamConfigBase
 from ..tools.tool_config_types import AnyToolConfig
 
 log = logging.getLogger(__name__)
 
+# Try to import TrustManagerConfig from enterprise repo
+try:
+    from solace_agent_mesh_enterprise.common.trust.config import TrustManagerConfig
+
+    _TRUST_MANAGER_CONFIG_AVAILABLE = True
```
cyrus2281: This constant is never used

src/solace_agent_mesh/agent/sac/app.py
```diff
@@ -23,14 +23,28 @@
...
 from ...agent.sac.component import SamAgentComponent
 from ...agent.utils.artifact_helpers import DEFAULT_SCHEMA_MAX_KEYS
 from ...common.utils.pydantic_utils import SamConfigBase
 from ..tools.tool_config_types import AnyToolConfig
 
 log = logging.getLogger(__name__)
 
+# Try to import TrustManagerConfig from enterprise repo
+try:
+    from solace_agent_mesh_enterprise.common.trust.config import TrustManagerConfig
```
cyrus2281, line 42: Do we need an import from enterprise only for a type? (which will not be enforced since it is unionized with a generic dictionary)

src/solace_agent_mesh/agent/protocol/event_handlers.py
```diff
@@ -248,6 +267,81 @@ async def handle_a2a_request(component, message: SolaceMessage):
...
+                        log.info(
+                            "%s Successfully authenticated user '%s' for task %s",
+                            component.log_identifier,
+                            verified_user_identity.get("user_id"),
+                            verification_task_id,
+                        )
+
+                except Exception as e:
+                    # Authentication failed - enterprise provides error details
+                    log.warning(
```
gregmeldrum: Should this be an `error` so it stands out in the logs?

gregmeldrum (approved):
Looks good. One comment/question about log level.

src/solace_agent_mesh/common/sac/sam_component_base.py
```diff
@@ -54,8 +56,268 @@ def __init__(self, info: dict[str, Any], **kwargs: Any):
...
+        - TIMER events are routed to registered callbacks
+        - Other events are passed to parent class
+        
+        Args:
+            event: Event object from SAC framework
+        """
+        from solace_ai_connector.common.event import Event, EventType
+        from solace_ai_connector.common.message import Message as SolaceMessage
+        
+        if event.event_type == EventType.MESSAGE:
```
alimosaed, line 142: This does not send an ACK when the message is processed successfully, and so the message stays in the queue.

src/solace_agent_mesh/common/sac/sam_component_base.py
```diff
@@ -54,8 +56,268 @@ def __init__(self, info: dict[str, Any], **kwargs: Any):
...
+            )
+        else:
+            log.error(
+                "%s Async loop not available. Cannot process message on topic: %s",
+                self.log_identifier,
+                topic,
+            )
+            raise RuntimeError("Async loop not available for message processing")
+    
+    def _handle_async_message_completion(
```
alimosaed: [Option] You can optionally improve the coverage of edge cases:

```
def _handle_async_message_completion(
    self,
    future: asyncio.Future,
    topic: str
):
    """Callback to handle completion of async message processing."""
    try:
        if future.cancelled():
            log.warning(
                "%s Message processing for topic %s was cancelled.",
                self.log_identifier,
                topic
            )
        elif future.done():
            exception = future.exception()
            if exception is not None:
                log.error(
                    "%s Message processing for topic %s failed: %s",
                    self.log_identifier,
                    topic,
                    exception,
                    exc_info=exception
                )
            else:
                # Handle successful completion
                try:
                    result = future.result()
                    log.debug(
                        "%s Message processing for topic %s completed successfully.",
                        self.log_identifier,
                        topic
                    )
                    # Optional: Process the result if needed
                    # self._process_successful_result(result, topic)
                except Exception as result_exception:
                    # This catches exceptions that might occur when getting the result
                    log.error(
                        "%s Error retrieving result for topic %s: %s",
                        self.log_identifier,
                        topic,
                        result_exception,
                        exc_info=result_exception
                    )
        else:
            # This case shouldn't normally occur in a completion callback,
            # but it's good defensive programming
            log.warning(
                "%s Future for topic %s is not done in completion handler.",
                self.log_identifier,
                topic
            )
    except Exception as e:
        log.error(
            "%s Error in async message completion handler for topic %s: %s",
            self.log_identifier,
            topic,
            e,
            exc_info=True
        )
```

src/solace_agent_mesh/agent/sac/app.py
efunneko: Good question. I think that this is good as a link to where the full config is defined so that people and AI can find the full definition. I should verify that the validation will be done for the config, but considering it is defined as a Union with a generic Dict, I am not sure that it will be. The intention is that if there is something defined in the config, it would be validated by the pydantic definitions...

src/solace_agent_mesh/common/a2a/protocol.py
```diff
@@ -197,6 +197,72 @@ def get_sam_events_subscription_topic(namespace: str, category: str) -> str:
...
+        topic: Trust card topic
+    
+    Returns:
+        Tuple of (component_type, component_id)
+    
+    Raises:
+        ValueError: If topic format is invalid
+    """
+    parts = topic.split('/')
+    if len(parts) < 6 or parts[1] != 'a2a' or parts[2] != 'v1' or parts[3] != 'trust':
```
alimosaed, line 270: [Option] This validation condition can be problematic when we migrate A2A endpoints to v2. We should manually update version here.

src/solace_agent_mesh/agent/protocol/event_handlers.py
efunneko: In general, I wouldn't say that auth failures are errors, but in this case, it should only fail if someone is doing something malicious or there is a bug. Both deserve errors

src/solace_agent_mesh/common/a2a/protocol.py
efunneko: It is hard to know what to expect from v2 of our topic format. However, when we get there we will have to change quite a few things. 

We don't intend to be forward compatible with newer versions, though. If there is a new version that wants to work with old versions, the newer components will have to speak v1 to older compoenents

PR #430, opened 2025-10-22
feat(DATAGO-112615): Better pydantic error messages

RZuchniak:
- Added a function to the Pydantic BaseModel for formatting config validation error messages
- Uses `app_name` and `agent_name` for identification when available
- Recurses through classes to find description of misconfigured field
- Works for agent and proxy configurations

Example output:
```
---- Configuration validation failed for orchestrator-agent_app ----

   Agent Name: OrchestratorAgent

ERROR 1:
   Error: Input should be a valid boolean, unable to interpret input
   Location: app_config.supports_streaming
   Description: Whether this host supports A2A streaming (tasks/sendSubscribe).

ERROR 2:
   Missing required field: 'type'
   Location: app_config.session_service.type
   Description: Service type (e.g., 'memory', 'sql', 'vertex_rag').

---- Please update your YAML configuration ----
```

### DIFF
--- a/tests/unit/common/utils/test_pydantic_utils.py
+++ b/tests/unit/common/utils/test_pydantic_utils.py
@@ -6,7 +6,6 @@ Tests the validation and formatting of validation errors from Pydantic models.
 
 from solace_agent_mesh.common.utils.pydantic_utils import SamConfigBase
 from pydantic import Field, ValidationError
-from typing import List, Optional
 
 
 class TestPydanticFormatting:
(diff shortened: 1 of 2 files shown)


tests/unit/common/utils/test_pydantic_utils.py
```diff
@@ -0,0 +1,76 @@
...
+        except ValidationError as e:
+            message = DummyModel.format_validation_error_message(e, "TestApp", "TestAgent")
+            assert "TestApp" in message
+            assert "TestAgent" in message
+            assert "app_config.required_field" in message
+            assert "A required field for testing." in message
+            assert "app_config.optional_field" not in message
+            assert "app_config.wrong_type_field" in message
+            assert "An integer field for testing." in message
+            assert "Input should be a valid integer" in message
```
carolmorneau: Since the purpose of the formatting function is to produce a user-friendly error message which is all nicely formatted, I believe you should have a single assert on the whole error message. 

same for the other tests

RZuchniak: I changed it to use a single assertion with the full message, should also help with debugging if anything breaks

carolmorneau:
Looking at the reported SonarQube issues and most of them seem resolvable without too much effort. Could we take a look at them?

PR #442, opened 2025-10-24
fix(DATAGO-115411): move timestamp to next line if model name is too long

g-raman:
# With enough space
Screenshot 2025-10-24 at 3 23 05 PM

# Without enough space
Screenshot 2025-10-24 at 5 08 02 PM

### DIFF
--- a/client/webui/frontend/src/lib/components/activities/VisualizerStepCard.tsx
+++ b/client/webui/frontend/src/lib/components/activities/VisualizerStepCard.tsx
@@ -207,13 +207,13 @@ const VisualizerStepCard: React.FC<VisualizerStepCardProps> = ({ step, isHighlig
 
     return (
         <div className={cardClasses} style={indentationStyle} onClick={onClick}>
-            <div className="mb-1.5 flex items-center">
+            <div className="mb-1.5 flex w-full items-center gap-1">
                 {getStepIcon()}
-                <div className="flex w-full flex-wrap justify-between gap-2">
-                    <h4 className="text-sm font-semibold" title={step.title}>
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/components/activities/VisualizerStepCard.tsx
```diff
@@ -205,16 +205,16 @@ const VisualizerStepCard: React.FC<VisualizerStepCardProps> = ({ step, isHighlig
...
-
     return (
         <div className={cardClasses} style={indentationStyle} onClick={onClick}>
             <div className="mb-1.5 flex items-center">
                 {getStepIcon()}
-                <h4 className="flex-grow text-sm font-semibold text-gray-700 dark:text-gray-200" title={step.title}>
-                    {displayTitle}
-                </h4>
-                <span className="font-mono text-xs text-gray-500 dark:text-gray-400">{displayTimestamp}</span>
+                <div className="flex w-full flex-wrap justify-between gap-2">
```
lgh-solace: 🤔 I'm not sure this solution will work well for agents with names with no spaces (like the agent builder names)  Also, I see this is old code that hasn't be updated for theme changes, could you try this? Please test with long text both with and without spaces and light and dark themes?

```
<div className="mb-1.5 flex items-center">
                {getStepIcon()}
                <h4 className="flex-grow text-sm font-semibold truncate mr-2" title={step.title}>
                    {displayTitle}
                </h4>
                <span className="font-mono text-xs text-muted-foreground text-nowrap">{displayTimestamp}</span>
</div>
```

PR #451, opened 2025-10-27
fix(DATAGO-115433): extract banner colours into reusable variables

g-raman:


client/webui/frontend/src/lib/providers/themes/palettes/solace.ts
```diff
@@ -60,6 +61,7 @@ export const solace: ThemePalette = {
 
     error: {
         w100: "#C33135",
+        "w100-60": "rgba(195, 49, 53, 0.6)",
```
lgh-solace: Sorry for the confusion, but I don't think we want to add anything to this interface as we want to leave it matching the Solace Cloud version as much as possible.

Instead, can we do something more direct? Just create a constants file somewhere logical that exports the group as whole?

Just something like... 
```
export const messageVariants = {
  error: "bg-[var(--color-error-w10)] text-[var(--color-error-wMain)] border-[var(--color-error-wMain)] dark:bg-[var(--color-error-wMain)] dark:text-[var(--color-primary-text-w10)] dark:border-[var(--color-error-w10)]",
... yadda

``` 

Then it could be used something like... 
```
<div className={messageVariants.error} 
```

And, if you could make a new PR to avoid the `feat` bump? Thanks!

PR #452, opened 2025-10-27
chore(DATAGO-115444): disabled state styles for menu and buttons

g-raman:
# Disabled button
https://github.com/user-attachments/assets/d299064f-fdba-494b-8827-a55fc297b820

# Disabled Menu
Screenshot 2025-10-27 at 10 11 04 AM

### DIFF
--- a/client/webui/frontend/src/stories/Menu.stories.tsx
+++ b/client/webui/frontend/src/stories/Menu.stories.tsx
@@ -1,0 +1,47 @@
+import { Menu } from "@/lib/components/ui/menu";
+import type { Meta, StoryContext, StoryFn, StoryObj } from "@storybook/react-vite";
+
+const meta = {
+    title: "Common/Menu",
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/components/ui/menu.tsx
```diff
@@ -65,6 +66,7 @@ const Menu = React.forwardRef<HTMLDivElement, MenuProps>(({ actions, className,
                     <div
                         role="menuitem"
                         tabIndex={0}
+                        data-disabled={action.disabled}
```
lgh-solace, line 69: Do we need to add "enabled" to the hover styles below to prevent them from appearing when the menu item is disabled?

g-raman: from my testing no, it seems to only be an issue with the button component

PR #454, opened 2025-10-27
fix: add default S3 endpoint

RZuchniak:
- Added a fallback endpoint url for the S3 artifact service


mo-radwan1:
Can you verify that this change works for both if you provide the url and if you don't provide the url

RZuchniak:
> Can you verify that this change works for both if you provide the url and if you don't provide the url

I ran it without a provided URL and it connected properly. I also tried with a correct and incorrect URL specified and it worked as expected; with the correct URL working and the invalid one failing. I also ran this change on beta and there were no issues.

PR #458, opened 2025-10-28
fix(DATAGO-115497): Increase DB Connection Pool

Hugo-Pare:
__Problem:__ Under high concurrent load, the application exhausts the default SQLAlchemy database connection pool (size 5, overflow 10). This leads to `sqlalchemy.exc.TimeoutError` exceptions, causing requests to fail with `503 Service Unavailable` and other downstream ASGI errors. This was identified during load testing with 25+ concurrent requests.

__Solution:__ This PR increases the database connection pool size to more robust values, preventing pool exhaustion under load.

- __File Modified__: `.../gateway/http_sse/dependencies.py`
- __Change__: The `create_engine` call in `init_database` is updated to set `pool_size` and `max_overflow`.
- __New Values__: The pool size is increased to 30 with an overflow of 20, allowing the application to handle up to 50 concurrent database connections. These values are defined as constants (`DB_POOL_SIZE`, `DB_MAX_OVERFLOW`) for clarity and ease of configuration.

This change directly resolves the timeout errors and improves the stability and performance of the application under concurrent workloads.

### DIFF
--- a/src/solace_agent_mesh/gateway/http_sse/dependencies.py
+++ b/src/solace_agent_mesh/gateway/http_sse/dependencies.py
@@ -47,9 +47,6 @@ SessionLocal: sessionmaker = None
 
 api_config: dict[str, Any] | None = None
 
-DB_POOL_SIZE = 30
-DB_MAX_OVERFLOW = 20
-
 
 def set_component_instance(component: "WebUIBackendComponent"):
     """Called by the component during its startup to provide its instance."""
@@ -65,11 +62,7 @@ def init_database(database_url: str):
     """Initialize database with direct sessionmaker."""
     global SessionLocal
     if SessionLocal is None:
-        engine = engine = create_engine(
-            database_url,
-            pool_size=DB_POOL_SIZE,
-            max_overflow=DB_MAX_OVERFLOW
-        )
+        engine = create_engine(database_url)
 
         # Enable foreign keys for SQLite only (database-agnostic)
         from sqlalchemy import event


src/solace_agent_mesh/gateway/http_sse/dependencies.py
```diff
@@ -48,6 +47,9 @@ class BaseArtifactService:
 
 api_config: dict[str, Any] | None = None
 
+DB_POOL_SIZE = 30
+DB_MAX_OVERFLOW = 20
```
Hugo-Pare: Should we be able to configure the values in the `webui.yaml` file?

examples/gateways/webui_gateway_example.yaml
```diff
@@ -37,6 +37,7 @@ apps:
         database_url: "${WEB_UI_GATEWAY_DATABASE_URL, sqlite:///webui-gateway.db}"
         default_behavior: "PERSISTENT"
       # --- Optional with Defaults ---
+      model: *general_model # Optional: defaults to None - Points to shared_config.yaml
```
cyrus2281, line 40: Why does a gateway need access to model?

preset/agents/basic/webui.yaml
```diff
@@ -27,6 +27,8 @@ apps:
 
       artifact_service: *default_artifact_service
 
+      model: *general_model
```
cyrus2281, line 30: Same question

src/solace_agent_mesh/gateway/http_sse/app.py
```diff
@@ -89,6 +89,13 @@ class WebUIBackendApp(BaseGatewayApp):
             "default": "",
             "description": "General guidelines on how agent responses should be structured, to be optionally used by agents.",
         },
+        {
+            "name": "model",
+            "required": False,
+            "type": "dict",
+            "default": None,
+            "description": "The model to use for the WebUI gateway.",
+        },
```
cyrus2281, line 98: Why?

PR #483, opened 2025-10-31
fix(DATAGO-115916): disable auth buttons when request is cancelled

g-raman:
# Demo
https://github.com/user-attachments/assets/2e9ee23a-03a3-4011-879f-a9312be88a8a


gregmeldrum (approved):
Look good, thanks!

PR #494, opened 2025-11-03
fix(DATAGO-116362): remove Postgres FTS on chat messages + filter only by title

RohanSreelesh:
This change remove FTS indexes on chat content and only keeps them for session title (Postgres). For Sqlite we continue using ILIKE but only on the session name


src/solace_agent_mesh/gateway/http_sse/repository/session_repository.py
```diff
@@ -237,49 +235,15 @@ def search(
         use_fts = dialect_name == 'postgresql' and len(query.strip()) >= 3
```
mo-radwan1: Why do we need to check the length of the query to more than 3 ? is that the minimum for fts ? I assume so right ? Since it probably doesnt create tokens less than 3 characters

mo-radwan1 (approved):

PR #496, opened 2025-11-04
fix(DATAGO-116370): Handle cache expiry in SAM

gregmeldrum:


src/solace_agent_mesh/common/sac/sam_component_base.py
```diff
@@ -219,6 +219,43 @@ def process_event(self, event):
...
                     timer_id,
                 )
+        elif event.event_type == EventType.CACHE_EXPIRY:  
+            import asyncio
+            import inspect
+
+            cache_data = event.data
+            handler = self.handle_cache_expiry_event
+
+            # Check if the handler is async
```
gregmeldrum, line 229: The base class use a `sync` cache expiry handler and this sub-class uses an `async` handler. Making the base class async is difficult since the async would need to propagate to all calling methods up the call stack. To help contain the fix, instead we check if the handler is sync or async and call the method appropriately.

PR #498, opened 2025-11-04
fix(DATAGO-116301): Convert logging templates to YAML

RZuchniak:
- Converted logging config files in `templates` and `presets` to YAML
- Tests now use YAML for logging
- SAM initializes a default logging config written in YAML

### DIFF
--- a/docs/docs/documentation/deploying/logging.md
+++ b/docs/docs/documentation/deploying/logging.md
@@ -80,7 +80,7 @@ handlers:
 loggers:
   # Keys are logger names used in the application code
   solace_ai_connector:
-    level: ${LOGGING_SOLACE_AI_CONNECTOR_LEVEL, INFO}
+    level: ${LOGGING_SAC_LEVEL, INFO}
     handlers: []
   
   solace_agent_mesh:


preset/logging_config.yaml
```diff
@@ -0,0 +1,43 @@
...
+  solace_agent_mesh:
+    level: INFO
+    handlers: []
+
+  sam_trace:
+    level: INFO
+    handlers: []
+
+root:
+  level: WARNING
```
carolmorneau: `${LOGGING_ROOT_LEVEL, WARNING}`

preset/logging_config.yaml
```diff
@@ -0,0 +1,43 @@
...
+    class: logging.handlers.RotatingFileHandler
+    formatter: simpleFormatter
+    filename: sam.log
+    mode: a
+    maxBytes: 52428800
+    backupCount: 10
+
+loggers:
+  solace_ai_connector:
+    level: INFO
```
carolmorneau: ${LOGGING_SAC_LEVEL, INFO}

preset/logging_config.yaml
```diff
@@ -0,0 +1,43 @@
...
+    handlers: []
+
+  solace_agent_mesh:
+    level: INFO
+    handlers: []
+
+  sam_trace:
+    level: INFO
```
carolmorneau: ${LOGGING_SAM_TRACE_LEVEL, INFO}

preset/logging_config.yaml
```diff
@@ -0,0 +1,43 @@
...
+    maxBytes: 52428800
+    backupCount: 10
+
+loggers:
+  solace_ai_connector:
+    level: INFO
+    handlers: []
+
+  solace_agent_mesh:
+    level: INFO
```
carolmorneau: ${LOGGING_SAM_LEVEL, INFO}

templates/logging_config_template.yaml
```diff
@@ -0,0 +1,43 @@
...
+    class: logging.handlers.RotatingFileHandler
+    formatter: simpleFormatter
+    filename: sam.log
+    mode: a
+    maxBytes: 52428800
+    backupCount: 10
+
+loggers:
+  solace_ai_connector:
+    level: ${LOGGING_SOLACE_AI_CONNECTOR_LEVEL, INFO}
```
carolmorneau: wonder if we should rename this to `LOGGING_SAC_LEVEL` to be consistent with the other variables?

RZuchniak: I think that's a better name, it also helps keep the variables brief

Hugo-Pare (changes requested):
Could we keep the `templates/logging_config_template.ini` and update our tests to run against both `.ini` and `.yaml` files since we will continue to support them?

The `examples` folder should also have examples for both logging configurations.

Hugo-Pare (approved):

PR #512, opened 2025-11-05
fix(DATAGO-116603): Update remote MCP with auth to not use parallel tool calls.

gregmeldrum:
Update the instruction of the remote MCP examples with auth to not use parallel tool calls.


mahtabahmed255 (approved):
LGTM

PR #523, opened 2025-11-07
DATAGO-115239: Register an agent after tool validation

alimosaed:
**Goal:**
- SAM loads agents even if a tool fails. This makes confusion because the agent is up but the tools are unavailable.
- If a MCP server runs locally and the directory is unavailable, the tool fails without an informative log.

**Changes:**
- Reordered the agent registration to ensure that an agent is registered after tool validation.
- Added some exception handling and logs to catch MCP server failures.

**What should reviewers pay attention to?**
- The edge cases


cyrus2281 (approved):
LGTM

PR #526, opened 2025-11-10
fix(DATAGO-116901): Format click logs

RZuchniak:
- Created a default logging config for setup
- Replaced click echo's with log methods
- Ensured that if logging was not configured, the current root handler is removed to allow SAC to configure per config logging

```
2025-11-10 15:34:37,517 - root - INFO - Starting Solace Application Run...
2025-11-10 15:34:37,517 - root - INFO - Loading environment variables from: /Users/robertzuchniak/Code/solace-agent-mesh-enterprise/.env
INFO: Detected YAML format for logging configuration.
2025-11-10 15:34:37,524 | INFO  | MainThread | solace_ai_connector.common.logging_config | Root logger successfully configured based on LOGGING_CONFIG_PATH=/Users/robertzuchniak/Code/solace-agent-mesh-enterprise/configs/logging_config.yaml
2025-11-10 15:34:37,524 | INFO  | MainThread | root | Logging reconfigured from LOGGING_CONFIG_PATH
2025-11-10 15:34:37,530 | INFO  | MainThread | solace_agent_mesh_enterprise.init_enterprise | [EnterpriseInit] Initializing enterprise features...
```

Requires https://github.com/SolaceLabs/solace-ai-connector/pull/103

### DIFF
--- a/cli/commands/run_cmd.py
+++ b/cli/commands/run_cmd.py
@@ -199,7 +199,7 @@ def run(files: tuple[str, ...], skip_files: tuple[str, ...], system_env: bool):
         return 0
 
     file_list = "\n".join(f"  - {cf}" for cf in config_files_to_run)
-    log.warning("Final list of configuration files to run:\n%s", file_list)
+    log.info("Final list of configuration files to run:\n%s", file_list)
 
     if reset_logging:
         for handler in log.handlers[:]:


cli/commands/run_cmd.py
```diff
@@ -78,20 +90,31 @@ def run(files: tuple[str, ...], skip_files: tuple[str, ...], system_env: bool):
...
+            log.info(f"Loaded environment variables from {env_path}")
         else:
-            click.echo(
-                click.style(
-                    "Warning: .env file not found in the current directory or parent directories. Proceeding without loading .env.",
-                    fg="yellow",
-                )
-            )
+            _setup_backup_logger()
+            log.warning("Warning: .env file not found in the current directory or parent directories. Proceeding without loading .env.")
```
carolmorneau: `.env` file is optional. It is possible to set everything as environment variables without an `.env` file. 
For instance, removing `LOGGING_CONFIG_PATH` from `.env` file and doing: 
```
(.venv) ~/repos/solace-agent-mesh/app1 😀  export LOGGING_CONFIG_PATH="configs/logging_config.yaml"
(.venv) ~/repos/solace-agent-mesh/app1 😀  sam run
```
should configure logging via the solace-ai-connector.

carolmorneau: could the call to `configure_from_file()` be done just once and just before the `initialize()` around line 118?
Maybe that would involve delaying this log `log.info(f"Loaded environment variables from {env_path}")`  to line 119

cli/commands/run_cmd.py
```diff
@@ -128,73 +148,71 @@ def run(files: tuple[str, ...], skip_files: tuple[str, ...], system_env: bool):
...
-            click.style(
-                "No configuration files to run after filtering. Exiting.", fg="yellow"
-            )
+        log.warning(
+            "No configuration files to run after filtering. Exiting."
         )
         return 0
 
-    click.echo(click.style("Final list of configuration files to run:", bold=True))
+    log.warning("Final list of configuration files to run:")
```
carolmorneau: this should likely be INFO

carolmorneau: actually, maybe we should combine the log from lines 203 and those from line 205 in one cohesive statement.
We can use `\n` between each config file to maintain readability.

cli/commands/run_cmd.py
RZuchniak: Ah ok, I assumed that the intended behaviour was to have it attempt to setup the logging even without the .env. I'll go ahead and also move the function call

cli/commands/run_cmd.py
```diff
@@ -62,12 +63,23 @@ def run(files: tuple[str, ...], skip_files: tuple[str, ...], system_env: bool):
     This command accepts paths to individual YAML files (`.yaml`, `.yml`) or directories.
     When a directory is provided, it is recursively searched for YAML files.
     """
-    click.echo(click.style("Starting Solace Application Run...", bold=True, fg="blue"))
+    # Set up initial logging to root logger (will be overwritten by LOGGING_CONFIG_PATH if provided)
+    log = logging.getLogger()
+    handler = logging.StreamHandler(sys.stdout)
+    reset_logging = True
```
carolmorneau, line 69: could line 67-69 be moved to `_setup_backup_logger()`? 

We are half-configuring the backup solution and we expect it to be erased if LOGGING_CONFIG_PATH is provided. 

I'd think it'd be simpler to reason about if we do one or the other: LOGGING_CONFIG_PATH or backup

RZuchniak: The problem I was running into is that if I move the `log = logging.getLogger()` line to `setup_backup_logger()`, then `log` becomes inaccessible by the rest of the `run` function.

carolmorneau: What about something like:
```
log = None
if configure_from_file():
   log = logging.getLogger(__name__)
else: 
   log = _setup_backup_logger() # this could return the root logger

```

cli/commands/run_cmd.py
```diff
@@ -128,73 +146,71 @@ def run(files: tuple[str, ...], skip_files: tuple[str, ...], system_env: bool):
...
+        log.warning(
+            "No configuration files to run after filtering. Exiting."
         )
         return 0
 
-    click.echo(click.style("Final list of configuration files to run:", bold=True))
-    for cf_path_str in config_files_to_run:
-            click.echo(f"  - {cf_path_str}")
+    file_list = "\n".join(f"  - {cf}" for cf in config_files_to_run)
+    log.warning("Final list of configuration files to run:\n%s", file_list)
```
carolmorneau: should be INFO

PR #529, opened 2025-11-11
fix(DATAGO-117121): Fix for inline templates

efunneko:
## Summary

Adds inline Liquid template support for dynamic rendering of artifact data (CSV, JSON, YAML) directly in agent responses, with automatic resolution by the gateway.

## Key changes:
  - New «««template: data="file.csv" ...»»» fenced block syntax alongside existing save_artifact
  - Liquid template renderer with data context preparation (CSV → headers/data_rows, JSON → items/object keys)
  - Template resolution in gateway's late embed pipeline (alongside existing artifact_content embeds)
  - Enhanced stream parser to detect/handle nested template blocks inside artifacts
  - Added TemplateBlockData A2A status messages for template metadata propagation
  - Consolidated fenced block syntax rules in agent instructions
  - New dependency: python-liquid==2.1.0

  Testing:
  - Template rendering with JSONPath/limit parameters
  - CSV/JSON/YAML data context handling
  - Nested templates inside save_artifact blocks
  - Automatic template resolution in artifact_return flow

  Behavioral notes:
  - Templates render inline faster than LLM reformatting artifacts manually
  - Late resolution ensures data freshness at display time
  - Nesting depth tracking prevents premature block closure
  
  
Note that the code has maintained its ability to process old style mustache templates in the «artifact_content » embeds. However, it has removed the prompting for the LLM to create this content. There may be artifacts in chat sessions that still contain these old templates

### DIFF
--- a/.github/workflows/ui-ci.yml
+++ b/.github/workflows/ui-ci.yml
@@ -99,6 +99,9 @@ jobs:
       - name: Build UI package
         run: npm run build-package
 
+      - name: Build Storybook
+        run: npm run build-storybook
+
       - name: Run WhiteSource Scan
         id: whitesource_scan
         uses: SolaceDev/solace-public-workflows/.github/actions/whitesource-scan@main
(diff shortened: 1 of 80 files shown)


gregmeldrum (approved):
Looks good!

PR #531, opened 2025-11-11
fix(DATAGO-117013): Migrate to ISO 8601 timestamp

RZuchniak:
- Migrated the default logging_config as well as the docker preset to use the ISO 8601 format
- Added the setting as well as an explanation to the documentation


carolmorneau (approved):
looks good

PR #533, opened 2025-11-12
chore(DATAGO-117265): Add search and replace artifact tool 

efunneko:
The `artifact_search_and_replace_regex` tool is a powerful utility for modifying the content of text-based artifacts. It allows you to find specific text or patterns within a file and replace them with new content.

### Key Capabilities:

1.  **Two Modes of Operation:**

    -   **Literal Search (**`is_regexp=False`**):** This is the simplest mode. It searches for an exact string of text and replaces every instance it finds with your specified replacement text.

    -   **Regular Expression Search (**`is_regexp=True`**):** This is a more advanced and flexible mode. Instead of searching for fixed text, it uses a "regular expression" (regex) to search for patterns. This is useful for finding and changing text that follows a certain structure but isn't always identical. For example, you could find all email addresses, dates in a specific format, or any line that starts with a particular word.

2.  **Advanced Regex Features:**

    -   **Capture Groups:** In regex mode, you can "capture" parts of the text you find and reuse them in the replacement. For example, you could find a full name like `"John Doe"` and use capture groups to replace it with `"Doe, John"`.

    -   **Control Flags:** You can modify the regex behavior with flags:

        -   `g` (global): Replaces all matches in the file, not just the first one.

        -   `i` (case-insensitive): Ignores whether letters are uppercase or lowercase when searching.

        -   `m` (multiline): Allows the search pattern to work across multiple lines.

3.  **Flexible Output Options:**

    -   **New File:** You can save the modified content as a completely new artifact with a different filename.

    -   **New Version:** You can create a new version of the existing artifact, preserving the original version as a historical record.

### Common Use Cases:

-   **Correcting Typos:** Quickly fix a recurring spelling mistake throughout a document.

-   **Reformatting Data:** Change the structure of text, like swapping the order of names or reformatting dates.

-   **Bulk Updates:** Update a product name, version number, or any other piece of information that appears in multiple places.

-   **Code Refactoring:** Perform simple refactoring tasks in code files, like renaming a variable or function.

### DIFF
--- a/tests/unit/agent/tools/test_builtin_artifact_tools.py
+++ b/tests/unit/agent/tools/test_builtin_artifact_tools.py
@@ -972,3 +972,38 @@ class TestArtifactSearchAndReplaceRegex:
 
             # Verify the actual content has literal $ before each number
             assert saved_content == "item,$100\nproduct,$200\nservice,$300"
+
+    @pytest.mark.asyncio
+    async def test_regex_no_matches_with_multiline_flag(self, mock_tool_context):
+        """Test that multiline flag works correctly and reports no matches when pattern doesn't match."""
+        with patch('solace_agent_mesh.agent.tools.builtin_artifact_tools.load_artifact_content_or_metadata') as mock_load, \
+             patch('solace_agent_mesh.agent.tools.builtin_artifact_tools.get_original_session_id') as mock_session, \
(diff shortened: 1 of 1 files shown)


amir-ghasemi (approved):
LGTM.

PR #541, opened 2025-11-13
DATAGO-117317 :  add scopes to submit task + filter out agents based on scopes for webui

RohanSreelesh:
Accompanying enterprise PR:
https://github.com/SolaceDev/solace-agent-mesh-enterprise/pull/320

Fixed the issue by making this role  `agent:{target_agent}:delegate`  also apply to submitting a task. This means that if you don't have the delegate permission you will not be able to use the agent at all. Even if you somehow make it to the submit task function (such as by sending a rest request to /tasks for example, using another gateway etc)


src/solace_agent_mesh/agent/sac/component.py
```diff
@@ -2953,6 +2953,36 @@ def submit_a2a_task(
...
+                "%s Peer delegation to '%s' denied for sub-task %s. Required scopes: %s. Reason: %s",
+                log_identifier_helper,
+                target_agent_name,
+                sub_task_id,
+                required_scopes,
+                reason,
+            )
+            raise PermissionError(
+                f"Delegation to agent '{target_agent_name}' not allowed. Required scopes: {required_scopes}"
+            )
```
enavitan: can we please introduce unit test coverage for this one ? (a happy path + no happy path)
having this in first place would have prevented it to happen. WDYT? 

we should also add a check for the expected error mesage/format so these don't deviate with future works around it

src/solace_agent_mesh/gateway/base/component.py
```diff
@@ -285,6 +285,33 @@ async def submit_a2a_task(
...
+                "%s User '%s' denied access to agent '%s'. Required scopes: %s. Reason: %s",
+                log_id_prefix,
+                user_identity.get("id"),
+                target_agent_name,
+                required_scopes,
+                reason,
+            )
+            raise PermissionError(
+                f"Access denied to agent '{target_agent_name}'. Required scopes: {required_scopes}"
+            )
```
enavitan: aside from validation_context this looks like a duplicate? can we refactor it/reuse ?

enavitan (commented):
looks good, 

left a few comments.

Thanks 
Eugene

src/solace_agent_mesh/agent/sac/component.py
RohanSreelesh: Good idea, added the unit tests thanks

src/solace_agent_mesh/gateway/base/component.py
RohanSreelesh: Refactored into a helper

tests/unit/agent/sac/test_agent_submit_task_rbac.py
```diff
@@ -0,0 +1,102 @@
...
+        message.content = []
+        return message
+
+    @patch("src.solace_agent_mesh.common.utils.rbac_utils.MiddlewareRegistry")
+    def test_submit_a2a_task_calls_validate_operation_config(
+        self, mock_registry, mock_component, sample_a2a_message
+    ):
+        """Test that submit_a2a_task calls config_resolver.validate_operation_config."""
+        mock_config_resolver = Mock()
+        mock_config_resolver.validate_operation_config.return_value = {"valid": True}
```
enavitan, line 50: this looks like we are mocking the result directly rather than letting validation logic to kick in, any chances we can trully engage validator logic?
i would expect component's scopes here to be mocked to smth like:
```
 scopes = [scope1, scope2, scope3..]
# then for happy path a check against a valid scope: i.e say scope3
# unhappy path a check against an invalid scope scope4
```
I understand this might be out of scope of this task since such test would require going deep down in the actual validator so feel free to ignore this comment.

tests/unit/gateway/base/test_gateway_submit_task_rbac.py
```diff
@@ -0,0 +1,110 @@
...
+        """Test that submit_a2a_task calls config_resolver.validate_operation_config."""
+        # Mock for gateway's config resolution
+        mock_gateway_config_resolver = AsyncMock()
+        resolved_user_config = {"_enterprise_capabilities": ["agent:researcher:delegate"]}
+        mock_gateway_config_resolver.resolve_user_config.return_value = resolved_user_config
+        mock_gateway_registry.get_config_resolver.return_value = mock_gateway_config_resolver
+
+        # Mock for RBAC validation
+        mock_rbac_config_resolver = Mock()
+        mock_rbac_config_resolver.validate_operation_config.return_value = {"valid": True}
```
enavitan, line 52: as above, i was hoping we trully test validator by mocking scopes agent has vs mocked scope user has.

tests/unit/agent/sac/test_agent_submit_task_rbac.py
RohanSreelesh: Yeah this can be added but it needs to be added in enterprise. In community validate_operation_config doesn't do anything since there is no RBAC

tests/unit/gateway/base/test_gateway_submit_task_rbac.py
RohanSreelesh: Yes agree this would be integration test in enterprise as mentioned above.

PR #543, opened 2025-11-13
fix(DATAGO-117192): Dual logging issue

RZuchniak:
- ~Upgraded uvicorn log level to CRITICAL to prevent errors from being logged by both the Application and Uvicorn~
- If Uvicorn does write a log, it will now fallback to the same logging format as the `generic_exception_handler()` instead of the built in logging format, preserving log consistency

Example (`log_level` set to ERROR for demonstration and using the jsonFormatter): 

{"timestamp": "2025-11-13T18:16:36.583798+00:00", "levelname": "ERROR", "threadName": "FastAPI_Thread", "name": "solace_agent_mesh.gateway.http_sse.main", "message": "Gene ........ ValueError: Test exception in AuthMiddleware"}
{"timestamp": "2025-11-13T18:16:36.587806+00:00", "levelname": "ERROR", "threadName": "FastAPI_Thread", "name": "uvicorn.error", "message": "Exception in ASGI applic..... ValueError: Test exception in AuthMiddleware"}

### DIFF
--- a/tests/unit/gateway/http_sse/test_uvicorn_config.py
+++ b/tests/unit/gateway/http_sse/test_uvicorn_config.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+'''
+Unit tests for uvicorn configuration in WebUIBackendComponent.
+
+Tests ensure that:
+1. log_config is explicitly set to None
+'''
+
+from unittest.mock import MagicMock, patch
+
+from solace_agent_mesh.gateway.http_sse.component import WebUIBackendComponent
+
+
+class TestUvicornConfig:
+    '''Test suite for uvicorn configuration.'''
+
+    @patch('solace_agent_mesh.gateway.http_sse.component.uvicorn.Server')
+    @patch('solace_agent_mesh.gateway.http_sse.component.uvicorn.Config')
+    @patch('solace_agent_mesh.gateway.http_sse.component.threading.Thread')
+    @patch('solace_agent_mesh.gateway.http_sse.component.TaskLoggerService')
+    def test_log_config_is_none(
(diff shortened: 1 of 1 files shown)


src/solace_agent_mesh/gateway/http_sse/component.py
```diff
@@ -1253,11 +1253,12 @@ def _start_fastapi_server(self):
                 app=self.fastapi_app,
                 host=self.fastapi_host,
                 port=port,
-                log_level="warning",
+                log_level="critical",
```
carolmorneau: Looks like setting `log_config=None` is the solution as it stops the uvicorn server from configuring its own loggers. 

If the above is the case, I believe we should keep this at `warning`. Silencing warnings could hurt troubleshooting.

src/solace_agent_mesh/gateway/http_sse/component.py
```diff
@@ -1253,11 +1253,12 @@ def _start_fastapi_server(self):
...
                 lifespan="on",
                 ssl_keyfile=self.ssl_keyfile,
                 ssl_certfile=self.ssl_certfile,
                 ssl_keyfile_password=self.ssl_keyfile_password,
+                log_config=None
             )
             self.uvicorn_server = uvicorn.Server(config)
```
carolmorneau, line 1263: The oauth2 component also starts a uvicorn server. We should disable its log_config as well
https://github.com/SolaceDev/solace-agent-mesh-enterprise/blob/main/src/components/oauth2_component.py#L190

src/solace_agent_mesh/gateway/http_sse/component.py
RZuchniak: My only concern is that keeping it at `warning` won't solve the issues with duplicate logs. If this is OK given that both are now properly formatted I can make the change.

PR #544, opened 2025-11-13
chore(DATAGO-117202): storybook ci tests

g-raman:
# Description
- Upgrade to latest version of `Vite` and `Storybook`
- Use `Vitest` to run storybook tests
- Run Storybook tests in CI

### DIFF
--- a/client/webui/frontend/src/lib/providers/AudioSettingsProvider.tsx
+++ b/client/webui/frontend/src/lib/providers/AudioSettingsProvider.tsx
@@ -25,7 +25,7 @@ export interface SpeechSettings {
     advancedMode: boolean;
 }
 
-interface AudioSettingsContextValue {
+export interface AudioSettingsContextValue {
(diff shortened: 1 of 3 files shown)


client/webui/frontend/package.json
```diff
@@ -52,7 +52,8 @@
         "precommit": "tsc -b && lint-staged",
         "preview": "vite preview",
         "storybook": "storybook dev -p 6006",
-        "build-storybook": "storybook build"
+        "test:storybook": "vitest --project=storybook",
+        "ci:storybook": "CI=true npm run test:storybook"
```
lgh-solace, line 56: This is different from enterprise versions and usage - can you add a comment to the PR about why this way is chosen here? We should keep them consistent so are you thinking we need to update enterprise as well?

g-raman: Yes the other PR is for enterprise and includes similar changes

PR #549, opened 2025-11-14
chore(DATAGO-115917): use new confirmation dialog

g-raman:


client/webui/frontend/src/lib/components/common/ConfirmationDialog.tsx
```diff
@@ -13,16 +13,28 @@ type ConfirmationDialogProps =
...
           trigger: React.ReactNode;
           triggerText?: never;
+          open?: never;
+          onOpenChange?: never;
+      })
+    | (BaseDialogProps & {
+          trigger?: never;
+          triggerText?: never;
+          open: boolean;
+          onOpenChange: (open: boolean) => void;
```
lgh-solace: Do we need all these variations on the props? Could we simplify them... perhaps use more `?` You could just add a comment about the possible combinations?

client/webui/frontend/src/lib/contexts/UnsavedChangesContext.tsx
```diff
@@ -39,22 +42,22 @@ export const UnsavedChangesProvider: React.FC<{ children: React.ReactNode }> = (
...
             {children}
-            {showDialog && (
-                <ConfirmationDialog
-                    title="Unsaved Changes Will Be Discarded"
-                    message="Leaving the form will discard any unsaved changes. Are you sure you want to leave?"
-                    onClose={handleCancel}
-                    onConfirm={handleConfirm}
-                />
-            )}
+            <ConfirmationDialog
```
lgh-solace, line 45: I am hoping to delete this file... but we will see :D

g-raman: Maybe I can do some refactoring in a follow up PR?

PR #554, opened 2025-11-14
docs: DATAGO-116432 prompt library

amir-ghasemi:
This PR adds documentation for the prompt library.

### DIFF
--- a/docs/docs/documentation/components/prompts.md
+++ b/docs/docs/documentation/components/prompts.md
@@ -19,7 +19,7 @@ The Prompt Library lets you create reusable prompt templates with variables, acc
 
 3. **Version Management**: Track changes to your prompts over time with automatic versioning, and designate specific versions as "Active" for active use.
 
-4. **Organization**: Categorize prompts by topic (Development, Analysis, Documentation, etc.) and assign custom chat shortcuts for even faster access.
+4. **Organization**: Tag prompts by topic (Development, Analysis, Documentation, etc.) and assign custom chat shortcuts for even faster access.
 
 5. **AI-Assisted Creation**: Use the built-in prompt builder assistant to help you design effective templates through natural conversation.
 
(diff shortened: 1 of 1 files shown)


docs/docs/documentation/components/prompts.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Understanding Prompts
+
+### Prompt Groups
+
+A prompt group is a container that holds multiple versions of a prompt template. Each group includes:
+
+- **Name**: A descriptive title for the prompt (e.g., "Code Review Template")
+- **Description**: Optional details about the prompt's purpose
+- **Category**: Organizational label (e.g., "Development", "Analysis", "Documentation")
```
solace-mdupls: Should this be "Tags" or is the UI that I'm looking at out-of-date?

docs/docs/documentation/components/prompts.md
```diff
@@ -0,0 +1,189 @@
...
+- **Shortcut**: Optional shortcut for quick access (e.g., typing `/review` instantly finds your code review prompt)
```
solace-mdupls, line 35: Minor - the UI calls this "Chat shortcut" I actually prefer just "Shortcut" because we may expose this via other actions.

ie. I have an affinity to Slack's cmd-k shortcut to quickly find what I'm looking for. So actions don't have to originate from a "chat".

solace-mdupls (approved):
Looks good! Some tweaks to consider.

docs/docs/documentation/components/prompts.md
amir-ghasemi: updated

docs/docs/documentation/components/prompts.md
amir-ghasemi: that was a UX suggestion but open to updating it in the UI later.

PR #555, opened 2025-11-14
feat DATAGO-117595: Message actions button

amir-ghasemi:


src/solace_agent_mesh/gateway/http_sse/dependencies.py
```diff
@@ -617,3 +618,29 @@ def get_session_business_service_optional(
         )
         return None
     return SessionService(component=component)
+
+
+def get_audio_service(
+    component: "WebUIBackendComponent" = Depends(get_sac_component),
+) -> AudioService:
+    """FastAPI dependency to get an instance of AudioService."""
+    log.debug("get_audio_service called")
```
RohanSreelesh: Log is not using a prefix, the logs after are using one

src/solace_agent_mesh/gateway/http_sse/dependencies.py
```diff
@@ -617,3 +618,29 @@ def get_session_business_service_optional(
...
+    # AudioService expects app_config which contains the speech configuration
+    app_config = component.component_config.get('app_config', {}) if hasattr(component, 'component_config') else {}
+    log.debug(f"[get_audio_service] app_config keys: {app_config.keys()}")
+    return AudioService(config=app_config)
+
+
+def get_current_user(request: Request) -> dict:
```
RohanSreelesh: This function already exists in `src/solace_agent_mesh/gateway/http_sse/shared/auth_utils.py`

src/solace_agent_mesh/gateway/http_sse/main.py
```diff
@@ -620,6 +621,7 @@ def _setup_routers() -> None:
     app.include_router(projects.router, prefix=api_prefix, tags=["Projects"])
     app.include_router(feedback.router, prefix=api_prefix, tags=["Feedback"])
     app.include_router(prompts.router, prefix=f"{api_prefix}/prompts", tags=["Prompts"])
+    app.include_router(speech.router, prefix="/api/speech", tags=["Speech"])
```
RohanSreelesh: prefix hardcoded please use `{api_prefix}`

src/solace_agent_mesh/gateway/http_sse/routers/config.py
```diff
@@ -195,6 +195,65 @@ async def get_app_config(
         else:
             log.debug("%s Projects feature flag is disabled.", log_prefix)
 
+        
+        # Check tool configuration status
+        tool_config_status = {}
+        
+        # Check TTS configuration (Gemini or Azure)
+        gemini_key = os.getenv("GEMINI_API_KEY")
```
RohanSreelesh: These should be gotten through the component config not env directly

src/solace_agent_mesh/gateway/http_sse/routers/config.py
```diff
@@ -195,6 +195,65 @@ async def get_app_config(
...
+        
+        # Check for explicit TTS provider preference via environment variable
+        # TTS_PROVIDER can be "gemini" or "azure"
+        preferred_provider = os.getenv("TTS_PROVIDER", "").lower()
+        
+        # Determine which providers are available
+        gemini_available = bool(gemini_key)
+        azure_available = bool(azure_speech_key and azure_speech_region)
+        polly_available = bool(aws_access_key and aws_secret_key)
+        tts_configured = gemini_available or azure_available or polly_available
```
RohanSreelesh: This will probably be covered once the keys are read in the from the config. But I wonder if we should also add another check for the speech section to be defined in the webui config. This way if the config is not given the feature will be disabled even if keys are present.

src/solace_agent_mesh/gateway/http_sse/routes/speech.py
```diff
@@ -0,0 +1,341 @@
...
+
+router = APIRouter()
+
+
+class TTSRequest(BaseModel):
+    """Request model for text-to-speech"""
+    input: str
+    voice: Optional[str] = None
+    messageId: Optional[str] = None
+    provider: Optional[str] = None  # NEW: Allow provider selection from UI
```
RohanSreelesh: Please remove this comment

src/solace_agent_mesh/gateway/http_sse/routes/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
```
RohanSreelesh: Any reason for this new directory? Should the speech endpoint just live with the other routers in /routers?

src/solace_agent_mesh/gateway/http_sse/services/audio_service.py
```diff
@@ -0,0 +1,1154 @@
+"""
```
RohanSreelesh, line 1: Curious if we could use LiteLLM SDK for some parts of this code instead of constructing API calls by ourselves?

src/solace_agent_mesh/gateway/http_sse/routers/speech.py
```diff
@@ -0,0 +1,341 @@
...
+        HTTPException: If generation fails
+    """
+    log.info(
+        "[SpeechAPI] TTS stream request from user=%s, text_len=%d",
+        user.get("user_id"),
+        len(request.input)
+    )
+    
+    try:
+        # Validate input
```
RohanSreelesh, line 183: Should we validate for a MAX input as well?

src/solace_agent_mesh/gateway/http_sse/services/audio_service.py
```diff
@@ -0,0 +1,1154 @@
...
+            HTTPException: If transcription fails
+        """
+        
+        try:
+            # Validate file
+            if not audio_file.filename:
+                raise HTTPException(400, "No filename provided")
+            
+            # Check file size (max 25MB)
+            content = await audio_file.read()
```
RohanSreelesh, line 406: I think we should check the file size before reading it in completely or else could cause oom potentially if the file is massive?

src/solace_agent_mesh/gateway/http_sse/services/audio_service.py
```diff
@@ -0,0 +1,1154 @@
...
+        
+        log.debug("[AudioService] Speech config: %s", config.keys())
+        return config
+    
+    def _get_file_extension(self, filename: str) -> str:
+        """Get file extension from filename"""
+        import os
+        return os.path.splitext(filename)[1] or ".wav"
+    
+    async def _mock_save_artifact(self, *args, **kwargs):
```
RohanSreelesh: should be removed?

src/solace_agent_mesh/gateway/http_sse/services/audio_service.py
```diff
@@ -0,0 +1,1154 @@
...
+            .replace('"', "&quot;")
+            .replace("'", "&apos;"))
+        
+        # For HD voices, the format is "locale-Name:DragonHDLatestNeural"
+        # Azure expects just the voice name in SSML, not the :DragonHDLatestNeural suffix
+        # The HD quality is specified via the voice name itself
+        voice_name = voice  # Use the full voice name as-is for HD voices
+        
+        return f"""<speak version="1.0" xmlns="http://www.w3.org/2001/10/synthesis" xml:lang="en-US">
+    <voice name="{voice_name}">
```
RohanSreelesh: Should we escape the voice name as well to avoid injection?

templates/webui.yaml
```diff
@@ -51,3 +51,128 @@ apps:
...
+        stt:
+          provider: azure
+          azure:
+            api_key: "${AZURE_SPEECH_KEY}"
+            region: "${AZURE_SPEECH_REGION}"
+            language: en-US
+        
+        # Text-to-Speech Configuration
+        tts:
+          provider: ${TTS_PROVIDER, gemini}  # "gemini" or "azure" (can be overridden by UI)
```
RohanSreelesh: Small idea but if we take in provider anyway should we directly take the config under it, instead of having azure twice for example like 
```
tts:
  provider: "azure
     azure:
        config...
```

RohanSreelesh: I do recognize that they all require different configuration based on the provider so maybe its better to be explicit

src/solace_agent_mesh/gateway/http_sse/routers/config.py
amir-ghasemi: Fixed, thanks!

src/solace_agent_mesh/gateway/http_sse/main.py
amir-ghasemi: Good catch!

src/solace_agent_mesh/gateway/http_sse/dependencies.py
amir-ghasemi: Done.

src/solace_agent_mesh/gateway/http_sse/dependencies.py
amir-ghasemi: Fixed.

src/solace_agent_mesh/gateway/http_sse/services/audio_service.py
amir-ghasemi: Removed

src/solace_agent_mesh/gateway/http_sse/services/audio_service.py
amir-ghasemi: good idea. done.

src/solace_agent_mesh/gateway/http_sse/services/audio_service.py
amir-ghasemi: good idea, done

src/solace_agent_mesh/gateway/http_sse/routers/speech.py
amir-ghasemi: done

src/solace_agent_mesh/gateway/http_sse/routes/speech.py
amir-ghasemi: done

src/solace_agent_mesh/gateway/http_sse/routes/__init__.py
amir-ghasemi: consolidated

src/solace_agent_mesh/gateway/http_sse/routers/config.py
amir-ghasemi: good idea, added the extra config check upfront

templates/webui.yaml
amir-ghasemi: yes, we should be explicit and also the users might want to configure multiple providers simultaneously and be able to switch via the UI.

src/solace_agent_mesh/gateway/http_sse/services/audio_service.py
amir-ghasemi: in principle, yes, for the simpler stuff, but using provider-specific SDKs offers more choice down the road if we want to enhance the speech feature further.

client/webui/frontend/src/lib/components/chat/MessageHoverButtons.tsx
```diff
@@ -0,0 +1,109 @@
...
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
+      // Check for Ctrl+Shift+C
+      if (event.ctrlKey && event.shiftKey && event.key.toLowerCase() === 'c') {
+        event.preventDefault();
+        handleCopy();
+        
+        // Flash the button to provide visual feedback
+        if (buttonRef.current) {
+          buttonRef.current.classList.add('bg-gray-100', 'dark:bg-gray-700');
```
lgh-solace: As a general comment, we try to stick to the Solace palette of colours. Every colour should try to reference one of the variables. Just guessing but this would be something like: `bg-[var(--color-secondary-w20)] dark:bg-[var(--color-secondary-w80)]`

client/webui/frontend/src/lib/components/chat/MessageHoverButtons.tsx
```diff
@@ -0,0 +1,109 @@
...
+    )}>
+      {/* TTS Button - for AI messages */}
+      {!message.isUser && <TTSButton message={message} />}
+
+      {/* Copy button - all messages */}
+      <Button
+        ref={buttonRef}
+        variant="ghost"
+        size="icon"
+        className="h-8 w-8 hover:bg-gray-100 dark:hover:bg-gray-800"
```
lgh-solace: Should not need to add hover styles with the standard `Button` component?

client/webui/frontend/src/lib/components/chat/MessageHoverButtons.tsx
```diff
@@ -0,0 +1,109 @@
...
+        onClick={handleCopy}
+        tooltip={isCopied ? "Copied!" : "Copy to clipboard"}
+      >
+        {isCopied ? (
+          <Check className="h-4 w-4 text-green-500" />
```
lgh-solace: `text-[var(--color-success-wMain)]`

client/webui/frontend/src/lib/components/navigation/NavigationList.tsx
```diff
@@ -30,11 +30,15 @@ export const NavigationList: React.FC<NavigationListProps> = ({ items, bottomIte
             {/* Bottom items */}
             {bottomItems && bottomItems.length > 0 && (
                 <ul className="space-y-1">
-                    {bottomItems.map(item => (
+                    {bottomItems.filter(item => item.id !== "theme-toggle").map(item => (
```
lgh-solace: Could we remove the `theme-toggle` from the bottomItems instead of skipping it here?

client/webui/frontend/src/lib/components/settings/SpeechSettings.tsx
```diff
@@ -0,0 +1,560 @@
...
+                                Azure Speech {sttConfigured === false ? "(Not Configured)" : ""}
+                            </SelectItem>
+                        </SelectContent>
+                    </Select>
+                </div>
+
+                {/* STT Configuration Warning - Only show for External API */}
+                {settings.speechToText && settings.sttProvider !== "browser" && sttConfigured === false && (
+                    <div className="rounded-md bg-yellow-50 dark:bg-yellow-950/20 p-3 border border-yellow-200 dark:border-yellow-900">
+                        <div className="flex gap-2">
```
lgh-solace, line 213: Looks like a `MessageBanner` of the `warning` variant?

client/webui/frontend/src/lib/components/settings/SpeechSettings.tsx
```diff
@@ -0,0 +1,560 @@
...
+                                </p>
+                            </div>
+                        </div>
+                    </div>
+                )}
+
+                {/* Browser STT Info */}
+                {settings.speechToText && settings.sttProvider === "browser" && (
+                    <div className="rounded-md bg-blue-50 dark:bg-blue-950/20 p-3 border border-blue-200 dark:border-blue-900">
+                        <div className="flex gap-2">
```
lgh-solace: Looks like a `MessageBanner` of the `info` variant? (note: there's no title in the message banner so it would need to be single string, the icon and styling is all there).

client/webui/frontend/src/lib/components/ui/switch.tsx
```diff
@@ -0,0 +1,40 @@
...
+    return (
+        <button
+            type="button"
+            role="switch"
+            aria-checked={checked}
+            disabled={disabled}
+            onClick={handleClick}
+            className={cn(
+                "peer inline-flex h-6 w-11 shrink-0 cursor-pointer items-center rounded-full border-2 border-transparent transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 focus-visible:ring-offset-background disabled:cursor-not-allowed disabled:opacity-50",
+                checked ? "bg-primary" : "bg-input",
```
lgh-solace: I might adjust the unchecked background to be something like `bg-primary/50` just because it looks a bit more like disabled rather than inactive.

client/webui/frontend/src/lib/components/settings/SpeechSettings.tsx
amir-ghasemi: this should not render any more as we reset to browser and show the browser info instead.

client/webui/frontend/src/lib/components/navigation/NavigationList.tsx
amir-ghasemi: done

client/webui/frontend/src/lib/components/chat/MessageHoverButtons.tsx
amir-ghasemi: done

PR #558, opened 2025-11-17
DATAGO-117236: Limit upload file size

alimosaed:
**Goal:**
- Uploading large files fails.
- LLM is unaware of upload failures and therefore it responds with a misleading message.

**Changes:**
- Updated the front-end to call upload API with proper format.
- Updated the backend to handle large file uploads corresponding to front-end.
- Checked the file size limit in backend.
- Added the error to the LLM context to generate an informative response.

**What should reviewers consider?**
- The backend checks the file size using the header and the content size. Maybe the content size verification is duplicate.
- Is there any edge case?

### DIFF
--- a/tests/unit/gateway/http_sse/routers/test_artifacts.py
+++ b/tests/unit/gateway/http_sse/routers/test_artifacts.py
@@ -102,15 +102,23 @@ class TestUploadArtifactWithSession:
     @pytest.fixture
     def mock_dependencies(self):
         """Create mock dependencies for upload tests."""
-        # Mock FastAPI Request
+        # Mock FastAPI Request - don't include Content-Length to avoid early size check
         mock_request = MagicMock()
-        mock_request.headers = {}
+        mock_request.headers = {}  # No Content-Length header
         
-        # Mock UploadFile
+        # Mock UploadFile with chunked reading simulation
         mock_upload_file = MagicMock(spec=UploadFile)
         mock_upload_file.filename = "test.txt"
         mock_upload_file.content_type = "text/plain"
-        mock_upload_file.read = AsyncMock(return_value=b"test content")
+        
+        # Simulate chunked reading: first chunk returns content, second returns empty (EOF)
+        async def mock_read_chunks(size=-1):
+            if not hasattr(mock_read_chunks, 'called'):
+                mock_read_chunks.called = True
+                return b"test content"
+            return b""  # EOF
+        
+        mock_upload_file.read = mock_read_chunks
         mock_upload_file.close = AsyncMock()
         
         # Mock artifact service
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/providers/ChatProvider.tsx
```diff
@@ -1651,6 +1686,13 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
                 if (currentInput) {
                     messageParts.push({ kind: "text", text: currentInput });
                 }
+                
+                // Add file upload errors as text parts so LLM knows what happened
```
efunneko: Q: Maybe we should stop the chat input being posted to the backend if the file upload fails? It seems a bit weird to know that the upload fails and still go ahead with the request, but tell the LLM that the upload failed...

src/solace_agent_mesh/gateway/http_sse/routers/artifacts.py
```diff
@@ -290,8 +290,47 @@ async def upload_artifact_with_session(
     )
 
     try:
-        # Read and validate file content
+        # ===== VALIDATE FILE SIZE BEFORE READING =====
+        max_upload_size = component.get_config("gateway_max_upload_size_bytes", 52428800)
```
efunneko: This is the 3rd place where we have the number 52528800. Can you check to see if we need anything here? I think that get_config will automatically fill in the default value if there isn't one specified, which means we shouldn't have a default value here or above

client/webui/frontend/src/lib/providers/ChatProvider.tsx
alimosaed: Makes sense. I check if we can pause the process without side-effects.

client/webui/frontend/src/lib/providers/ChatProvider.tsx
```diff
@@ -342,27 +342,56 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
...
+                    if (response.status === 413) {
+                        // Extract file size information if available
+                        const actualSize = errorData.actual_size_bytes;
+                        const maxSize = errorData.max_size_bytes;
+                        
+                        let errorMessage;
+                        if (actualSize && maxSize) {
+                            const actualSizeMB = (actualSize / (1024 * 1024)).toFixed(2);
+                            const maxSizeMB = (maxSize / (1024 * 1024)).toFixed(2);
+                            errorMessage = `File "${file.name}" is too large: ${actualSizeMB} MB exceeds the maximum allowed size of ${maxSizeMB} MB`;
```
ghaithdallaali, line 379: Thank you for sending a human-readable error message back to the user!

src/solace_agent_mesh/gateway/http_sse/routers/artifacts.py
```diff
@@ -290,8 +290,47 @@ async def upload_artifact_with_session(
...
+                    
+                    raise HTTPException(
+                        status_code=status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
+                        detail=error_msg  # Use string instead of dict
+                    )
+            except ValueError:
+                log.warning("%s Invalid Content-Length header: %s", log_prefix, content_length)
+        
+        # Read file content
         content_bytes = await upload_file.read()
```
enavitan: prob premature optimization but we are loading entire file in mem here to later confirm its size, 
we should prob read data in chunks and abrupt further processing if we crossed 50M

```
chunk = await upload_file.read(chunk_size)
        if not chunk:
            break
        total += len(chunk)
        if total > max_upload_size:
           err
```

additionally perhaps worth storing data in tmp file to avoid loading entire 50M at once ?

alimosaed: Technically, the previous condition should discard files larger than 50MB, but I loaded chunk by chunk.

src/solace_agent_mesh/gateway/http_sse/routers/artifacts.py
alimosaed: It uses the default value. So, I removed the value from here.

client/webui/frontend/src/lib/providers/ChatProvider.tsx
alimosaed: I updated the code to remove uploaded files when a file fails. Also, it does not send the request to LLM if uploading fails.

PR #560, opened 2025-11-17
fix(DATAGO-117709): use v4 for uuid generation instaed of crypto

TamimiGitHub:


client/webui/frontend/src/lib/providers/ChatProvider.tsx
Hugo-Pare: Could you update this to use `v4()`

Hugo-Pare: Same thing here

Hugo-Pare (changes requested):
Looks good, I think you missed two places

client/webui/frontend/src/lib/providers/ChatProvider.tsx
TamimiGitHub: oh slipped

PR #568, opened 2025-11-19
fix(DATAGO-116432): saving new prompt version when only metadata changes

amir-ghasemi:
also fixes DATAGO-117982

### DIFF
--- a/client/webui/frontend/src/lib/components/prompts/hooks/usePromptTemplateBuilder.ts
+++ b/client/webui/frontend/src/lib/components/prompts/hooks/usePromptTemplateBuilder.ts
@@ -266,16 +266,6 @@ export function usePromptTemplateBuilder(editingGroup?: PromptGroup | null) {
                             addNotification(errorMessage, "error");
                             throw new Error(errorMessage);
                         }
-
-                        // If we just overwrote a non-active version, optionally make it active
-                        // Only do this if we're editing the currently active version
-                        if (isEditingActiveVersion) {
-                            // The active version was updated, no need to change production_prompt_id
-                            // It's already pointing to the correct prompt
-                        } else {
-                            // We overwrote an older version - it stays inactive
-                            // No need to update production_prompt_id
-                        }
                     }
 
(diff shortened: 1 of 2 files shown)


client/webui/frontend/src/lib/components/prompts/hooks/usePromptTemplateBuilder.ts
```diff
@@ -129,183 +129,182 @@ export function usePromptTemplateBuilder(editingGroup?: PromptGroup | null) {
...
+                        if (!promptResponse.ok) {
+                            const error = await promptResponse.json();
+                            const errorMessage = error.message || error.detail || "Failed to update prompt text";
+                            addNotification(errorMessage, "error");
+                            throw new Error(errorMessage);
+                        }
+
+                        // If we just overwrote a non-active version, optionally make it active
+                        // Only do this if we're editing the currently active version
+                        if (isEditingActiveVersion) {
```
lgh-solace: This if/else does nothing? Can we remove it?

src/solace_agent_mesh/gateway/http_sse/routers/prompts.py
```diff
@@ -842,6 +843,57 @@ async def create_prompt_version(
         )
 
 
+@router.patch("/{prompt_id}", response_model=PromptResponse)
+async def update_prompt(
```
lgh-solace, line 847: This new function is for updating the prompt only?

client/webui/frontend/src/lib/components/prompts/hooks/usePromptTemplateBuilder.ts
amir-ghasemi: removed, thanks.

src/solace_agent_mesh/gateway/http_sse/routers/prompts.py
amir-ghasemi: Yes, the new function is specifically for updating an existing prompt's content only and overwriting it.

PR #571, opened 2025-11-19
feat(DATAGO-117185): Add an option to suppress warning logs

RZuchniak:
- When running SAM some warning logs are printed from various libraries that to not align with the configured formatting
- To prevent these warnings from showing up in Datadog, a flag can be used to disable them while still giving developers full exposure to any problems.
- A choice has to be made between preventing all errors, or only a selected few and updating the list as fixes are made or new errors appear

Example warnings:
`RuntimeWarning: Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work
  warn("Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work", RuntimeWarning)`
  
  `FutureWarning: Support for google-cloud-storage < 3.0.0 will be removed in a future version of google-cloud-aiplatform.`

Usage example:
Running `sam run --suppress-warnings` will mute the warnings above

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -29,7 +29,7 @@ from cli.commands.docs_cmd import docs
     is_flag=True,
     expose_value=False,
     is_eager=True,
-    help='Suppress deprecation warnings from dependencies.'
+    help="Suppress warnings emitted by Python's warnings module."
 )
 def cli():
     """Solace CLI Application"""


carolmorneau:
Question: with `import warnings` is there a way to redirect warnings to file rather than silencing completely?

RZuchniak:
> Question: with `import warnings` is there a way to redirect warnings to file rather than silencing completely?

Yes, it looks like you can use a custom `showwarning` function to override printing to stderr: https://docs.python.org/3/library/warnings.html#warnings.showwarning

carolmorneau:
> > Question: with `import warnings` is there a way to redirect warnings to file rather than silencing completely?
> 
> Yes, it looks like you can use a custom `showwarning` function to override printing to stderr: https://docs.python.org/3/library/warnings.html#warnings.showwarning

Thank you for looking this up. 

I believe, the `--suppress-warnings` flag is good as you have it. It targets all warnings which is the purpose. Users have to explicitly provide the flag in environments where they want to silence warnings.

cli/main.py
```diff
@@ -18,6 +24,13 @@
 @click.version_option(
     __version__, "-v", "--version", help="Show the CLI version and exit."
 )
+@click.option(
+    '--suppress-warnings',
+    is_flag=True,
+    expose_value=False,
+    is_eager=True,
+    help='Suppress deprecation warnings from dependencies.'
```
carolmorneau: This flag also silences warnings coming from the main application, not just dependencies. 
What about: 

> Suppress warnings emitted by Python\'s warnings module

PR #574, opened 2025-11-20
feat(DATAGO-117225): Web Search and Deep Research

amir-ghasemi:
This pull request introduces several improvements and new features to the frontend codebase, focusing on enhanced support for research-related content, improved citation handling, and dependency updates. The most notable changes include support for image search results in activity visualization, improved citation parsing and display in chat messages, and updates to dependencies for better compatibility and new features.

### Research and Citation Features

* Added support for displaying image search results in activity visualizations by introducing the `ImageSearchGrid` component and updating the tool result rendering logic in `VisualizerStepCard.tsx`. This allows users to see image results directly in the activity step cards. [[1]](diffhunk://#diff-b4fc0eabe020ee6e1753df0f227923a18fbbcc1e93ff18cbb83f9ad8ba2801b6L146-R203) [[2]](diffhunk://#diff-b4fc0eabe020ee6e1753df0f227923a18fbbcc1e93ff18cbb83f9ad8ba2801b6R6)
* Improved citation parsing and display in chat messages by integrating citation extraction and rendering logic. Citations are now parsed from message text and linked to relevant RAG (Retrieval-Augmented Generation) sources, with clickable citations that open the RAG panel for more information. [[1]](diffhunk://#diff-f9b9a25a941b9debe58c474838e457efe96c638ee37362656c18e86024aca620L99-R195) [[2]](diffhunk://#diff-f9b9a25a941b9debe58c474838e457efe96c638ee37362656c18e86024aca620L10-R18)
* Enhanced embedded content handling in chat messages, including proper citation parsing and rendering for modified message text after extracting embedded files and content.

### UI and Component Enhancements

* Updated the `ChatInputArea` component to improve agent selection logic and button layout, providing a more robust and user-friendly interface for agent selection during chat interactions.
* Extended `MessageActions` to optionally display sources/citations alongside workflow and feedback actions, supporting richer context for chat messages. [[1]](diffhunk://#diff-f9b9a25a941b9debe58c474838e457efe96c638ee37362656c18e86024aca620L31-R39) [[2]](diffhunk://#diff-f9b9a25a941b9debe58c474838e457efe96c638ee37362656c18e86024aca620L58-R66) [[3]](diffhunk://#diff-f9b9a25a941b9debe58c474838e457efe96c638ee37362656c18e86024aca620R86)

### Dependency Updates

* Added new dependencies for research features (`@ariakit/react` and related packages) and updated several existing dependencies for improved compatibility, including `@types/react`, `@types/react-dom`, and `csstype`. These updates affect both `package.json` and `package-lock.json` files. [[1]](diffhunk://#diff-2a0b1b97cd2a8a0f6ee43d592510c4b820af1558c2225e11df9b45533a36e36cR68) [[2]](diffhunk://#diff-2a0b1b97cd2a8a0f6ee43d592510c4b820af1558c2225e11df9b45533a36e36cL123-R128) [[3]](diffhunk://#diff-3b3f5e9b2ea87bef48dcb1250a1bd19828033e48b449de12fbae545438e3b1b7R12) [[4]](diffhunk://#diff-3b3f5e9b2ea87bef48dcb1250a1bd19828033e48b449de12fbae545438e3b1b7R147-R184) [[5]](diffhunk://#diff-3b3f5e9b2ea87bef48dcb1250a1bd19828033e48b449de12fbae545438e3b1b7L60-R62) [[6]](diffhunk://#diff-3b3f5e9b2ea87bef48dcb1250a1bd19828033e48b449de12fbae545438e3b1b7L2808-R2863) [[7]](diffhunk://#diff-3b3f5e9b2ea87bef48dcb1250a1bd19828033e48b449de12fbae545438e3b1b7L4426-R4467)

These changes collectively improve the frontend's capabilities for research workflows, citation management, and overall user experience.


configs/agents/deep_research_agent.yaml
```diff
@@ -0,0 +1,234 @@
...
+        1. **Extract the settings** from the message header:
+           - Duration in minutes → Convert to seconds for max_runtime_seconds
+           - Max Iterations → Use as max_iterations parameter
+           - Sources → Use as sources array
+        
+        2. **Call deep_research with extracted settings:**
+           ```
+           deep_research(
+               research_question="<actual research question without the settings header>",
+               max_runtime_seconds=<duration in minutes * 60>,
```
efunneko: Why don't we pass in the minutes to the tool rather than asking the LLM to do math? The software can convert to seconds

configs/agents/deep_research_agent.yaml
```diff
@@ -0,0 +1,234 @@
...
+        The tool will show real-time progress through:
+        - Planning: Generating initial queries
+        - Searching: Finding sources across multiple platforms
+        - Reflecting: Assessing quality and identifying gaps
+        - Synthesizing: Creating the final report
+        
+        Remember: Always check for and use the user's deep_research_settings from metadata!
+      
+      tools:
+        - tool_type: builtin
```
efunneko: Include artifact management tools?

configs/agents/deep_research_agent.yaml
```diff
@@ -0,0 +1,234 @@
...
+              report_generation:
+                model: "openai/gpt-5-nano"
+                temperature: 1
+                timeout: 300
+
+
+      session_service: *default_session_service
+      artifact_service: *default_artifact_service
+      
+      artifact_handling_mode: "embed"
```
efunneko: I think this should be reference like our other agents

configs/agents/deep_research_agent.yaml
```diff
@@ -0,0 +1,234 @@
...
+
+      # Agent Card Definition
+      agent_card:
+        version: "1.0.0"
+        description: |
+          Advanced research agent that conducts comprehensive, iterative research across web and knowledge base sources.
+          
+          Capabilities:
+          • Deep Research: Iterative research with LLM-powered reflection and query refinement
+          • Web Search: Google and Tavily search integration for public information
```
efunneko: I don't think that we should hard code that we have Google and Tavily, since it is unlikely that both will be available.

configs/agents/deep_research_agent.yaml
```diff
@@ -0,0 +1,234 @@
...
+          • Report Generation: Comprehensive reports with proper citations
+          
+          Perfect for:
+          - Complex research topics requiring depth and breadth from public sources
+          - Multi-perspective analysis using web and internal documents
+          - Comprehensive information gathering from web and knowledge bases
+          - Research requiring both public web sources and internal documentation
+          - Topics where citation and source quality matter
+          
+          Note: This is the web-only version. Requires Google or Tavily API keys for web search.
```
efunneko: We should think a bit about this agent card. While it for sure has utility for people to read, this description is going to be put into the context of any agents that are able to call it. We need the right balance of understanding what it can do (probably best suited for skills) but not too much text that is not useful to the calling agent (e.g. orchestrator). We can put more text in the documentation for deep research for people to read

Definitely, saying you need keys is not useful in this context

configs/agents/deep_research_agent.yaml
```diff
@@ -0,0 +1,234 @@
...
+            description: "Searches across web and knowledge bases in parallel for comprehensive coverage."
+          - id: "research_reflection"
+            name: "Research Reflection"
+            description: "LLM-powered quality assessment that identifies gaps and refines search strategy."
+          - id: "citation_management"
+            name: "Citation Management"
+            description: "Automatic citation tracking with relevance scores and source metadata."
+          - id: "report_generation"
+            name: "Report Generation"
+            description: "LLM synthesizes findings into coherent, well-structured research reports."
```
efunneko: I think this skills might need a refinement. My understanding about skills is that these are what you want to advertise as the 'things I can do for you'. When I look at the instruction for the agent, it is only about doing deep research. So, I think that this agent really only has one skill. 

I think in general, each skill listed here should have some information in the agent's instruction to explain how it should tackle that skill

Note that some of this information (e.g. automatic citation tracking) may be relevant, but it should be incorporated into the deep research skill I think

examples/agents/deep_research_agent.yaml
```diff
@@ -0,0 +1,257 @@
...
+        The tool will show real-time progress through:
+        - Planning: Generating initial queries
+        - Searching: Finding sources across multiple platforms
+        - Reflecting: Assessing quality and identifying gaps
+        - Synthesizing: Creating the final report
+        
+        Remember: Use defaults unless user explicitly requests different parameters!
+      
+      tools:
+        - tool_type: builtin
```
efunneko, line 194: Should we add any other tools? 

I think the built-in artifact management ones could possibly be useful...

examples/agents/deep_research_agent.yaml
```diff
@@ -0,0 +1,257 @@
...
+      
+            # API keys for web search
+            tavily_api_key: "${TAVILY_API_KEY}"
+            google_search_api_key: "${GOOGLE_SEARCH_API_KEY}"
+            google_cse_id: "${GOOGLE_CSE_ID}"
+            
+            # Phase-specific model configurations (optional)
+            model_configs:
+              report_generation:
+                model: "openai/gpt-5-mini"
```
efunneko: I know this is an example, but do we want to hardcode the model here?

examples/agents/deep_research_agent.yaml
```diff
@@ -0,0 +1,257 @@
...
+            description: "Searches across web and knowledge bases in parallel for comprehensive coverage."
+          - id: "research_reflection"
+            name: "Research Reflection"
+            description: "LLM-powered quality assessment that identifies gaps and refines search strategy."
+          - id: "citation_management"
+            name: "Citation Management"
+            description: "Automatic citation tracking with relevance scores and source metadata."
+          - id: "report_generation"
+            name: "Report Generation"
+            description: "LLM synthesizes findings into coherent, well-structured research reports."
```
efunneko: I don't think this agent really has all these skills. For example, you would not call this agent to generate a report. You would call it for deep research and get back a deep research report, but that is the output of deep research. I think the agent only has one skill -> deep web research

src/solace_agent_mesh/agent/tools/web_tools.py
```diff
@@ -70,9 +71,10 @@ async def web_request(
     output_artifact_filename: Optional[str] = None,
     tool_context: ToolContext = None,
     tool_config: Optional[Dict[str, Any]] = None,
+    max_retries: int = 2,
```
efunneko, line 78: I know we didn't have it before, but I wonder if this tool should support a max size for fetching? I am not sure what protection the underlying library could offer, but if it got a URL to a huge file, it would bring down the agent

efunneko: Note that this doesn't need to be an input parameter necessarily. It could just be a tool config - though we would need to be careful about how it is set to not allow people in a cloud deployment to put a huge limit.

src/solace_agent_mesh/agent/tools/deep_research_tools.py
```diff
@@ -0,0 +1,1761 @@
...
+            )
+        else:
+            # Fallback to simple text progress
+            progress_data = AgentProgressUpdateData(status_text=message)
+        
+        logical_task_id = a2a_context.get("logical_task_id")
+        context_id = a2a_context.get("contextId")
+        
+        # Create status update event using the standard data signal pattern
+        status_update_event = a2a.create_data_signal_event(
```
efunneko: Hmm... this really makes me think we need to get a better tool framework that can provide these services to tools rather than having them call publishing routines directly. Way too much boilerplate here. Ideally, we could have a way to define status updates for the tool in a generic way and then be able to just call a higher level function with status updates... 

I wouldn't hold up this feature for that, but I think we need to build that before we do more of this

src/solace_agent_mesh/agent/tools/deep_research_tools.py
```diff
@@ -0,0 +1,1761 @@
...
+        # Create artifact for the research report
+        from ..tools.builtin_artifact_tools import _internal_create_artifact
+        from ..adk.tool_wrapper import ADKToolWrapper
+        
+        # Generate filename from research question
+        import re
+        safe_filename = re.sub(r'[^\w\s-]', '', research_question.lower())
+        safe_filename = re.sub(r'[-\s]+', '_', safe_filename)
+        safe_filename = safe_filename[:50]  # Limit length
+        artifact_filename = f"{safe_filename}_report.md"
```
efunneko: Can we move this to a utils module so that others can use it? I wouldn't be surprised if there is already this concept somewhere else (possibly not nicely placed in a utils file, but it should be)

src/solace_agent_mesh/agent/tools/deep_research_tools.py
```diff
@@ -0,0 +1,1761 @@
...
+        
+        # Create the artifact - IMPORTANT: Use original tool_context to preserve state
+        wrapped_creator = ADKToolWrapper(
```
efunneko: We shouldn't need to use the ADKToolWrapper here. It is only needed in the callbacks.py file so that we can create a fake tool call for the agent's history. Within tools, we can just call the artifact service directly - see other tools that create artifacts

src/solace_agent_mesh/agent/tools/deep_research_tools.py
```diff
@@ -0,0 +1,1761 @@
...
+                                with host_component.active_tasks_lock:
+                                    task_context = host_component.active_tasks.get(logical_task_id)
+                                if task_context:
+                                    task_context.register_produced_artifact(artifact_filename, artifact_version)
+                                    log.info("%s Explicitly registered artifact '%s' v%d in task context",
+                                            log_identifier, artifact_filename, artifact_version)
+                                    
+                                    # CRITICAL: Signal the artifact for return to chat
+                                    # This triggers the artifact to be sent as a message part via _handle_artifact_return_signals
+                                    task_context.add_artifact_signal({
```
efunneko: Doing this artifact signalling is not really used anymore. It was originally here to allow tools to have artifacts returned to the gateway without the LLM really being away of it, but now we want the LLM to be aware for its history. Your agent config is handling this with its direction to return the artifact, so all of this should be unnecessary

src/solace_agent_mesh/agent/tools/deep_research_tools.py
```diff
@@ -0,0 +1,1761 @@
...
+from google.genai import types as adk_types
+from google.adk.models import LlmRequest
+from solace_ai_connector.common.log import log
+
+from .tool_definition import BuiltinTool
+from .registry import tool_registry
+from .web_search_tools import _web_search_tavily, _web_search_google
+from .web_tools import web_request
+from ...common import a2a
+from ...gateway.http_sse.routers.dto.rag_dto import create_rag_source, create_rag_search_result
```
efunneko: This feels like a weird import. A tool that needs to reach deep into a gateway feels like the organization of the code is not correct

src/solace_agent_mesh/tools/web_search/brave_search.py
```diff
@@ -0,0 +1,378 @@
...
+logger = logging.getLogger(__name__)
+
+
+class BraveSearchTool(WebSearchTool):
+    """Brave Web Search API implementation.
+    
+    Brave Search provides privacy-focused web search with support for:
+    - Web search results with snippets
+    - News results
+    - Video results
```
efunneko: There is a lot of code in this directory that I really fear is going to be a maintenance issue if we have no testing. I don't think that we will be using most of these services in Solace Chat (likely only Google) and unless we have someone that may use them, I would fear that by the time someone might want to use them, they might not work due to something changing in the service. 

I think most of this belongs as a community plugin (we do have a SAM community plugins repo)

src/solace_agent_mesh/agent/tools/deep_research_tools.py
amir-ghasemi: moving to common

src/solace_agent_mesh/agent/tools/deep_research_tools.py
amir-ghasemi: removed

src/solace_agent_mesh/agent/tools/deep_research_tools.py
amir-ghasemi: done

configs/agents/deep_research_agent.yaml
amir-ghasemi: fixed

configs/agents/deep_research_agent.yaml
amir-ghasemi: done

configs/agents/deep_research_agent.yaml
amir-ghasemi: fixed

configs/agents/deep_research_agent.yaml
amir-ghasemi: done

configs/agents/deep_research_agent.yaml
amir-ghasemi: done

configs/agents/deep_research_agent.yaml
amir-ghasemi: fixed and simplified

examples/agents/deep_research_agent.yaml
amir-ghasemi: done

src/solace_agent_mesh/agent/tools/deep_research_tools.py
amir-ghasemi: done

examples/agents/deep_research_agent.yaml
amir-ghasemi: fixed

examples/agents/deep_research_agent.yaml
amir-ghasemi: commented out

src/solace_agent_mesh/agent/tools/web_tools.py
amir-ghasemi: added max response size limit

src/solace_agent_mesh/agent/tools/deep_research_tools.py
amir-ghasemi: fully agreed

src/solace_agent_mesh/agent/tools/deep_research_tools.py
```diff
@@ -0,0 +1,1762 @@
...
+        # Create status update event using the standard data signal pattern
+        status_update_event = a2a.create_data_signal_event(
+            task_id=logical_task_id,
+            context_id=context_id,
+            signal_data=progress_data,
+            agent_name=host_component.agent_name,
+        )
+        
+        # Publish via the host component's async method
+        loop = host_component.get_async_loop()
```
efunneko: Move most of this into a function in component.py

src/solace_agent_mesh/agent/tools/deep_research_tools.py
```diff
@@ -0,0 +1,1762 @@
...
+from dataclasses import dataclass, field
+
+from google.adk.tools import ToolContext
+from google.genai import types as adk_types
+from google.adk.models import LlmRequest
+from solace_ai_connector.common.log import log
+
+from .tool_definition import BuiltinTool
+from .registry import tool_registry
+from .web_search_tools import _web_search_tavily, _web_search_google
```
efunneko: Ideally we aren't importing functions starting with '_' which should be internal to the file they live in

examples/agents/deep_research_agent.yaml
```diff
@@ -0,0 +1,237 @@
...
+             "response_artifact": {
+               "filename": "history_of_ai_report.md",
+               "version": 1
+             }
+           }
+           ```
+           
+           Your response should be:
+           ```markdown
+           «artifact_content:history_of_ai_report.md:1»
```
efunneko: Change to artifact_return - here and everywhere else we mention it

src/solace_agent_mesh/agent/tools/deep_research_tools.py
```diff
@@ -0,0 +1,1762 @@
...
+        await _send_research_progress(
+            f"Searching web for: {query[:60]}...",
+            tool_context
+        )
+    
+    # Try Google first (preferred for deep research)
+    # The web search functions will handle API key checking via tool_config or os.getenv
+    try:
+        log.info("%s Attempting Google search", log_identifier)
+        result = await _web_search_google(
```
efunneko: I don't think this needs to be addressed right now, but I wonder if we should in the future define a web_search_service that has a standard interface and can be configured against an agent. Then any tools that need this service simply call it without needing to understand the underlying implementation. This like we have for the artifact service

src/solace_agent_mesh/agent/tools/deep_research_tools.py
```diff
@@ -0,0 +1,1762 @@
...
+    """
+    log_identifier = f"[DeepResearch:ModelSelection:{phase}]"
+    
+    # Get agent's default model
+    inv_context = tool_context._invocation_context
+    agent = getattr(inv_context, 'agent', None)
+    default_model = agent.canonical_model if agent else None
+    
+    if not default_model:
+        raise ValueError(f"{log_identifier} No default model available")
```
efunneko, line 241: Should we allow this to fall back to the model configured for the agent itself? We do this for some other tools that use the LLM - see the builtin artifact tools

src/solace_agent_mesh/tools/web_search/brave_search.py
amir-ghasemi: removed the extra engines.

src/solace_agent_mesh/agent/tools/deep_research_tools.py
amir-ghasemi: done

src/solace_agent_mesh/agent/tools/deep_research_tools.py
amir-ghasemi: good idea given this will likely be a heavily used feature.

src/solace_agent_mesh/agent/tools/deep_research_tools.py
amir-ghasemi: done

examples/agents/deep_research_agent.yaml
amir-ghasemi: already done

src/solace_agent_mesh/agent/tools/deep_research_tools.py
amir-ghasemi: done

client/webui/frontend/src/lib/components/activities/VisualizerStepCard.tsx
```diff
@@ -143,19 +144,63 @@ const VisualizerStepCard: React.FC<VisualizerStepCardProps> = ({ step, isHighlig
...
+                if (innerResult?.images && Array.isArray(innerResult.images) && innerResult.images.length > 0) {
+                    hasImages = true;
+                }
+            }
+        } catch {
+            // Not JSON or parsing failed, will display as normal
+        }
+
+        return (
+            <div className="mt-1.5 rounded-md bg-gray-50 p-2 text-xs text-gray-700 dark:bg-gray-700 dark:text-gray-300">
```
lgh-solace: If you have time, please try to replace with theme palette colours. I realize the visualizer doesn't yet use palette colours, but we try not to add more.

client/webui/frontend/package.json
```diff
@@ -120,9 +121,11 @@
         "@testing-library/cypress": "^10.0.3",
         "@types/js-yaml": "^4.0.9",
         "@types/node": "^22.15.29",
-        "@types/react": "19.0.0",
-        "@types/react-dom": "19.0.0",
+        "@types/react": "^19.1.2",
```
lgh-solace: Did you require this update? The react versions are locked down to match enterprise versions, so if updated here we will need to update there as well and please remove the carets. Also, check the peer dependencies.

amir-ghasemi: fixed, thanks.

client/webui/frontend/package.json
```diff
@@ -65,6 +65,7 @@
         ]
     },
     "dependencies": {
+        "@ariakit/react": "^0.4.19",
```
lgh-solace: Please ensure any new dependencies (direct or indirectly) added do not appear on the compromised list: https://github.com/DataDog/indicators-of-compromise/blob/main/shai-hulud-2.0/consolidated_iocs.csv

amir-ghasemi: not on the list

ghaithdallaali: @amir-ghasemi , we already have a component library, this is a major addition that I don't think we should be adding to SAM unless absolutely necessary.
@JKaram , can you please point Amir to the library we use?

client/webui/frontend/src/lib/components/web/Citation.tsx
```diff
@@ -0,0 +1,184 @@
...
+ * Extract clean domain from URL
+ */
+export function getCleanDomain(url: string): string {
+    try {
+        const domain = url.replace(/^https?:\/\//, "").split("/")[0];
+        return domain.startsWith("www.") ? domain.substring(4) : domain;
+    } catch {
+        return url;
+    }
+}
```
ghaithdallaali: This function is defined twice. There should only be one definition.

client/webui/frontend/src/lib/components/web/Citation.tsx
```diff
@@ -0,0 +1,184 @@
...
+import * as Ariakit from "@ariakit/react";
+import { ChevronDown, ExternalLink } from "lucide-react";
+import { cn } from "@/lib/utils";
+
+/**
+ * Get favicon URL from Google's service
+ */
+function getFaviconUrl(domain: string): string {
+    return `https://www.google.com/s2/favicons?domain=${domain}&sz=32`;
+}
```
ghaithdallaali: This function is defined twice. There should only be one definition.

client/webui/frontend/src/lib/components/web/Citation.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * Extract clean domain from URL
+ */
+export function getCleanDomain(url: string): string {
+    try {
+        const domain = url.replace(/^https?:\/\//, "").split("/")[0];
+        return domain.startsWith("www.") ? domain.substring(4) : domain;
```
ghaithdallaali: Shouldn't the regex here be defined in common utilities?

client/webui/frontend/src/lib/utils/citations.ts
```diff
@@ -0,0 +1,194 @@
...
+ * Uses [[cite:type0]] pattern that LLMs can reliably reproduce
+ * Also supports [[cite:research0]] for deep research citations
+ */
+
+import type { RAGSource, RAGSearchResult } from "@/lib/types/fe";
+
+// Citation marker pattern: [[cite:file0]], [[cite:ref0]], [[cite:search0]], [[cite:research0]], etc.
+// Also matches [[cite:0]] (treats as search citation when no type prefix)
+// Also matches single bracket [cite:xxx] in case LLM uses wrong format
+export const CITATION_PATTERN = /\[?\[cite:(?:(file|ref|search|research))?(\d+)\]\]?/g;
```
ghaithdallaali, line 15: Back to my comment on the regex pattern, shouldn't it go here?

client/webui/frontend/src/lib/components/chat/ConnectionRequiredModal.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { Button } from "@/lib/components/ui/button";
+
+interface ConnectionRequiredModalProps {
+    isOpen: boolean;
+    onClose: () => void;
+    sourceName: string;
+    onNavigateToConnections: () => void;
+}
+
+export const ConnectionRequiredModal: React.FC<ConnectionRequiredModalProps> = ({ isOpen, onClose, sourceName, onNavigateToConnections }) => {
```
ghaithdallaali, line 20: Is this a new dialog component? If so, we already have one, so unless there is a justifiable need for creating a new one, let's please use the existing component instead.

amir-ghasemi: done.

client/webui/frontend/src/lib/components/web/Citation.tsx
amir-ghasemi: done. Updated to import from @/lib/utils

client/webui/frontend/src/lib/components/web/Citation.tsx
amir-ghasemi: fixed

client/webui/frontend/src/lib/utils/citations.ts
amir-ghasemi: moved to common utilities

client/webui/frontend/src/lib/components/web/Citation.tsx
amir-ghasemi: moved to common utilities.

client/webui/frontend/package.json
amir-ghasemi: switched to Radix UI Popover instead of Ariakit

client/webui/frontend/package-lock.json
```diff
@@ -1201,6 +1201,8 @@
         },
         "node_modules/@radix-ui/react-arrow": {
             "version": "1.1.7",
+            "resolved": "https://registry.npmjs.org/@radix-ui/react-arrow/-/react-arrow-1.1.7.tgz",
+            "integrity": "sha512-F+M1tLhO+mlQaOWspE8Wstg+z6PwxwRd8oQ8IXceWz92kfAmalTRf0EjrouQeo7QssEPfCn05B4Ihs1K9WQ/7w==",
```
JKaram: Can you check why these changes are in the `lock` file? 

If you are not adding anything into package.json there shouldn't be any changes here.

client/webui/frontend/src/lib/components/activities/VisualizerStepCard.tsx
```diff
@@ -143,19 +144,63 @@ const VisualizerStepCard: React.FC<VisualizerStepCardProps> = ({ step, isHighlig
...
+        let parsedResult = null;
+        let hasImages = false;
+
+        try {
+            // Try to parse the result if it's a string
+            if (typeof data.resultData === "string") {
+                parsedResult = JSON.parse(data.resultData);
+            } else if (typeof data.resultData === "object") {
+                parsedResult = data.resultData;
+            }
```
JKaram: Nitpick but I see a lot of `x === "string"` or `x === "object"`

This would be a good time to create a general util in the utils directory, for comparisons like this.

client/webui/frontend/src/lib/components/activities/VisualizerStepCard.tsx
```diff
@@ -143,19 +144,63 @@ const VisualizerStepCard: React.FC<VisualizerStepCardProps> = ({ step, isHighlig
...
+                        </details>
+                    </>
+                ) : (
+                    <>
+                        <p className="mt-1">
+                            <strong>Result:</strong>
+                        </p>
+                        <div className="max-h-40 overflow-y-auto rounded bg-gray-100 p-1.5 dark:bg-gray-700">
+                            {typeof data.resultData === "object" ? <JSONViewer data={data.resultData} /> : <pre className="font-mono text-xs break-all whitespace-pre-wrap">{String(data.resultData)}</pre>}
+                        </div>
```
JKaram: This line is used twice with exact same classnames and logic
```
 {typeof data.resultData === "object" ? <JSONViewer data={data.resultData} /> : <pre className="font-mono text-xs break-all whitespace-pre-wrap">{String(data.resultData)}</pre>}
 ```

Anytime there is duplication of logic like this it is a good idea to abstract it into one spot, so it makes updating it easier less error prone.

File: client/webui/frontend/src/lib/components/chat/ChatSidePanel.tsx [hunk 33]
```diff
@@ -18,14 +19,33 @@ interface ChatSidePanelProps {
...
+                    const sourceType = source.sourceType || "web";
+                    // For images, check if they have a source link
+                    if (sourceType === "image") {
+                        return source.sourceUrl || source.metadata?.link;
+                    }
+                    // For other sources, check if they have a URL or sourceUrl
+                    return source.url || source.sourceUrl;
+                })
+        );
+    }, [ragData]);
```
JKaram: This is hard to read IMO. I gave it to claude to help understand it. 

Since you're repeating URL extraction logic across the codebase, create a unified helper:
 
 1. Inconsistent field access - accessing source.url, source.sourceUrl, source.metadata?.link without a clear pattern
  2. Duplicate logic - same URL checking pattern we saw in Citation.tsx
  3. Hardcoded string - "web" magic string
  4. Unclear purpose - what makes a source "valid"?
  5. Type safety - no type guards
  
  
  ```Since you're repeating URL extraction logic across the codebase, create a unified helper:

  // src/lib/utils/sourceUrlHelpers.ts

  export type SourceType = "web" | "image" | "file" | "web_search" | "deep_research";

  export interface SourceUrlInfo {
    url: string | null;
    hasUrl: boolean;
    sourceType: SourceType;
  }

  /**
   * Extracts URL from a source object regardless of where it's stored
   * Handles: source.url, source.sourceUrl, source.metadata.link
   */
  export function getSourceUrl(source: RAGSource | CitationType['source']): SourceUrlInfo {
    if (!source) {
      return { url: null, hasUrl: false, sourceType: "web" };
    }

    const sourceType = (source.sourceType || source.metadata?.type || "web") as SourceType;

    // Priority order: sourceUrl > url > metadata.link
    const url = source.sourceUrl || source.url || source.metadata?.link || null;

    return {
      url,
      hasUrl: !!url,
      sourceType
    };
  }

  /**
   * Check if RAG data contains any sources with valid URLs
   */
  export function hasSourcesWithUrls(ragData: RAGSearchResult[] | undefined): boolean {
    if (!ragData?.length) return false;

    return ragData.some(search =>
      search.sources?.some(source => getSourceUrl(source).hasUrl) ?? false
    );
  }

  /**
   * Get all sources with valid URLs from RAG data
   */
  export function getSourcesWithUrls(ragData: RAGSearchResult[]): RAGSource[] {
    return ragData.flatMap(search =>
      (search.sources ?? []).filter(source => getSourceUrl(source).hasUrl)
    );
  }

  Usage:

  // In your component
  const hasSourcesInSession = useMemo(() =>
    hasSourcesWithUrls(ragData),
    [ragData]
  );

  // You can also use the other utilities:
  const sourcesWithUrls = useMemo(() =>
    getSourcesWithUrls(ragData ?? []),
    [ragData]
  );
  ```

client/webui/frontend/src/lib/components/chat/DeepResearchSettingsPanel.tsx
```diff
@@ -0,0 +1,88 @@
...
+    onSettingsChange: (settings: Partial<DeepResearchSettings>) => void;
+    isOpen: boolean;
+    onToggle: () => void;
+}
+
+const DURATION_OPTIONS = [
+    { label: "Quick", value: 120 },
+    { label: "Standard", value: 300 },
+    { label: "Deep", value: 600 },
+];
```
JKaram: These are `magic numbers`. They should be labeled as constants and be shared anywhere they are used.

I see `DEFAULT_DEEP_RESEARCH_SETTINGS` defined. This would be a good spot to define these durations.

client/webui/frontend/src/lib/components/chat/DeepResearchSettingsPanel.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+    // Close panel when clicking outside
+    useEffect(() => {
+        if (!isOpen) return;
+
+        const handleClickOutside = (event: MouseEvent) => {
+            if (panelRef.current && !panelRef.current.contains(event.target as Node)) {
+                onToggle();
+            }
+        };
```
JKaram: There is already a `useClickOutside` hook in this repo to handle this logic

client/webui/frontend/src/lib/components/chat/ToolsSelector.tsx
```diff
@@ -0,0 +1,212 @@
...
+    webSearchConfigured = true,
+    deepResearchConfigured = true,
+}) => {
+    const [isOpen, setIsOpen] = useState(false);
+
+    // Check if deep research is available
+    const hasDeepResearch = agents.some(a => a.skills?.some(s => s.id === "deep_research"));
+
+    // Check if web search is available
+    const hasWebSearch = agents.some(a => a.skills?.some(s => s.id === "web_search"));
```
JKaram: These are logics that I see repeated a lot. I think it could even be in one utility that returns if these are true or false. 

You don't have to go this overboard necessarily (adding SKILL_IDS). But It could be a better abstraction, and used for future use.

```
 // src/lib/utils/agentHelpers.ts

  export const SKILL_IDS = {
    DEEP_RESEARCH: "deep_research",
    WEB_SEARCH: "web_search",
  } as const;

  /**
   * Check if any agent has a specific skill
   */
  export function hasAgentSkill(agents: Agent[], skillId: string): boolean {
    return agents.some(agent =>
      agent.skills?.some(skill => skill.id === skillId) ?? false
    );
  }

  /**
   * Check for multiple skills at once
   */
  export function getAgentSkills(
    agents: Agent[], 
    skillIds: string[]
  ): Record<string, boolean> {
    return Object.fromEntries(
      skillIds.map(id => [id, hasAgentSkill(agents, id)])
    );
  }

  /**
   * Get all unique skill IDs from agents
   */
  export function getAllAgentSkillIds(agents: Agent[]): string[] {
    return [...new Set(agents.flatMap(a => a.skills?.map(s => s.id) ?? []))];
  }

  /**
   * Get agents that have a specific skill
   */
  export function getAgentsWithSkill(agents: Agent[], skillId: string): Agent[] {
    return agents.filter(agent =>
      agent.skills?.some(skill => skill.id === skillId) ?? false
    );
  }

  Usage:

  import { hasAgentSkill, SKILL_IDS } from "@/lib/utils/agentHelpers";

  const hasDeepResearch = hasAgentSkill(agents, SKILL_IDS.DEEP_RESEARCH);
  const hasWebSearch = hasAgentSkill(agents, SKILL_IDS.WEB_SEARCH);

  // Or get both at once:
  const { deep_research, web_search } = getAgentSkills(agents, [
    SKILL_IDS.DEEP_RESEARCH,
    SKILL_IDS.WEB_SEARCH
  ]);
  ```

client/webui/frontend/src/lib/components/activities/VisualizerStepCard.tsx
amir-ghasemi: Utilities added in guard.ts

client/webui/frontend/src/lib/components/activities/VisualizerStepCard.tsx
amir-ghasemi: Extracted the duplicated result data rendering logic into a reusable [renderResultData()](vscode-webview://1gq9o55mo7k98c7gjauus71glhfrotkiostih8p6402ftbgi55in/client/webui/frontend/src/lib/components/activities/VisualizerStepCard.tsx:154) helper function that handles both object and primitive data types

client/webui/frontend/src/lib/components/chat/ChatSidePanel.tsx
amir-ghasemi: Created `sourceUrlHelpers.ts` for this.

New unified utility file for URL extraction with:

[getSourceUrl()](vscode-webview://1gq9o55mo7k98c7gjauus71glhfrotkiostih8p6402ftbgi55in/client/webui/frontend/src/lib/utils/sourceUrlHelpers.ts:49) - Extracts URL from source objects (handles sourceUrl, url, metadata.link)
[hasSourcesWithUrls()](vscode-webview://1gq9o55mo7k98c7gjauus71glhfrotkiostih8p6402ftbgi55in/client/webui/frontend/src/lib/utils/sourceUrlHelpers.ts:76) - Checks if RAG data has sources with valid URLs
[getSourcesWithUrls()](vscode-webview://1gq9o55mo7k98c7gjauus71glhfrotkiostih8p6402ftbgi55in/client/webui/frontend/src/lib/utils/sourceUrlHelpers.ts:95) - Gets all sources with valid URLs
[isWebBasedSource()](vscode-webview://1gq9o55mo7k98c7gjauus71glhfrotkiostih8p6402ftbgi55in/client/webui/frontend/src/lib/utils/sourceUrlHelpers.ts:107) - Checks if source type is web-based
[hasClickableExternalUrl()](vscode-webview://1gq9o55mo7k98c7gjauus71glhfrotkiostih8p6402ftbgi55in/client/webui/frontend/src/lib/utils/sourceUrlHelpers.ts:117) - Checks if source should open external URL

client/webui/frontend/src/lib/components/chat/DeepResearchSettingsPanel.tsx
amir-ghasemi: this component is now removed entirely.

client/webui/frontend/src/lib/components/chat/DeepResearchSettingsPanel.tsx
amir-ghasemi: component removed

client/webui/frontend/src/lib/components/chat/ToolsSelector.tsx
amir-ghasemi: we are no longer using ToolSelector.

client/webui/frontend/package-lock.json
amir-ghasemi: regenerated the package lock.

JKaram (approved):

PR #579, opened 2025-11-21
fix(DATAGO-116432): add STT to AI prompt builder

amir-ghasemi:
Also fixes a bug for STT language settings of external APIs not taking effect.

### DIFF
--- a/client/webui/frontend/src/lib/components/pages/PromptsPage.tsx
+++ b/client/webui/frontend/src/lib/components/pages/PromptsPage.tsx
@@ -214,7 +214,7 @@ export const PromptsPage: React.FC = () => {
     const handleTogglePin = async (id: string, currentStatus: boolean) => {
         try {
             // Optimistic update
-            setPromptGroups(prev => prev.map(p => (p.id === id ? { ...p, is_pinned: !currentStatus } : p)));
+            setPromptGroups(prev => prev.map(p => (p.id === id ? { ...p, isPinned: !currentStatus } : p)));
 
             const response = await authenticatedFetch(`/api/v1/prompts/groups/${id}/pin`, {
                 method: "PATCH",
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/components/prompts/PromptBuilderChat.tsx
```diff
@@ -35,6 +38,13 @@ export const PromptBuilderChat: React.FC<PromptBuilderChatProps> = ({ onConfigUp
     const inputRef = useRef<HTMLTextAreaElement>(null);
     const initRef = useRef(false);
 
+    // Speech-to-text support
+    const { settings } = useAudioSettings();
+    const { configFeatureEnablement } = useConfigContext();
+    const sttEnabled = configFeatureEnablement?.speechToText ?? true;
```
lgh-solace, line 44: Should this default be false? Not sure if the intention is to be on or off by default.

client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
```diff
@@ -98,7 +98,7 @@ export const ChatInputArea: React.FC<{ agents: AgentCardInfo[]; scrollToBottom?:
                 setPendingPromptGroup({
                     id: groupId,
                     name: groupName,
-                    production_prompt: { prompt_text: promptText },
+                    production_prompt: { promptText: promptText },
```
lgh-solace: Could we change `production_prompt` to camel case as well?

client/webui/frontend/src/lib/components/prompts/PromptBuilderChat.tsx
amir-ghasemi: Not sure. For prompts and projects we enable them by default. This is following the same pattern but can be false as well.

PR #582, opened 2025-11-24
feat (DATAGO-118265): colour-coded console logging

amir-ghasemi:


carolmorneau:
Adding colours is a neat idea. 

Application logging as a concept is owned by the main application at the [solace-ai-connector level](https://github.com/SolaceLabs/solace-ai-connector/blob/main/src/solace_ai_connector/common/logging_config.py). If we want to support colours, I believe supports for it should be added there. 

As it is, the main application logging configuration gets overridden. 
Here's what I see: 
image

When I enable JSON Logs, the JSON formatting is lost as soon as the colouring gets configured. 
image

### A few considerations: 
- Colouring likely only makes sense for non-JSON logs for instance the [simpleFormatter](https://github.com/SolaceLabs/solace-agent-mesh/blob/main/templates/logging_config_template.yaml#L6) we generate by default
- SAM Cloud deployments tail stdout logs; I am not sure if that has an impact but colouring should play well with that. SAM Cloud enables JSON logs so if colouring is OFF with JSON logs that should be fine. 
- [There seem to be libraries](https://github.com/borntyping/python-colorlog) that do that. Maybe we could use one of them or get inspired. Using a proven 3rd-party library also simplifies our testing responsibilities. 
- If there's a performance impact, coloured logs should be off in production

### Implementation Idea

Your custom logging formatter could live in `solace-ai-connector.logging.py` (similar to [this unmerged and closed PR which was attempting to add a custom formatter](https://github.com/SolaceLabs/solace-ai-connector/pull/98/files#diff-437836a092deb89a558fa559f3a448ec6716273f8fe46552e493928efd615d74)) and users could choose to use it in their logging configuration files: 
```
formatters:
  simpleFormatter:
    format: "%(asctime)s | %(levelname)-5s | %(threadName)s | %(name)s | %(message)s"

  simpleColouredFormatter:
    class: solace_ai_connector.logging.colouredFormatter
    format: "%(asctime)s | %(levelname)-5s | %(threadName)s | %(name)s | %(message)s"

  jsonFormatter:
    "()": pythonjsonlogger.json.JsonFormatter
    format: "%(timestamp)s %(levelname)s %(threadName)s %(name)s %(message)s"
    timestamp: "timestamp"

handlers:
  streamHandler:
    class: logging.StreamHandler
    formatter: simpleColouredFormatter
    stream: ext://sys.stdout
```

The main benefit of this approach is that it wouldn't conflict with the application logging configuration since colouring is part of it rather than added later on. This solution also naturally only targets non-JSON formatters. 

It is likely possible to add colouring programmatically (like this PR is doing) but we need to be more careful. Users have a lot of flexibility with their logging config so whatever we do needs to play nice with users' configs.

examples/logging/colored-logging-config.yaml
```diff
@@ -0,0 +1,67 @@
+# Example logging configuration with colored output
```
carolmorneau, line 1: rather than create a new config, we could add the coloredFormatter to the existing https://github.com/SolaceLabs/solace-agent-mesh/blob/main/templates/logging_config_template.yaml
We could consider defaulting to it as well

carolmorneau: this would involve a small doc update: https://solacelabs.github.io/solace-agent-mesh/docs/documentation/deploying/logging#default-logging-configuration

PR #584, opened 2025-11-24
chore(DATAGO-115917): use new confirmation dialog

g-raman:


client/webui/frontend/src/lib/components/prompts/PromptDeleteDialog.tsx
g-raman, line 1: checked in with @amir-ghasemi, and it should be fine to remove the manual cleanup of `radix-dialog` overlays

client/webui/frontend/src/lib/components/chat/FeedbackModal.tsx
```diff
@@ -47,28 +47,19 @@ export const FeedbackModal = React.memo<FeedbackModalProps>(({ isOpen, onClose,
...
-        <Dialog open={isOpen} onOpenChange={handleClose}>
-            <DialogContent className="sm:max-w-[750px]" showCloseButton={false}>
-                <DialogHeader>
-                    <DialogTitle>Provide Feedback</DialogTitle>
-                    <DialogDescription className="flex flex-col gap-2">
-                        <span className="font-medium">{feedbackPrompt}</span>
-                        <span>Providing more details will help improve AI responses over time.</span>
-                    </DialogDescription>
-                </DialogHeader>
+        <ConfirmationDialog
```
lgh-solace: Could you revert the changes to this file? As discussed offline, let's just use the ConfirmationDialog when no user input is required for now. Thanks!

client/webui/frontend/src/lib/components/prompts/PromptDeleteDialog.tsx
lgh-solace: Yes, it's duplicated in AppLayout, I believe - it will be great to get rid of entirely one day...

client/webui/frontend/src/lib/components/prompts/PromptDeleteDialog.tsx
```diff
@@ -11,84 +10,15 @@ interface PromptDeleteDialogProps {
...
-        return null;
-    }
-
     return (
-        <Dialog open={isOpen} onOpenChange={handleOpenChange} modal={true}>
-            <DialogContent>
-                <DialogHeader>
-                    <DialogTitle>Delete Prompt?</DialogTitle>
-                    <DialogDescription>
-                        This action cannot be undone. This will permanently delete the prompt and all its versions: <strong>{promptName}</strong>
```
lgh-solace, line 80: Could we pass this in as content and keep the `strong`?

client/webui/frontend/src/lib/components/prompts/PromptDeleteDialog.tsx
```diff
@@ -11,84 +10,15 @@ interface PromptDeleteDialogProps {
...
-                    </Button>
-                </DialogFooter>
-            </DialogContent>
-        </Dialog>
+        <ConfirmationDialog
+            title="Delete Prompt?"
+            description={`This action cannot be undone. This will permanently delete the prompt and all its versions: ${promptName}`}
+            actionLabels={{ confirm: "Delete" }}
+            open={isOpen}
+            onOpenChange={() => !isOpen && onClose()}
```
lgh-solace, line 23: `onOpenChange` has a boolean param, should we use that instead of `isOpen`?

client/webui/frontend/src/lib/components/prompts/PromptRestoreDialog.tsx
```diff
@@ -1,38 +0,0 @@
...
-import { Button } from "@/lib/components/ui/button";
-
-interface PromptRestoreDialogProps {
-    isOpen: boolean;
-    onClose: () => void;
-    onConfirm: () => void;
-    versionNumber: number;
-}
-
-export const PromptRestoreDialog = React.memo<PromptRestoreDialogProps>(({ isOpen, onClose, onConfirm, versionNumber }) => {
```
lgh-solace, line 13: Thanks!

lgh-solace (commented):
Thanks! Just a couple of nitpicks!

client/webui/frontend/src/lib/components/prompts/PromptDeleteDialog.tsx
g-raman: the functionality would be the same, but i think this makes it clear that it's externally controlled

lgh-solace (approved):

PR #585, opened 2025-11-24
fix(DATAGO-118327): Remove requirement for users to type a message when sending a file created by pasting large text

amir-ghasemi:


client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
```diff
@@ -313,7 +313,27 @@ export const ChatInputArea: React.FC<{ agents: AgentCardInfo[]; scrollToBottom?:
...
+            const artifactFiles: File[] = pastedArtifactItems.map(item => {
+                // Find the artifact in the artifacts list to get its mimeType
+                const artifact = artifacts.find(a => a.uri === item.artifactId);
+
+                // Create a special File object that contains the artifact URI
+                const artifactData = JSON.stringify({
+                    isArtifactReference: true,
+                    uri: item.artifactId,
+                    filename: item.filename,
+                    mimeType: artifact?.mime_type || "application/octet-stream",
```
lgh-solace: Can we add mimeType to the `PastedArtifactItem` model and add it on creation to avoid this looping lookup to get it from the artifacts?

client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
```diff
@@ -313,7 +313,27 @@ export const ChatInputArea: React.FC<{ agents: AgentCardInfo[]; scrollToBottom?:
                 fullMessage = `${fullMessage}\n\nContext: "${contextText}"`;
             }
 
-            await handleSubmit(event, selectedFiles, fullMessage);
+            const artifactFiles: File[] = pastedArtifactItems.map(item => {
+                // Find the artifact in the artifacts list to get its mimeType
+                const artifact = artifacts.find(a => a.uri === item.artifactId);
```
lgh-solace: If the artifact isn't found, we should skip the next step? I guess it's thrown away later on, so just an option to throw away earlier and reduce some processing.

lgh-solace (approved):
Nothing blocking but a couple of comments for consideration, cheers!

client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
amir-ghasemi: Done.

client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
amir-ghasemi: we now skip invalid items early.

PR #591, opened 2025-11-26
fix(DATAGO-118537): stopping clicks from propagating through the dialog

lgh-solace:
This pull request refactors the `DocumentListItem` component to simplify its click handling and button rendering logic. The main improvements include removing custom click logic in favor of direct handler assignment, streamlining the rendering of action buttons, and updating the delete confirmation dialog trigger.

Component refactoring and simplification:

* Removed the custom `handleClick` function that prevented event propagation from buttons, replacing it with direct assignment of the `onClick` handler to the main container. This simplifies the click logic and makes the component easier to understand.

Button rendering improvements:

* Simplified the rendering of action buttons (`Edit Description`, `Download`, and `Delete`) by removing unnecessary props like `size` and custom class names, and by rendering icon components without additional sizing props.
* Changed the delete button to be always visible (when `onDelete` is provided) and to open the confirmation dialog when clicked, instead of using the dialog's `trigger` prop. [[1]](diffhunk://#diff-058d00439b382973574986aebe596b5ff965c418599d6579a4eac8f6cad442d0L56-R84) [[2]](diffhunk://#diff-058d00439b382973574986aebe596b5ff965c418599d6579a4eac8f6cad442d0L92-R96)

Delete confirmation dialog update:

* Updated the confirmation dialog title to "Delete Project File" and improved the dialog's placement in the component tree for clarity.

--- 
BEFORE

https://github.com/user-attachments/assets/c1a50797-9f21-459a-b832-168929aaf3fc

AFTER

Clicking inside the dialog no longer opens the file details.

https://github.com/user-attachments/assets/f6e1cfe3-e062-4029-b90a-1aa3a0675ed4

### DIFF
--- a/client/webui/frontend/src/lib/components/projects/DocumentListItem.tsx
+++ b/client/webui/frontend/src/lib/components/projects/DocumentListItem.tsx
@@ -18,84 +18,81 @@ interface DocumentListItemProps {
 export const DocumentListItem: React.FC<DocumentListItemProps> = ({ artifact, onDownload, onDelete, onClick, onEditDescription }) => {
     const [showDeleteDialog, setShowDeleteDialog] = useState(false);
 
-    const handleClick = (e: React.MouseEvent) => {
-        // Don't trigger onClick if clicking on buttons
-        if ((e.target as HTMLElement).closest("button")) {
-            return;
-        }
-        onClick?.();
-    };
-
     return (
-        <div className={`hover:bg-accent/50 group flex items-center justify-between rounded-md p-2 ${onClick ? "cursor-pointer" : ""}`} onClick={handleClick}>
-            <div className="flex min-w-0 flex-1 items-center gap-2">
-                {getFileIcon(artifact, "h-4 w-4 flex-shrink-0 text-muted-foreground")}
-                <div className="min-w-0 flex-1">
-                    <p className="text-foreground truncate text-sm font-medium" title={artifact.filename}>
-                        {artifact.filename}
-                    </p>
-                    <div className="text-muted-foreground flex items-center gap-2 text-xs">
-                        {artifact.last_modified && (
-                            <span className="truncate" title={formatRelativeTime(artifact.last_modified)}>
-                                {formatRelativeTime(artifact.last_modified)}
-                            </span>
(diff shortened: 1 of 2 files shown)


client/webui/frontend/src/lib/components/projects/DocumentListItem.tsx
```diff
@@ -18,86 +18,81 @@ interface DocumentListItemProps {
...
+                                setShowDeleteDialog(true);
+                            }}
+                        >
+                            <Trash />
+                        </Button>
+                    )}
+                </div>
             </div>
-        </div>
+            {onDelete && (
```
lgh-solace, line 82: The issue was occurring because the dialog was inside the clickable row, so the whole dialog became clickable. The fix was to remove the dialog trigger usage and move it outside the clickable row, leaving only the button in the row.

client/webui/frontend/src/lib/components/projects/DocumentListItem.tsx
```diff
@@ -18,86 +18,81 @@ interface DocumentListItemProps {
...
                     </Button>
-                )}
-                <Button
-                    variant="ghost"
-                    size="sm"
-                    onClick={e => {
-                        e.stopPropagation();
-                        onDownload();
-                    }}
-                    className="h-8 w-8 p-0"
```
lgh-solace, line 74: Removed some of the extra styles to leave the icon buttons with standard styles.

client/webui/frontend/src/lib/components/projects/DocumentListItem.tsx
```diff
@@ -18,86 +18,81 @@ interface DocumentListItemProps {
 export const DocumentListItem: React.FC<DocumentListItemProps> = ({ artifact, onDownload, onDelete, onClick, onEditDescription }) => {
     const [showDeleteDialog, setShowDeleteDialog] = useState(false);
 
-    const handleClick = (e: React.MouseEvent) => {
```
lgh-solace, line 21: No longer necessary - buttons don't propagate the events.

ziyanwan (approved):

PR #595, opened 2025-11-27
feat(DATAGO-117735): Add About section and version metadata API

bwiebe-solace:
This PR adds an "About" section to the Settings dialog, which currently displays version information for installed SAM products.

The data for this is provided by a new /version API, which returns the installed products, their version, some additional metadata, and optionally versions for their dependencies. The API checks the installed Python packages and pulls the versions dynamically to ensure that the data is up-to-date. For the UI, a ui-version.json file is generated at build-time and placed in the static directory which gets distributed. This ensures that the version shown reflects what was built into the application, regardless of it was in CI or local.

image

### DIFF
--- a/tests/integration/apis/test_router_registration.py
+++ b/tests/integration/apis/test_router_registration.py
@@ -1,0 +1,43 @@
+"""
+Tests for router registration in main.py.
+
+Verifies that the version router is correctly registered in the FastAPI app
+at line 609 of main.py.
+"""
+
+import pytest
+from fastapi.testclient import TestClient
+
+
+class TestVersionRouterRegistration:
+    """Tests that the version router is properly registered in the app."""
+
+    def test_version_router_is_registered_with_correct_prefix(
+        self, api_client: TestClient
+    ):
+        """Test that version router is accessible at /api/v1/version."""
+        response = api_client.get("/api/v1/version")
+        assert response.status_code == 200
+
+    def test_version_router_not_accessible_without_prefix(
+        self, api_client: TestClient
+    ):
+        """Test that version endpoint requires the /api/v1 prefix."""
+        response = api_client.get("/version")
+        assert response.status_code == 404
+
+    def test_version_router_has_correct_tag_in_openapi(
+        self, api_client: TestClient
+    ):
+        """Test that version router is tagged with 'Version' in OpenAPI schema."""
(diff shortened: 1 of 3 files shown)


client/webui/frontend/src/lib/components/settings/AboutProduct.tsx
```diff
@@ -0,0 +1,80 @@
...
+                        ))}
+                </TableBody>
+            </Table>
+        );
+    };
+
+    useEffect(() => {
+        const fetchVersions = async (): Promise<void> => {
+            try {
+                const response = await authenticatedFetch("/api/v1/version");
```
ziyanwan: You might want to merge the latest main branch and refactor the usage of authenticatedFetch to fetchJsonWithError as did in [this PR ](https://github.com/SolaceLabs/solace-agent-mesh/pull/608/files )by @lgh-solace

src/solace_agent_mesh/gateway/http_sse/routers/version.py
```diff
@@ -0,0 +1,343 @@
...
+from .dto.responses.version_responses import ProductInfo, VersionResponse
+
+log = logging.getLogger(__name__)
+
+router = APIRouter()
+
+# Constants
+NOT_INSTALLED = "not-installed"
+UNKNOWN_VERSION = "unknown"
+UI_VERSION_RELATIVE_PATH = Path("client") / "webui" / "frontend" / "static" / "ui-version.json"
```
mo-radwan1, line 21: I am curious, do we actually need a backend for this ? It seems like the ui-version is a JSON file that is serviced to the UI. Does the UI not have direct access to that file ? Is there any purpose for having an API instead of having the UI directly parse the json file ?

bwiebe-solace: Quoting Ziyang from the discussion in Slack, so that the convo is reflected here too for others:

> version.json is only available in enterprise, not community. The about page is in community.
Besides, the data in vesion.json is from workflow, so local env won't work for community.

One more thing to note is that since we can't read the UI version in the same way as the Python packages, and we likely don't want to serve our full package.json, the version from package.json had to be included in the production bundle so that we could access the version at all. We could request that file directly, but it made more sense to me to keep everything unified, even if that does mean that little extra parsing on the backend.

PR #604, opened 2025-12-01
feat(DATAGO): tanstack query

g-raman:
# Description
Replace manual management of fetch calls and use react query

# Co-pilot summary
This pull request introduces React Query for managing project-related API calls in the frontend, refactoring the projects feature to use a more robust and scalable data fetching and caching approach. The changes include new dependencies, context/provider updates, and a complete migration of project API logic to React Query hooks.

**React Query integration and API management:**

* Added `@tanstack/react-query` and `@lukemorales/query-key-factory` dependencies to `package.json` and `package-lock.json` to support advanced query caching and key management. [[1]](diffhunk://#diff-2a0b1b97cd2a8a0f6ee43d592510c4b820af1558c2225e11df9b45533a36e36cR69) [[2]](diffhunk://#diff-2a0b1b97cd2a8a0f6ee43d592510c4b820af1558c2225e11df9b45533a36e36cR83) [[3]](diffhunk://#diff-3b3f5e9b2ea87bef48dcb1250a1bd19828033e48b449de12fbae545438e3b1b7R13) [[4]](diffhunk://#diff-3b3f5e9b2ea87bef48dcb1250a1bd19828033e48b449de12fbae545438e3b1b7R27) [[5]](diffhunk://#diff-3b3f5e9b2ea87bef48dcb1250a1bd19828033e48b449de12fbae545438e3b1b7R856-R868) [[6]](diffhunk://#diff-3b3f5e9b2ea87bef48dcb1250a1bd19828033e48b449de12fbae545438e3b1b7R2570-R2595)
* Wrapped the root `App` component with `QueryClientProvider` to provide React Query context throughout the application. [[1]](diffhunk://#diff-0ef8a06447cf92a9d371d4085aa46ebfffa04c9b6cfcc7cc268649e627fd60f7R5-R17) [[2]](diffhunk://#diff-0ef8a06447cf92a9d371d4085aa46ebfffa04c9b6cfcc7cc268649e627fd60f7R35)

**Projects API refactor:**

* Created new React Query hooks in `src/features/projects/api/hooks.ts` to handle all project-related API interactions (fetching, creating, updating, deleting projects and artifacts, etc.), replacing previous context-based approaches.
* Added a query key factory in `src/features/projects/api/keys.ts` for consistent and scalable query key management for all project operations.
* Implemented project API service functions in `src/features/projects/api/services.ts` for all CRUD operations, artifact management, and session retrieval.

**Component updates for new API usage:**

* Updated `ChatSessions` and `SessionList` components to use the new React Query hooks for project data, removing reliance on `ProjectProvider`. [[1]](diffhunk://#diff-0594ef9d2ab79f9d03f39195475221a18d005192e1c66f482cb795bba1d4f0fdR1-R12) [[2]](diffhunk://#diff-b907bb650b3bbc5c541fd2e664a1f5770f1e435067c499a5d62a8bb2258db2bfL21-R21)
* Refactored `AddProjectFilesDialog` to use a callback signature compatible with the new API, passing files and metadata directly instead of a `FormData` object. [[1]](diffhunk://#diff-8d8b7ee77e46c811c2f434ade8c8dfba01f60cfd16ae39ceb9bc916f45f5ce91L12-R16) [[2]](diffhunk://#diff-8d8b7ee77e46c811c2f434ade8c8dfba01f60cfd16ae39ceb9bc916f45f5ce91L42-R46)

These changes modernize the frontend data layer for projects, improving reliability, scalability, and developer experience.


client/webui/frontend/package.json
```diff
@@ -79,6 +80,7 @@
         "@radix-ui/react-tabs": "^1.1.12",
         "@radix-ui/react-tooltip": "^1.2.7",
         "@tailwindcss/vite": "^4.1.10",
+        "@tanstack/react-query": "^5.90.11",
```
lgh-solace: Should this be a peer dependency and nailed down to a specific version for enterprise usage too?

Further, please ensure no add dependency pulls in anything from the recent list: https://github.com/DataDog/indicators-of-compromise/blob/main/shai-hulud-2.0/consolidated_iocs.csv

client/webui/frontend/src/lib/api/projects/hooks.ts
```diff
@@ -0,0 +1,150 @@
+import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
```
lgh-solace, line 1: Only source under `lib` is exported into the npm package for use in enterprise, so I believe these `features` be under `lib` as well otherwise things will be broken.

client/webui/frontend/src/features/projects/api/services.ts
```diff
@@ -0,0 +1,130 @@
...
+    files.forEach(file => {
+        formData.append("files", file);
+    });
+
+    if (fileMetadata && Object.keys(fileMetadata).length > 0) {
+        formData.append("fileMetadata", JSON.stringify(fileMetadata));
+    }
+
+    const url = `/api/v1/projects/${projectId}/artifacts`;
+    const response = await fetchWithError(url, {
```
lgh-solace: This should be `fetchJsonWithError` so any potential parsing error is also handled nicely.

There are a few places in this file.

client/webui/frontend/src/features/projects/api/services.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export const getProjectArtifacts = async (projectId: string) => {
+    const url = `/api/v1/projects/${projectId}/artifacts`;
+    const response = (await fetchJsonWithError(url, { credentials: "include" })) as ArtifactInfo[];
+
+    return response;
+};
+
+export const getProjectSessions = async (projectId: string) => {
+    const url = `/api/v1/sessions?project_id=${projectId}&pageNumber=1&pageSize=100`;
```
lgh-solace: Was this hard-coded pageSize there before? It likely should loop.. fetching all of them?

client/webui/frontend/src/features/projects/api/services.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    if (fileMetadata && Object.keys(fileMetadata).length > 0) {
+        formData.append("fileMetadata", JSON.stringify(fileMetadata));
+    }
+
+    const url = `/api/v1/projects/${projectId}/artifacts`;
+    const response = await fetchWithError(url, {
+        method: "POST",
+        body: formData,
+        credentials: "include",
```
lgh-solace: I believe we don't actually need to add `credentials: "include"`  everywhere - using any of the utility methods will do the right thing.

client/webui/frontend/src/lib/components/projects/ProjectDescription.tsx
```diff
@@ -1,87 +0,0 @@
...
-
-interface ProjectDescriptionProps {
-    project: Project;
-    isEditing: boolean;
-    onToggleEdit: () => void;
-    onSave: (description: string) => Promise<void>;
-    isSaving: boolean;
-}
-
-export const ProjectDescription: React.FC<ProjectDescriptionProps> = ({
```
lgh-solace, line 15: Were these components unused? If so, I wonder if there was a planned future usage? 🤔

client/webui/frontend/src/features/projects/api/services.ts
JKaram: Two things: 

I am interested in writing more utility functions and types for apis like this. It would take in `projectId: string, pagination: Pagination, filters: Filters` (or something to this effect). I also want to add an `enabled` param for each hook. I'm still thinking about the right implementation, not for this PR.


For Lindas comment: 

I believe the fetch call itself should be responsible for fetching ALL pages. Like a fetchAllByPage util.

And IMO that should be it's own hook like `useGetAllProjectSessions`. Once we introduce pagination we can just throw the hook out and use this one.

client/webui/frontend/src/features/projects/api/services.ts
```diff
@@ -0,0 +1,130 @@
...
+    });
+
+    return await response.json();
+};
+
+export const removeFileFromProject = async (projectId: string, filename: string) => {
+    const url = `/api/v1/projects/${projectId}/artifacts/${encodeURIComponent(filename)}`;
```
lgh-solace: One other note: https://github.com/SolaceLabs/solace-agent-mesh/pull/626 This PR is making the usage of `configServerUrl` across all components. I believe it might be required for the change to the apis, so my guess is that this PR should follow that lead and not use relative paths.

client/webui/frontend/package.json
g-raman: Yes it would probably have to be the same version. And yes the dependency graph checks out

client/webui/frontend/src/features/projects/api/services.ts
g-raman: Yeah I think so too, I can follow up with a separate PR that handles this

client/webui/frontend/src/lib/components/projects/ProjectDescription.tsx
g-raman: Yep unused

PR #607, opened 2025-12-02
feat(DATAGO-117907): Show created artifacts in the workflow diagram

RZuchniak:
## Changes
  - Modified `save_artifact_with_metadata()` to emit `artifact_creation_progress` signals with `function_call_id` field linking artifacts to their creating tool calls
  - LLM generated artifacts: Backend emits artifact creation signal before the `notify_artifact_save` tool call is sent over.
  - Frontend handler distinguishes normal tool call signals from the LLM generated creation signals by ID prefix and queues LLM generated artifacts until tool result arrives
  - Added workflow diagram logic to create artifact nodes linked to parent tools via `function_call_id` lookup
  - Created new `GenericArtifactNode` React component for artifact visualization
  - Removed redundant `_notify_artifact_save` handler in `ChatProvider` (handled by new `artifact_creation_progress` handler)
  
 ## Considerations
  - Current implementation assumes that a tool call can only create 1 artifact. Frontend can be updated to support multiple
  - LLM generated artifacts no longer rely on the `_notify_artifact_save` tool call for completion status and metadata
  
  
image
image

### DIFF
--- a/src/solace_agent_mesh/agent/adk/callbacks.py
+++ b/src/solace_agent_mesh/agent/adk/callbacks.py
@@ -480,9 +480,19 @@ async def process_artifact_blocks_callback(
                                         host_component, a2a_context, final_progress_data
                                     )
 
-                            # Don't send completion signal here - it will be sent later
-                            # when we inject the _notify_artifact_save tool call
-                            # This ensures the signal and tool call arrive together with matching IDs
+                            # Publish completion status immediately via SSE
+                            if a2a_context:
+                                progress_data = ArtifactCreationProgressData(
+                                    filename=filename,
+                                    description=params.get("description"),
+                                    status="completed",
+                                    bytes_transferred=len(event.content),
+                                    mime_type=params.get("mime_type"),
+                                    version=version_for_tool,
+                                )
+                                await _publish_data_part_status_update(
+                                    host_component, a2a_context, progress_data
+                                )
                         else:
                             status_for_tool = "error"
                             version_for_tool = 0


src/solace_agent_mesh/agent/adk/callbacks.py
```diff
@@ -342,19 +343,9 @@ async def process_artifact_blocks_callback(
...
-                                    description=params.get("description"),
-                                    status="completed",
-                                    bytes_transferred=len(event.content),
-                                    mime_type=params.get("mime_type"),
-                                    version=version_for_tool,
-                                )
-                                await _publish_data_part_status_update(
-                                    host_component, a2a_context, progress_data
-                                )
+                            # Don't send completion signal here - it will be sent later
```
efunneko: I don't really like that we are overloading this event for all artifact creation events. There is a protocol to how they are created with a start ... updates ... end which we would be violating. My preference would be to have another event that would be just to indicate that an artifact has been save and we restore this code to what it was

src/solace_agent_mesh/agent/utils/artifact_helpers.py
```diff
@@ -306,6 +307,63 @@ async def save_artifact_with_metadata(
...
+                        status="completed",
+                        version=data_version,
+                        bytes_transferred=len(content_bytes),
+                        mime_type=mime_type,
+                        description=metadata_dict.get("description") if metadata_dict else None,
+                        function_call_id=function_call_id,
+                    )
+
+                    # Publish as status update with signal
+                    await _publish_data_part_status_update(
```
efunneko: I don't like having this in a utility file - especially calling a "_" function (I realize we do it in callbacks.py and that should be fixed).  We should define a function (probably in sac/component.py) that is called from here to notify_artifact_creation(). You should pass in an "ArtifactInfo" with all the information about the artifact and pass in an optional 'function_call_id". Then the publishing stuff can be where it belongs and this code is reduced a lot

RZuchniak: Should I update callbacks.py to use a separate function instead of `_publish_data_part_status_update()` when notifying about artifact completion, or consider it out of scope for these changes?

tests/unit/agent/utils/test_artifact_workflow_signals.py
```diff
@@ -0,0 +1,260 @@
+#!/usr/bin/env python3
+"""
+Unit tests for artifact workflow visualization signal emission.
+
+Tests verify that artifact creation properly emits signals for workflow diagram visualization:
```
efunneko: I really don't love these tests... Everything is mocked to the max - hundreds of lines of test code to basically test a couple of if statements and a function call in the artifacts helper. 

None of these tests actually prove that the event is created and sent to the gateway.

I would much prefer to have an integration test (one of the declarative ones) that actually verifies that we get the event when we expect it and that it carries the correct data. That is real proof that this feature works

RZuchniak: I decided to go with 2 tests: one for artifacts generated with a tool call, and one for llm generated artifacts. The code that runs in each case to send the event is slightly different, but if you feel this is unnecessary I can remove one of them.

PR #613, opened 2025-12-03
chore(DATAGO-118089): Platform Service Take-2

carolmorneau:
## Summary
Creates a dedicated Platform REST API service (port 8001) for platform management operations, separating them from the Web Chat Gateway. This enables independent scaling and deployment of chat vs. platform management features.

Tests will be added as a separate PR

## Usage

`sam run path/to/platform.yaml`
At this time, the platform needs to be ran by itself in its own process.

### DIFF
--- a/src/solace_agent_mesh/gateway/http_sse/dependencies.py
+++ b/src/solace_agent_mesh/gateway/http_sse/dependencies.py
@@ -396,6 +396,7 @@ async def get_user_config(
     """
     log.debug(f"get_user_config called for user_id: {user_id}")
 
+    # TODO: DATAGO-114659-split-cleanup
     gateway_context = {}
     if getattr(component, "gateway_id", None):
         gateway_context = {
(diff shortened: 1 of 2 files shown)


src/solace_agent_mesh/services/platform/alembic/.gitkeep
```diff
@@ -0,0 +1 @@
+# Alembic migration directory for community Platform Service
```
mo-radwan1: we can probably clean this file

src/solace_agent_mesh/gateway/http_sse/dependencies.py
```diff
@@ -395,11 +395,14 @@ async def get_user_config(
     FastAPI dependency to get the user-specific configuration.
     """
     log.debug(f"get_user_config called for user_id: {user_id}")
-    gateway_context = {
-        "gateway_id": component.gateway_id,
-        "gateway_app_config": app_config,
-        "request": request,
-    }
+
+    gateway_context = {}
```
mo-radwan1, line 400: Can we add a comment something like 

`# TODO: DATAGO-114659-cleanup`

So we can easily find these spots to cleanup during rollout

PR #614, opened 2025-12-03
fix(DATAGO-118574): Fix a few UI and chat issues

efunneko:
Comprehensive chat UX improvements focusing on artifact rendering, auto-scrolling, and layout refinements.

  Key changes:

  - DATAGO-118889: Artifact streaming: Auto-expand and stream markdown/text artifacts during creation; resolve math/datetime embeds in real-time
  - DATAGO-118883: Auto-scroll behavior: Fixed scroll-up detection to properly disable auto-scroll; observe content resizes (artifact expansions) to maintain scroll position
  - DATAGO-118572, DATAGO-118884, DATAGO-118574: Layout improvements: Normalized artifact heights (text/markdown: 6000px, HTML: 600px, other: 600px); increased message spacing from 2 to 5; added paragraph/list bottom
  margins; removed padding from assistant message bubbles
  - DATAGO-118885: UI polish: Files tab always shows file list; artifact bar styling with theme-aware shadows; skip rendering empty text parts
  - DATAGO-118886: SAM narrates its work too much to the user

### DIFF
--- a/client/webui/frontend/src/lib/components/chat/ChatMessage.tsx
+++ b/client/webui/frontend/src/lib/components/chat/ChatMessage.tsx
@@ -272,7 +272,7 @@ const getChatBubble = (message: MessageFE, chatContext: ChatContextValue, isLast
     const lastPartKind = groupedParts[lastPartIndex]?.kind;
 
     return (
-        <div key={message.metadata?.messageId} className="space-y-5">
+        <div key={message.metadata?.messageId} className="space-y-6">
             {/* Render parts in their original order to preserve interleaving */}
             {groupedParts.map((part, index) => {
                 const isLastPart = index === lastPartIndex;
--- a/client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
+++ b/client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
@@ -469,7 +469,7 @@ export const ChatInputArea: React.FC<{ agents: AgentCardInfo[]; scrollToBottom?:
 
     return (
         <div
-            className={`rounded-lg border p-4 shadow-sm ${isDragging ? "border-dotted border-[var(--primary-wMain)] bg-[var(--accent-background)]" : ""}`}
+            className={`bg-card rounded-lg border p-4 shadow-sm ${isDragging ? "border-dotted border-[var(--primary-wMain)] bg-[var(--accent-background)]" : ""}`}
             onDragEnter={handleDragEnter}
             onDragOver={handleDragOver}
             onDragLeave={handleDragLeave}
--- a/client/webui/frontend/src/lib/components/pages/ChatPage.tsx
+++ b/client/webui/frontend/src/lib/components/pages/ChatPage.tsx
@@ -10,7 +10,7 @@ import { Button, ChatMessageList, CHAT_STYLES, Badge } from "@/lib/components/ui
 import { Spinner } from "@/lib/components/ui/spinner";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/lib/components/ui/tooltip";
 import { ResizablePanelGroup, ResizablePanel, ResizableHandle } from "@/lib/components/ui/resizable";
-import { useChatContext, useTaskContext } from "@/lib/hooks";
+import { useChatContext, useTaskContext, useThemeContext } from "@/lib/hooks";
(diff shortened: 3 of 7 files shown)


lgh-solace:
Testing the branch, I'm seeing a double version of an artifact with a single question and sentences with no spaces between them. I realize the former may be my model (`openai/gemini-2.5-pro`) and the latter is a pre-existing issue, but not sure if you might make a tweak here to improve either behaviour?

image

efunneko:
> Testing the branch, I'm seeing a double version of an artifact with a single question and sentences with no spaces between them. I realize the former may be my model (`openai/gemini-2.5-pro`) and the latter is a pre-existing issue, but not sure if you might make a tweak here to improve either behaviour?
 

That is interesting. It looks like it created the report and then second-guessed itself to actually include Pluto. To do that, it would have done an edit on the doc which would create a second version. I think that is all okay, but I don't love the text in the chat. I was hoping to get away from that chattiness (works great with Sonnet 4.5).

lgh-solace:
Minor issue? It seems this `User` information has changed? There's now a user unfriendly prefix? I'm not 100% sure this issue came with the PR but... seems likely?

image

lgh-solace:
Minor UI suggestion? Could the header have the full border while streaming? The caret looks a bit sad and lost.

image

Also, it will make it nicer when closed...

image

lgh-solace:
After switching models, I'm have not seen duplicates, but I did get this scenario where an odd intermediate artifact was returned to the user that had mis-matching mime type to extension.

image

client/webui/frontend/src/lib/components/chat/ChatMessage.tsx
```diff
@@ -272,16 +272,30 @@ const getChatBubble = (message: MessageFE, chatContext: ChatContextValue, isLast
...
             {/* Render parts in their original order to preserve interleaving */}
             {groupedParts.map((part, index) => {
                 const isLastPart = index === lastPartIndex;
 
                 if (part.kind === "text") {
+                    // Skip rendering empty or whitespace-only text parts
+                    const textContent = (part as TextPart).text;
+                    if (!textContent || !textContent.trim()) {
+                        // If this is the last part and it's empty, still render actions if needed
+                        if (isLastPart && (showWorkflowButton || showFeedbackActions)) {
```
lgh-solace, line 285: What about the other buttons, like copy, etc. Shouldn't the menu actions still be shown?

client/webui/frontend/src/lib/components/chat/artifact/ArtifactBar.tsx
```diff
@@ -40,6 +40,7 @@ export interface ArtifactBarProps {
 
 export const ArtifactBar: React.FC<ArtifactBarProps> = ({ filename, description, mimeType, size, status, expandable = false, expanded = false, onToggleExpand, actions, bytesTransferred, error, expandedContent, context = "chat", isDeleted = false, version, source }) => {
     const [contentForAnimation, setContentForAnimation] = useState(expandedContent);
+    const [isDarkMode, setIsDarkMode] = useState(() => document.documentElement.classList.contains('dark'));
```
lgh-solace, line 43: Could we not use something like... `const { currentTheme } = useThemeContext();`

client/webui/frontend/src/lib/components/chat/artifact/ArtifactBar.tsx
```diff
@@ -135,9 +150,38 @@ export const ArtifactBar: React.FC<ArtifactBarProps> = ({ filename, description,
         }
     };
 
+    // Define shadow and background colors based on theme
+    // Light mode: background-w10, shadow using secondary-w8040
+    // Dark mode: background-wMain, shadow using primary-w90 (darker shadows)
+    const backgroundColor = isDarkMode ? 'var(--color-background-wMain)' : 'var(--color-background-w10)';
```
lgh-solace, line 156: We might be able to tidy this up after, extracting things, but not for this PR.

client/webui/frontend/src/lib/components/chat/ChatMessage.tsx
efunneko: I am not totally following. If there is no text, we won't create the ChatBubble (which just leads to a bit of blank space), but we will show the trailing feedback, workflow and copy buttons

PR #622, opened 2025-12-04
feat(DATAGO-118497, DATAGO-118496): Platform Service independence and shared utilities

mo-radwan1:
## Summary

Completes backend architecture split for Platform Service, making it fully independent from WebUI Gateway.

**Stories Completed:**
- ✅ DATAGO-118497: Platform service receives agent heartbeats
- ✅ DATAGO-118496: Platform validates bearer tokens with OAuth
- ✅ Architecture improvement: Shared utilities layer

### Background Tasks Migrated
- Added broker connection for heartbeat monitoring and agent registry
- Implemented heartbeat listener (monitors deployer status)
- Implemented deployment status checker (runs every 60s)
- Added agent registry for deployment verification
- All background tasks now run in Platform Service, not WebUI Gateway

### Message Publishing Added
- Implemented `publish_a2a()` method
- Sends deployment commands to deployer
- Topics: `{namespace}/deployer/agent/{id}/{deploy|update|undeploy}`

### Message Receiving
- Receives deployer heartbeats via `HeartbeatListener`
- Receives agent cards via `AgentRegistry`
- Complete deployment monitoring capability

### Configuration Added
- `deployment_timeout_minutes` (default: 5)
- `heartbeat_timeout_seconds` (default: 90)
- `deployment_check_interval_seconds` (default: 60)
- Created `examples/services/platform_service_example.yaml`

## OAuth Implementation (DATAGO-118496)

- Extracted OAuth middleware from http_sse to `shared/auth/middleware.py`
- Platform Service now validates bearer tokens with external OAuth service
- Same auth implementation for both gateways and services
- Supports production mode (real validation) and dev mode

### Created `src/solace_agent_mesh/shared/` Module
Moved 13 files from `http_sse/shared/` to enable code reuse:

- **api/**: pagination, response_utils, auth_utils
- **database/**: base_repository, database_exceptions, database_helpers
- **exceptions/**: exceptions, exception_handlers, error_dto
- **auth/**: dependencies (ValidatedUserConfig), middleware (OAuth)
- **utils/**: timestamp_utils, enums, types, utils

### Benefits
- Gateways and services import from `shared/` (no cross-dependencies)
- Platform Service completely independent of http_sse
- Clean architectural boundary
- Enables future services (billing, monitoring, etc.)

### Removed from http_sse
- ❌ Enterprise router mounting (no more `/api/v1/enterprise/*`)
- ❌ Background task startup/shutdown
- ❌ Platform migration calls
- ❌ Deployment monitoring dependencies

### Result
- WebUI Gateway serves chat endpoints only
- Single database (`WEB_UI_GATEWAY_DATABASE_URL`)
- No platform functionality

## Breaking Changes

- `http_sse/shared/` module removed (raises ImportError with migration guide)
- Enterprise routers no longer mounted in WebUI Gateway
- Platform database no longer used by WebUI Gateway
- **Users must run Platform Service separately for agent management**

## Architecture: Service vs Gateway

**SERVICES** (Platform Service):
- Provide internal platform functionality
- Admin-facing CRUD operations
- Background processing
- Example: `services/platform/`

**GATEWAYS** (WebUI, Slack, etc.):
- Handle external communication channels
- User-facing interactions
- Protocol-specific implementations
- Example: `gateway/http_sse/`

### Before
```bash
# Single service on port 8000
sam run configs/gateways/webui.yaml
```

### After
```bash
# Terminal 1: Chat service
sam run configs/gateways/webui.yaml  # Port 8000

# Terminal 2: Platform service
sam run configs/services/platform.yaml  # Port 8001
```

## Test Plan

- [ ] Platform Service starts with broker connection
- [ ] Background tasks initialize (heartbeat listener, deployment checker)
- [ ] OAuth middleware validates tokens correctly
- [ ] Deployment operations publish messages to deployer
- [ ] Agent registry tracks agent presence
- [ ] WebUI Gateway starts without platform functionality
- [ ] Shared utilities import correctly
- [ ] All unit tests pass
- [ ] Integration tests pass

## Related PRs

- Enterprise PR: (to be created with matching changes)
- Frontend PR: DATAGO-118499 (routing to platform service)

## Next Steps

1. Frontend routing to Platform Service (DATAGO-118499)
2. Helm chart updates (DATAGO-118498)
3. Docker/K8s configurations (DATAGO-118706)
4. Documentation (DATAGO-118501, DATAGO-118708)

### DIFF
--- a/examples/services/platform_service_example.yaml
+++ b/examples/services/platform_service_example.yaml
@@ -22,7 +22,7 @@ apps:
     app_config:
       # --- Required ---
       namespace: ${NAMESPACE} # Namespace for A2A communication
-      database_url: "${PLATFORM_DATABASE_URL, sqlite:///platform-service.db}" # Platform database (agents, connectors, deployments)
+      database_url: "${PLATFORM_DATABASE_URL, sqlite:///platform.db}" # Platform database (agents, connectors, deployments)
 
       # --- FastAPI Server Configuration ---
       fastapi_host: ${PLATFORM_API_HOST, localhost}
--- a/preset/agents/basic/platform.yaml
+++ b/preset/agents/basic/platform.yaml
@@ -19,7 +19,7 @@ apps:
 
     app_config:
       namespace: ${NAMESPACE}
-      database_url: "${PLATFORM_DATABASE_URL, sqlite:///platform-service.db}"
+      database_url: "${PLATFORM_DATABASE_URL, sqlite:///platform.db}"
 
       fastapi_host: ${PLATFORM_API_HOST, localhost}
       fastapi_port: ${PLATFORM_API_PORT, 8001}
--- a/templates/platform.yaml
+++ b/templates/platform.yaml
@@ -20,7 +20,7 @@ apps:
     app_config:
       # --- Required ---
       namespace: ${NAMESPACE}
-      database_url: "${PLATFORM_DATABASE_URL, sqlite:///platform-service.db}"
+      database_url: "${PLATFORM_DATABASE_URL, sqlite:///platform.db}"
 
       # --- FastAPI Server Configuration ---
       fastapi_host: ${PLATFORM_API_HOST, localhost}


src/solace_agent_mesh/gateway/http_sse/app.py
```diff
@@ -193,16 +193,19 @@ class WebUIBackendApp(BaseGatewayApp):
             "required": False,
             "type": "dict",
             "default": {},
-            "description": "Configuration for the Platform Service (enterprise features: agents, connectors, deployments).",
+            "description": "Configuration for connecting to the Platform Service (runs separately on port 8001).",
             "dict_schema": {
-                "database_url": {
```
carolmorneau, line 198: just confirming that removing the config doesn't break anything

src/solace_agent_mesh/services/platform/component.py
```diff
@@ -233,3 +435,53 @@ def get_session_manager(self) -> _StubSessionManager:
...
+            payload: Message payload dictionary
+            user_properties: Optional user properties for message metadata
+
+        Raises:
+            Exception: If publishing fails
+        """
+        log.debug("%s Publishing A2A message to topic: %s", self.log_identifier, topic)
+
+        try:
+            super().publish_a2a_message(payload, topic, user_properties)
```
carolmorneau: I believe we need to extend SamComponentBase to call this?

src/solace_agent_mesh/gateway/http_sse/app.py
mo-radwan1: Yeah, it shouldnt break anything, it would just be ignored if it exists

src/solace_agent_mesh/services/platform/component.py
mo-radwan1: You are right good catch, let me fix that

client/webui/frontend/src/lib/api/client.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    private async request(baseUrl: string, endpoint: string, options?: RequestInit) {
+        this.ensureConfigured();
+        const url = `${baseUrl}${endpoint}`;
+
+        if (options?.body && typeof options.body !== "string" && !(options.body instanceof FormData)) {
+            return fetchWithError(url, options);
+        }
+
+        return fetchJsonWithError(url, options);
```
lgh-solace: We're making the assumption that we're dealing with JSON, which I think is ok for almost all apis... but not all of them... 🤔

client/webui/frontend/src/lib/api/client.ts
```diff
@@ -0,0 +1,179 @@
...
+                    "Content-Type": "application/json",
+                    ...options?.headers,
+                },
+                body: JSON.stringify(body),
+            });
+        },
+    };
+
+    platform = {
+        get: (endpoint: string) => this.request(this.platformBaseUrl, endpoint),
```
lgh-solace: It's a bit hard to tell, but this looks identical to the "chat" object except for the different base URL. If so, can we extract object creation to some kind of factory function so the code isn't duplicated?

client/webui/frontend/src/lib/api/client.ts
```diff
@@ -0,0 +1,179 @@
...
+        const url = `${baseUrl}${endpoint}`;
+
+        if (options?.body && typeof options.body !== "string" && !(options.body instanceof FormData)) {
+            return fetchWithError(url, options);
+        }
+
+        return fetchJsonWithError(url, options);
+    }
+
+    chat = {
```
lgh-solace: Just curious about the choice of `chat` for this bucket? Is it all the UI endpoints? Just wondering if there might be a different term? Only concerned about confusion for folks that might think it should only be used within the context of the chat view?

client/webui/frontend/src/lib/api/client.ts
```diff
@@ -0,0 +1,179 @@
...
+        if (!this.configured) {
+            throw new Error("API client not configured. Call api.configure() from ConfigProvider first.");
+        }
+    }
+
+    private async request(baseUrl: string, endpoint: string, options?: RequestInit) {
+        this.ensureConfigured();
+        const url = `${baseUrl}${endpoint}`;
+
+        if (options?.body && typeof options.body !== "string" && !(options.body instanceof FormData)) {
```
lgh-solace: I'm not sure I understand the choice of when to use `fetchWithError` vs `fetchJsonWithError`?

client/webui/frontend/src/lib/components/activities/FlowChartDetails.tsx
```diff
@@ -71,7 +72,7 @@ export const FlowChartDetails: React.FC<{ task: VisualizedTask }> = ({ task }) =
 
     const handleDownloadStim = async () => {
         try {
-            const response = await fetchWithError(`${apiPrefix}/tasks/${task.taskId}`);
+            const response = await fetchWithError(`${chatBaseUrl}/api/v1/tasks/${task.taskId}`);
```
lgh-solace: It would be nice to clean these up as well? We can't do `await api.chat.get(...);` because it will try to parse a json response, correct? 

Can we consider having a 'get' variation that does not parse the response?  Thoughts?

client/webui/frontend/src/lib/providers/ConfigProvider.tsx
```diff
@@ -105,7 +107,8 @@ export function ConfigProvider({ children }: Readonly<ConfigProviderProps>) {
 
                 // Map backend fields to ConfigContextValue fields
                 const mappedConfig: ConfigContextValue = {
-                    configServerUrl: data.frontend_server_url,
+                    chatServerUrl: data.frontend_server_url,
```
bwiebe-solace: Should we be renaming the backend variables to use "chat" as well to keep them aligned and avoid potential confusion?

client/webui/frontend/src/lib/api/client.ts
mo-radwan1: Yeah, thats a good idea. Pushed the changed

client/webui/frontend/src/lib/api/client.ts
mo-radwan1: Updated to webui

client/webui/frontend/src/lib/api/client.ts
mo-radwan1: From the code, although I have not dug deeper into it but this is what it looks like 
```
  | Function           | Returns             | Use Case                                        |
  |--------------------|---------------------|-------------------------------------------------|
  | fetchWithError     | Raw Response object | Streaming, binary downloads, non-JSON responses |
  | fetchJsonWithError | Parsed JSON data    | Standard REST API calls                         |
  ```

File: client/webui/frontend/src/lib/components/activities/FlowChartDetails.tsx (hunk 7 above)
Comment (mo-radwan1) [reply client/webui/frontend/src/lib/components/activities/FlowChartDetails.tsx]: Yeah we can add that too something like `api.webui.get('/tasks/123', { raw: true })` just wanted to avoid confusion but yeah its a good idea

File: preset/agents/basic/webui.yaml [hunk 9]
```diff
@@ -57,6 +57,8 @@ apps:
             Format responses to the user in Markdown using appropriate formatting.
 
       # --- Frontend Config Passthrough ---
+      frontend_server_url: "${FRONTEND_SERVER_URL}"
+      frontend_enterprise_server_url: "${ENTERPRISE_API_URL}"
```
bwiebe-solace: I think we need to change this line to the below one to align with the other changes with variable names
```suggestion
      platform_service_url: "${PLATFORM_SERVICE_URL}"
```

client/webui/frontend/src/lib/components/chat/PromptsCommand.tsx
```diff
@@ -40,6 +40,7 @@ interface PromptsCommandProps {
 
 export const PromptsCommand: React.FC<PromptsCommandProps> = ({ isOpen, onClose, textAreaRef, onPromptSelect, messages = [], onReservedCommand }) => {
     const navigate = useNavigate();
+    // Migrated to api client
```
lgh-solace: Could you remove comments like this one?

client/webui/frontend/src/lib/components/chat/PromptsCommand.tsx
```diff
@@ -59,7 +60,7 @@ export const PromptsCommand: React.FC<PromptsCommandProps> = ({ isOpen, onClose,
         const fetchPromptGroups = async () => {
             setIsLoading(true);
             try {
-                const data = await fetchJsonWithError("/api/v1/prompts/groups/all");
+                const data = await api.webui.get(`/api/v1/prompts/groups/all`);
```
lgh-solace: I suppose it doesn't matter now - it used to be that backticks were slightly less efficient that quotes, but the difference is supposed to be negligible now.

client/webui/frontend/src/lib/api/client.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    return response;
+};
+
+const fetchJsonWithError = async (url: string, options: RequestInit = {}) => {
+    const response = await fetchWithError(url, options);
+    return response.json();
+};
+
+type InternalRequestOptions = RequestOptions & RequestInit & { raw?: boolean };
```
lgh-solace: I have been debating `raw`. I don't mind it, but wondering if we could pick something clearer? How do you feel about something like...?

`{ fullResponse?: boolean }`

client/webui/frontend/src/lib/components/chat/file/ArtifactMessage.tsx
```diff
@@ -269,8 +269,7 @@ export const ArtifactMessage: React.FC<ArtifactMessageProps> = props => {
                     apiUrl = `/api/v1/artifacts/null/${encodeURIComponent(filename)}/versions/${version || "latest"}`;
                 }
 
-                const response = await authenticatedFetch(apiUrl);
-                if (!response.ok) throw new Error(`Failed to fetch artifact content: ${response.statusText}`);
+                const response = await api.webui.get(apiUrl, { raw: true });
```
lgh-solace: This logic has changed? There's no longer an error thrown if `raw` is true?  I think we need to have the api do some error handling and throw an error when response is not `ok` or leave this handling here. Preference would be to have an error thrown although I know we have a few cases with special error handling so maybe we need another "optional" option?

client/webui/frontend/src/lib/components/chat/PromptsCommand.tsx
mo-radwan1: Updated to remove unnecessary backticks

client/webui/frontend/src/lib/api/client.ts
mo-radwan1: Updated

client/webui/frontend/src/lib/components/chat/file/ArtifactMessage.tsx
mo-radwan1: yeah, you are right, I have gone back and compared with main branch and added back all the error handling for the full response since they seem to have a detailed response

preset/agents/basic/webui.yaml
mo-radwan1: Yeah, good catch let me update that file it looks like it was missed

client/webui/frontend/src/lib/components/prompts/PromptBuilderChat.tsx
```diff
@@ -89,53 +88,32 @@ export const PromptBuilderChat: React.FC<PromptBuilderChatProps> = ({ onConfigUp
...
+                        }
 
-                            // Scroll to bottom after AI response
-                            setTimeout(() => scrollToBottom(), 100);
-                        } else {
-                            const errorData = await chatResponse.json().catch(() => ({}));
-                            console.error("Prompt builder API error:", errorData);
+                        onReadyToSave(chatData.ready_to_save);
 
-                            const errorMessage: Message = {
```
lgh-solace, line 132: I guess we've lost this specialized handling? I think it's ok... seems to go through a similar flow with only slightly different error message.

client/webui/frontend/src/lib/hooks/useBackgroundTaskMonitor.ts
```diff
@@ -77,25 +76,20 @@ export function useBackgroundTaskMonitor({ apiPrefix, userId, onTaskCompleted, o
...
-                const response = await authenticatedFetch(`${apiPrefix}/tasks/${taskId}/status`);
-                if (!response.ok) {
-                    if (response.status === 404) {
-                        unregisterBackgroundTask(taskId);
-                        return null;
-                    }
-                    throw new Error(`HTTP ${response.status}`);
+                return await api.webui.get(`/api/v1/tasks/${taskId}/status`);
+            } catch (error: unknown) {
+                if (error instanceof Error && error.message.includes("404")) {
```
lgh-solace: You've verified the message will contain a 404? if this is a case where the full response should be returned?

client/webui/frontend/src/lib/providers/AuthProvider.tsx
```diff
@@ -90,20 +73,13 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
...
         try {
             if (configUseAuthorization) {
-                const response = await authenticatedFetch("/api/v1/auth/logout", {
-                    method: "POST",
-                    headers: {
-                        "Content-Type": "application/json",
-                        Accept: "application/json",
-                    },
-                });
+                const response = await api.webui.post("/api/v1/auth/logout", undefined, { fullResponse: true });
```
lgh-solace: Might not need the `fullResponse: true` here?

client/webui/frontend/src/lib/providers/ChatProvider.tsx
```diff
@@ -675,28 +649,35 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
...
+                    versionsUrl = `/api/v1/artifacts/${sessionId}/${encodeURIComponent(artifactFilename)}/versions`;
                 } else if (activeProject?.id) {
-                    versionsUrl = `${apiPrefix}/artifacts/null/${encodeURIComponent(artifactFilename)}/versions?project_id=${activeProject.id}`;
+                    versionsUrl = `/api/v1/artifacts/null/${encodeURIComponent(artifactFilename)}/versions?project_id=${activeProject.id}`;
                 } else {
                     throw new Error("No valid context for artifact preview");
                 }
 
-                const availableVersions: number[] = await fetchJsonWithError(versionsUrl);
+                const versionsResponse = await api.webui.get(versionsUrl, { fullResponse: true });
```
lgh-solace: Do we need `fullResponse: true` here?

client/webui/frontend/src/lib/providers/ChatProvider.tsx
```diff
@@ -1561,29 +1533,17 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
...
-                        await authenticatedFetch(`${apiPrefix}/tasks/${currentTaskId}:cancel`, {
-                            method: "POST",
-                            headers: { "Content-Type": "application/json" },
-                            body: JSON.stringify(cancelRequest),
-                            credentials: "include",
-                        });
-                    } catch (error) {
-                        console.warn(`${log_prefix} Failed to cancel current task:`, error);
-                    }
+                console.log(`${log_prefix} Cancelling current task ${currentTaskId}`);
```
lgh-solace: This logic changed, no longer checking for `const isBackground = isTaskRunningInBackground(currentTaskId);`?

client/webui/frontend/src/lib/providers/ChatProvider.tsx
```diff
@@ -1671,50 +1631,13 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
...
         async (sessionId: string, newName: string) => {
             try {
-                const response = await authenticatedFetch(`${apiPrefix}/sessions/${sessionId}`, {
-                    method: "PATCH",
-                    headers: { "Content-Type": "application/json" },
-                    body: JSON.stringify({ name: newName }),
-                });
-
-                // Special handling for 422 validation errors
-                if (response.status === 422) {
```
lgh-solace, line 1704: Do we need this special handling for 422 errors?

client/webui/frontend/src/lib/providers/ChatProvider.tsx
```diff
@@ -1850,11 +1764,11 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
                 return;
             }
             try {
-                await submitFeedback({
-                    taskId: taskId,
-                    sessionId: sessionId,
-                    feedbackType: feedbackType,
-                    feedbackText: feedbackText,
+                await api.webui.post("/api/v1/feedback", {
```
lgh-solace, line 1779: Just curious, did you remove the `submitFeedback` function?

client/webui/frontend/src/lib/providers/ChatProvider.tsx
```diff
@@ -2378,48 +2270,9 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
     }, [handleSseMessage, handleSseOpen, handleSseError]);
 
     useEffect(() => {
-        if (currentTaskId && apiPrefix) {
+        if (currentTaskId) {
             const accessToken = getAccessToken();
-
-            // Check if this is a reconnection to a background task
-            // Use a ref to get the latest background tasks without triggering re-renders
-            const bgTask = backgroundTasksRef.current.find(t => t.taskId === currentTaskId);
```
lgh-solace, line 2386: background task logic removed?

client/webui/frontend/src/lib/providers/TaskProvider.tsx
```diff
@@ -134,32 +136,11 @@ export const TaskProvider: React.FC<TaskProviderProps> = ({ children }) => {
...
-            const subscribeResponse = await authenticatedFetch(`${apiPrefix}/visualization/subscribe`, {
-                method: "POST",
-                headers: { "Content-Type": "application/json" },
-                body: JSON.stringify(subscribePayload),
-                credentials: "include",
+            const subscriptionData: SubscriptionResponse = await api.webui.post("/api/v1/visualization/subscribe", {
+                subscription_targets: [{ type: "my_a2a_messages" }],
             });
-            if (!subscribeResponse.ok) {
-                const errorData = await subscribeResponse.json().catch(() => ({ detail: "Failed to subscribe" }));
```
lgh-solace, line 145: We've lost some error specifics here, but I think this is ok.

client/webui/frontend/src/lib/providers/ProjectProvider.tsx
```diff
@@ -70,54 +56,26 @@ export const ProjectProvider: React.FC<{ children: React.ReactNode }> = ({ child
...
-                const response = await authenticatedFetch(`${apiPrefix}/projects`, {
-                    method: "POST",
-                    // No 'Content-Type' header, browser will set it for FormData
-                    body: projectData,
-                    credentials: "include",
-                });
+            const newProject: Project = await api.webui.post("/api/v1/projects", projectData);
 
-                if (!response.ok) {
-                    const responseText = await response.text();
```
lgh-solace, line 90: Again, we may have lost some error specification, but I think that's ok!

client/webui/frontend/src/lib/providers/ProjectProvider.tsx
```diff
@@ -246,74 +125,22 @@ export const ProjectProvider: React.FC<{ children: React.ReactNode }> = ({ child
...
+            const updatedProject: Project = await api.webui.put(`/api/v1/projects/${projectId}`, data);
 
-                if (!response.ok) {
-                    let errorMessage = `Failed to update project: ${response.statusText}`;
-
-                    try {
-                        const errorData = await response.json();
-
-                        // Handle validation errors (422)
-                        if (response.status === 422) {
```
lgh-solace, line 264: Lost special 422 error handling? Not sure if we need to keep it, but just noting...

lgh-solace (commented):
Looks good! A nice clean up! 
Please just verify the few places where logic seems to have changed?

client/webui/frontend/src/lib/components/prompts/PromptBuilderChat.tsx
mo-radwan1: The new api.webui.post() throws errors for non-ok responses, which get caught by the outer catch block and show the fallback message. The behavior is functionally the same - just the error message might be slightly different.

client/webui/frontend/src/lib/hooks/useBackgroundTaskMonitor.ts
mo-radwan1: Yeah you are right about this one, we should use the full response. I will update it

client/webui/frontend/src/lib/providers/AuthProvider.tsx
mo-radwan1: Updated

client/webui/frontend/src/lib/providers/ChatProvider.tsx
mo-radwan1: Yes, all fullResponse: true usages are intentional:
  - Line 517 (/api/v1/artifacts/upload): Needs raw response to check response.status === 413 for file size errors
  - Lines 673, 737, 827 (artifact content): Needs raw response for response.blob() to get binary content
  - Line 1642 (sessions PATCH): Needs raw response to check response.status === 422 for validation errors
  - Line 1749 (task cancel): Needs raw response to check response.status === 202 for accepted status

  All cases require either specific status code checks or raw response methods that aren't available through the default JSON-parsing path.

client/webui/frontend/src/lib/providers/ChatProvider.tsx
mo-radwan1: Restored. The isTaskRunningInBackground(currentTaskId) check is now present at lines 1444 and 1536. Both locations check before cancelling tasks to ensure background tasks aren't incorrectly cancelled when switching sessions or starting new chats

client/webui/frontend/src/lib/providers/ChatProvider.tsx
mo-radwan1: Yes, kept the 422 handling. It provides a cleaner "Invalid name" error message for validation errors instead of exposing raw backend error details. The 422 status is the standard HTTP response for validation failures, and this handling was in the original code for better UX.

client/webui/frontend/src/lib/providers/ChatProvider.tsx
mo-radwan1: The standalone submitFeedback utility function was replaced with an inline API call using the new unified API client. The functionality is preserved in handleFeedbackSubmit (lines 1772-1793) which calls api.webui.post("/api/v1/feedback", {...}). Same behavior, just using the consolidated API client pattern.

client/webui/frontend/src/lib/providers/ChatProvider.tsx
mo-radwan1: Restored. The background task reconnection logic is present at lines 2288-2309:
  - Detects if current task is a background task (line 2288-2289)
  - Adds reconnect=true and last_event_timestamp=0 params for reconnection (lines 2297-2298)
  - Clears previous messages for the task when reconnecting to avoid duplicates (lines 2301-2308)

client/webui/frontend/src/lib/providers/TaskProvider.tsx
mo-radwan1: Restored the detailed error handling. Lines 146-156 now handle:
  - authorization_failure errors with message and suggested_action
  - subscription_failure errors with message and suggested_action
  - Generic fallback for other error types

  Using fullResponse: true to access the raw response and parse the structured error data.

client/webui/frontend/src/lib/providers/ProjectProvider.tsx
mo-radwan1: The detailed error handling is preserved in addFilesToProject (lines 89-108):
  - Parses JSON error response for detail or message
  - Falls back to plain text if JSON parsing fails (with 500 char limit)
  - Special handling for 413 status with user-friendly file size error message

  Also fixed the API client itself to properly handle 204 No Content responses and improve error message extraction for all endpoints.

client/webui/frontend/src/lib/providers/ProjectProvider.tsx
mo-radwan1: Restored the 422 Pydantic validation error handling in updateProject (lines 159-172):
  - Detects 422 status code
  - Parses Pydantic error array format (detail[].loc and detail[].msg)
  - Formats as user-friendly "Validation error: field: msg" message
  - Handles string detail as fallback

  This provides meaningful validation feedback to users instead of generic error messages.

lgh-solace (approved):
Approving for FE changes, thanks Mo!

PR #628, opened 2025-12-09
feat Refreshing Artifact Preview

RZuchniak:
- When viewing an artifact in preview mode, the UI automatically detects and displays new versions as they're created without requiring manual refresh
- Monitors the artifacts array for changes and compares the latest version of the previewed artifact against the current max version in the version dropdown
- When a new version is detected, fetches the updated version list and the latest version's content, then updates the preview state to display the new version

https://github.com/user-attachments/assets/73ee6309-19ca-4335-8e59-fe4b68d629c4

### DIFF
--- a/client/webui/frontend/src/lib/providers/ChatProvider.tsx
+++ b/client/webui/frontend/src/lib/providers/ChatProvider.tsx
@@ -721,16 +721,19 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
     );
 
     const navigateArtifactVersion = useCallback(
-        async (artifactFilename: string, targetVersion: number): Promise<FileAttachment | null> => {
+        async (artifactFilename: string, targetVersion: number, availableVersions?: number[]): Promise<FileAttachment | null> => {
+            // Use provided versions or fall back to state
+            const versions = availableVersions || previewedArtifactAvailableVersions;
+
             // If versions aren't loaded yet, this is likely a timing issue where this was called
             // before openArtifactForPreview completed. Just silently return - the artifact will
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/providers/ChatProvider.tsx
```diff
@@ -1888,6 +1897,118 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({ children }) => {
...
+            const versionToFetch = artifactLatestVersion;
+
+            // Mark as in-progress to prevent duplicate fetches
+            artifactFetchInProgressRef.current.add(currentFilename);
+
+            // Refresh versions and switch to latest
+            (async () => {
+                try {
+                    let versionsUrl: string;
+                    if (sessionId && sessionId.trim() && sessionId !== "null" && sessionId !== "undefined") {
```
lgh-solace, line 2206: 🤔 I'm a little nervous about the complexity here. I wonder if there's a simpler way to provide this kind of functionality? Further, this code looks very similar to the code within `navigateArtifactVersion`. Minimally, can you look to see if you can reuse that function or extract duplicated code to a utility function?

RZuchniak: Sorry about the complexity, I was running into some issues with the refresh not working properly when you preview the artifact while it's streaming. Copilot also had some issues with race conditions and state data that added to it as well. I've switched over to using the function you mentioned, but I had to make a small change to support it

lgh-solace:
Closing this one  due to changes in ChatProvider - will open a new PR.

PR #629, opened 2025-12-09
DATAGO-119360 FE | Fix form caching issue in prompts UI

amir-ghasemi:
This pull request introduces improvements to both the frontend and backend handling of prompt building, focusing on ensuring a fresh state for the prompt builder UI and making the backend more robust to variations in LLM response formats.

Frontend improvements for prompt builder state management:

* Added a `builderKey` state to `PromptsPage.tsx` and increment it when starting a new prompt, ensuring that the `PromptTemplateBuilder` component is always freshly mounted for each new prompt creation or editing session. [[1]](diffhunk://#diff-e7d441ec656259289d068051e99575b0808a0c1537ee9a193c690497e0be0fdaR28) [[2]](diffhunk://#diff-e7d441ec656259289d068051e99575b0808a0c1537ee9a193c690497e0be0fdaR78-R84)
* Updated the `PromptTemplateBuilder` component to use a dynamic `key` prop based on `builderKey` and the current group, further guaranteeing a fresh component instance when needed.

Backend robustness for LLM response parsing:

* Enhanced the LLM response parsing logic in `prompt_builder_assistant.py` to handle arbitrary nested keys (not just "response"), unwrapping the structure if it contains a single key with a dict value that includes a "message" field. This makes the backend more resilient to different LLM response formats.

### DIFF
--- a/client/webui/frontend/src/lib/components/pages/PromptsPage.tsx
+++ b/client/webui/frontend/src/lib/components/pages/PromptsPage.tsx
@@ -25,6 +25,7 @@ export const PromptsPage: React.FC = () => {
     const [initialMessage, setInitialMessage] = useState<string | null>(null);
     const [editingGroup, setEditingGroup] = useState<PromptGroup | null>(null);
     const [builderInitialMode, setBuilderInitialMode] = useState<"manual" | "ai-assisted">("ai-assisted");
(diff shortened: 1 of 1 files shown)


amir-ghasemi:
> Thanks! Question: Can these changes be delivered separately to enterprise or do they need to go together?

have not tested with enterprise yet but I don't see anything requiring them to be in sync.

PR #630, opened 2025-12-09
ci(DATAGO-119418): fix release job

artyom-morozov:
- Added always() to the release job's if condition to ensure it evaluates even when security-checks is skipped
- Updated condition to allow release when security-checks is either `success` OR `skipped`
- Updated AWS secrets to use MANIFEST_READ_ONLY_* credentials

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,7 +49,10 @@ jobs:
   # Release to PyPI using Trusted Publishing
   release:
     name: Release to PyPI
-    # Run if security checks passed OR if security checks were skipped
+    # We need always() here because when security-checks is skipped (via skip_security_checks input),
+    # GitHub Actions would by default skip ALL dependent jobs. The always() forces evaluation of our
+    # condition, allowing the release to proceed when security-checks was either successful OR skipped.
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -42,13 +42,18 @@ jobs:
...
+      MANIFEST_AWS_SECRET_ACCESS_KEY: ${{ secrets.MANIFEST_READ_ONLY_AWS_SECRET_ACCESS_KEY }}
       MANIFEST_AWS_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
 
   # Release to PyPI using Trusted Publishing
   release:
     name: Release to PyPI
+    # Run if security checks passed OR if security checks were skipped
+    needs: security-checks
+    if: |
+      always() &&
```
johnvincentcorpuz, line 58: @artyom-morozov could you. add on the PR description why is there a need to add `always()` here. 

On our release workflow, even if the security-check is skipped (aborted), we still want to release correct?

artyom-morozov: Updated the comment and description

PR #637, opened 2025-12-10
chore(DATAGO-119476): update docs for supabase troubleshooting

g-raman:


docs/docs/documentation/enterprise/connectors/connectors.md
```diff
@@ -14,12 +14,17 @@ SQL connectors enable agents to query and analyze database information using nat
...
 - MySQL
 - PostgreSQL
 - MariaDB
 
 Each database type follows the same configuration process but may have specific connection string requirements or authentication methods.
 
+:::tip
+For Supabase, use the Session Pooler string. Refer to the [Troubleshooting](#troubleshooting) section for more info.
+:::
+
```
solace-mdupls: "For Supabase" makes it sound like we've mentioned this in the docs before. I would suggest that you can omit this section - the troubleshooting guide should be sufficient.

PR #651, opened 2025-12-15
chore(DATAGO-119430): mcp upgrade in lock file

lgh-solace:
Upgrading mcp dependency in uv lock file to secure version for: 

cve: CVE-2025-66416
package_name: mcp
package_version: 1.10.1
severity: high
vulnerability_type: app_dependency
description: The MCP Python SDK, called `mcp` on PyPI, is a Python implementation of the Model Context Protocol (MCP). Prior to version 1.23.0, tThe Model Context Protocol (MCP) Python SDK does not enable DNS rebinding protection by default for HTTP-based servers. When an HTTP-based MCP server is run on localhost without authentication using FastMCP with streamable HTTP or SSE transport, and has not configured TransportSecuritySettings, a malicious website could exploit DNS rebinding to bypass same-origin policy restrictions and send requests to the local MCP server. This could allow an attacker to invoke tools or access resources exposed by the MCP server on behalf of the user in those limited circumstances. Note that running HTTP-based MCP servers locally without authentication is not recommended per MCP security best practices. This issue does not affect servers using stdio transport. This vulnerability is fixed in 1.23.0.
cvss_score: 7.6
status: fixed in 1.23.0

---

When this change is merged and adopted in enterprise, we can remove the override there for mcp.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2647,7 +2647,7 @@ wheels = [
 
 [[package]]
 name = "mcp"
-version = "1.10.1"
+version = "1.24.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "anyio" },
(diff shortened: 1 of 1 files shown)


uv.lock
```diff
@@ -2647,7 +2647,7 @@ wheels = [
 
 [[package]]
 name = "mcp"
-version = "1.10.1"
+version = "1.24.0"
```
lgh-solace, line 2650: My guess: It seems the uv lock file hadn't been updated with an update to google-adk so transitive dependencies were being installed locally that were not reflected in this lock file.

PR #652, opened 2025-12-15
DATAGO-119683: Add prefix support to SAM for MCP tools, to enable users to reduce collisions

m9p909:
## Summary
Add mappings to allow tool_prefixes with the MCP a2a toolset
## Purpose
Allow SAM enterprise to reduce collisions by adding prefixes to tools. 
Allow users to minimize collisions by adding prefixes to tools
## Testing Done
- Unit tests
- Manual agent tests

## Reviewer Questions
- Is this testing sufficient? Should I add an integration test? 
- Should I add an example? 
- Is the tool prefix name acceptable? 

## Other Info
- The OpenAPI version of this requires an update to a2a 1.21, we may need it in the near future. 

Screenshot 2025-12-15 at 3 45 32 PM


tests/unit/agent/adk/test_mcp_tool_loading.py
```diff
@@ -0,0 +1,57 @@
...
+    component = Mock()
+    component.log_identifier = "[TestAgent]"
+    return component
+
+
+class TestMCPToolLoadingWithPrefix:
+    """Test _load_mcp_tool with tool_name_prefix parameter."""
+
+    @pytest.mark.asyncio
+    async def test_load_mcp_tool_with_prefix(self, mock_component):
```
efunneko: I don't love these tests - there is a pile of stuff just to say that, yes, the value passed into the function was actually passed in. Is it possible to actually test that setting the prefix actually causes the tool name to be changed?

tests/unit/agent/tools/test_mcp_config_types.py
```diff
@@ -0,0 +1,46 @@
+"""Tests for MCP tool configuration types, specifically tool_name_prefix."""
+import pytest
+from pydantic import ValidationError
+from solace_agent_mesh.agent.tools.tool_config_types import McpToolConfig
+
+
+class TestMcpToolConfigPrefix:
+    """Test McpToolConfig model validation for tool_name_prefix feature."""
+
+    def test_mcp_config_prefix_valid_cases(self):
```
efunneko: These tests too feel very weak. If we feel the need to check that pydantic can validate these configs, I think they should be part of a test that checks all of the fields and not just a single one. 

Better would be that we have a test (as mentioned above) that actually checks that adding a prefix make a behaviour change on the tool name. If we did that, then these tests would be very unnecessary because they would be exercised as part of doing the functional test

src/solace_agent_mesh/agent/adk/setup.py
```diff
@@ -893,19 +895,7 @@ async def load_adk_tools(
...
-                                    mcp_tool.name, "mcp", loaded_tool_names
-                                )
-                        except Exception as e:
-                            log.error(
-                                "%s Failed to discover tools from MCP server for name registration: %s",
-                                component.log_identifier,
-                                str(e),
-                            )
-                            raise
+                        await _check_and_register_tool_name_mcp(component, loaded_tool_names, tool)
```
m9p909: Found a bug in this PR while testing

Hugo-Pare (approved):

m9p909:
I don't know why sonarqube is saying 0% coverage. There is plenty of coverage.  I don't think it's related to these changes because other pull requests have the same coverage score.

efunneko (approved):
That is better - thanks!

m9p909:
Forgot to do Conventional commits, sorry!

PR #656, opened 2025-12-16
fix(DATAGO-119912) unable to directly communicate with an Proxy Agent

enavitan:
This is a fix for an edge case when remote a2a agent abrupts execution, this was discovered when session was not matching expected UUID - format - as result when interacting with A2A agent directly (not via Orchestrator) call would hang forever.

to reproduce issue, on fresh master:
 - 1/ start SAM using debugger (chose SAM)
 - 2/ spin remote a2a agent from examples: https://github.com/a2aproject/a2a-samples/tree/main/samples/python/agents/langgraph
 - 3/ alter in above a2a agent:
 ```
+++ b/samples/python/agents/langgraph/app/agent_executor.py
@@ -85,6 +85,13 @@ class CurrencyAgentExecutor(AgentExecutor):
 
         query = context.get_user_input()
 
+        # Check if query contains the value 5 for currency conversion
+        import re
+        if re.search(r'\b5\b', query):
+            logger.error(f"[VALIDATION] Query contains value 5, silently aborting execution: {query}")
+            # Silently abort - just return without doing anything
+            return
```
- 4/ prompt  Currency Agent directly "convert CAD 5 to USD"
result: - call hangs forever.

apply current patch, repeat steps above confirm u get error immediately.

run unit tests:
make test-setup && make test-all

### CODE REVIEW

File: .vscode/launch.json [hunk 1]
```diff
@@ -11,7 +11,7 @@
             "module": "solace_ai_connector.main",
             "python": "${workspaceFolder}/.venv/bin/python",
             "console": "integratedTerminal",
-            "args": "--envfile .env examples/agents/orchestrator_example.yaml examples/agents/a2a_agents_example.yaml examples/gateways/webui_gateway_example.yaml",
+            "args": "--envfile .env examples/agents/orchestrator_example.yaml examples/a2a_currency.yaml examples/gateways/webui_gateway_example.yaml",
```
enavitan: temporary, for review

examples/a2a_currency.yaml
```diff
@@ -0,0 +1,47 @@
+# Agent Configuration File: A2A ADK Agents Configurations
+#
+# This file has the configuration for the A2A ADK agents.
+log:
```
enavitan: temporary for review

PR #667, opened 2025-12-16
fix(DATAGO-120075): Add fix for MCP spec to tool call conversion to fix LLM parsing errors

gregmeldrum:
### Summary

  Fixed a type serialization bug in lite_llm.py that prevented MCP tools from being called correctly. The issue occurred when converting Google ADK's types.Type enum
  values to the format expected by LiteLLM/OpenAI.

  ### Changes

  - Recursively convert Type enum values in nested schema properties and items
  - Handle schemas with no parameters by providing default empty object schema (required by OpenAI spec)
  - Convert entire parameters schema instead of manually reconstructing it

  ### Root Cause

  The original implementation only converted top-level type fields, missing nested Type enums in properties and items. When MCP tools had complex parameter schemas with
  nested objects or arrays, the unconverted enum values caused serialization errors during tool invocation.

###   Example

  Before: A schema like {"type": Type.OBJECT, "properties": {"items": {"type": Type.ARRAY}}} would fail because the nested Type.ARRAY enum wasn't converted to the string
  "array".

  After: All Type enums are recursively converted to their lowercase string values, ensuring proper serialization for LiteLLM/OpenAI.

  ### Testing

  Added comprehensive unit tests in tests/unit/agent/adk/models/test_lite_llm_schema_serialization.py

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -295,3 +295,42 @@ jobs:
                           }
                       }
               }"""
+      - name: Trigger RC Workflow in rc-sam-community
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+        with:
+          github-token: ${{ secrets.RC_TOKEN }}
+          script: |
+            const owner = 'SolaceDev';
+            const repo = 'rc-sam-community';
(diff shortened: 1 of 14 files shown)


efunneko (approved):
Looks good!

PR #672, opened 2025-12-17
fix(DATAGO-120073): Introduce allow_list / deny_list for MCP tool filtering

gregmeldrum:
Add allow_list and deny_list options to McpToolConfig for filtering
   which tools are loaded from an MCP server. This matches the pattern
   already established in OpenApiToolConfig.

   - allow_list: Include only specified tools (passed as list to ADK)
   - deny_list: Exclude specified tools (creates ToolPredicate)
   - tool_name, allow_list, and deny_list are mutually exclusive
   - Backward compatible: existing tool_name field still works

   Co-Authored-By: Claude <noreply@anthropic.com>"

### DIFF
--- a/tests/unit/agent/adk/test_mcp_tool_filtering.py
+++ b/tests/unit/agent/adk/test_mcp_tool_filtering.py
@@ -1,9 +1,10 @@
 """Tests for MCP tool filtering in McpToolConfig and setup.py."""
 import pytest
-from unittest.mock import Mock, AsyncMock, patch
+from unittest.mock import Mock, AsyncMock, patch, MagicMock
 from pydantic import ValidationError
 
 from solace_agent_mesh.agent.tools.tool_config_types import McpToolConfig
+from solace_agent_mesh.agent.adk.setup import _load_mcp_tool
 
 
 class TestMcpToolConfigFiltering:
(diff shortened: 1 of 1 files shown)


src/solace_agent_mesh/agent/adk/setup.py
```diff
@@ -625,14 +625,30 @@ async def _load_mcp_tool(component: "SamAgentComponent", tool_config: Dict) -> T
...
+    filter_description = "none (all tools)"
+
+    if tool_config_model.tool_name:
+        # Backward compatible: single tool name becomes a list
+        tool_filter = [tool_config_model.tool_name]
+        filter_description = f"tool_name='{tool_config_model.tool_name}'"
+    elif tool_config_model.allow_list:
+        # Allow list: pass directly as list of tool names
+        tool_filter = tool_config_model.allow_list
+        filter_description = f"allow_list={tool_config_model.allow_list}"
```
efunneko, line 640: Q: What does this serialize to? Is it a json array?

gregmeldrum: It looks like this:

```
MCP tool config specifies filter: allow_list=['addCommentToJiraIssue']
```

PR #678, opened 2025-12-18
fix(DATAGO-118675): Start the user in AI mode when editing a prompt

solace-pjones:
Problem
When users edit existing prompts in the Prompt Template Builder, the interface was defaulting to Manual mode. This meant users had to manually click "Build with AI" to access AI assistance features, adding an unnecessary step to the editing workflow.

Solution
Changed the default builder mode to always start in AI-assisted mode, regardless of whether the user is creating a new prompt or editing an existing one. 

Testing
Added new Storybook test case EditingModeDefaultsToAIAssisted to verify:
When editing a prompt, AI-assisted mode is active by default
"Edit Manually" button is visible (confirming AI mode is active)
"Build with AI" button is not present (confirming we're already in AI mode)

### DIFF
--- a/client/webui/frontend/src/lib/components/pages/PromptsPage.tsx
+++ b/client/webui/frontend/src/lib/components/pages/PromptsPage.tsx
@@ -60,7 +60,7 @@ export const PromptsPage: React.FC = () => {
                     try {
                         const data = await api.webui.get(`/api/v1/prompts/groups/${loaderData.promptId}`);
                         setEditingGroup(data);
-                        setBuilderInitialMode("manual");
+                        setBuilderInitialMode("ai-assisted"); // Always start in AI-assisted mode for editing
                         setShowBuilder(true);
                     } catch (error) {
                         displayError({ title: "Failed to Edit Prompt", error: getErrorMessage(error, "An error occurred while fetching prompt.") });


client/webui/frontend/src/lib/components/prompts/PromptTemplateBuilder.tsx
```diff
@@ -23,7 +23,7 @@ interface PromptTemplateBuilderProps {
 export const PromptTemplateBuilder: React.FC<PromptTemplateBuilderProps> = ({ onBack, onSuccess, initialMessage, editingGroup, isEditing = false, initialMode }) => {
     const { config, updateConfig, saveTemplate, updateTemplate, resetConfig, validationErrors, isLoading } = usePromptTemplateBuilder(editingGroup);
 
-    const [builderMode, setBuilderMode] = useState<"manual" | "ai-assisted">(initialMode || (isEditing ? "manual" : "ai-assisted"));
+    const [builderMode, setBuilderMode] = useState<"manual" | "ai-assisted">(initialMode || "ai-assisted");
```
lgh-solace, line 26: I'm curious how this fallback interacts with the incoming prop? If a user starts in the prompts page and clicks edit from here... should we also land in the ai-assisted mode? Could you confirm? Thanks!

image

solace-pjones: I missed a file in the original change. PromptsPage.tsx was already setting builderInitialMode to "manual" for editing scenarios. While we could remove that line now that the component defaults to "ai-assisted", I've decided to keep it for clarity and defensive programming. It makes the parent component's intent explicit and ensures the behavior is clear.

lgh-solace (approved):
Thanks!

PR #707, opened 2025-12-21
feat(DATAGO-114389): PR 2: Workflow Definition Models

efunneko:
# PR 2: Workflow Definition Models

## Overview

This PR introduces the Pydantic models that define the YAML schema for workflow definitions. These models provide Argo Workflows-compatible syntax with SAM extensions, enabling users to define deterministic, structured workflow DAGs.

## Branch Information

- **Branch Name:** `pr/workflows-2-models`
- **Target:** `pr/workflows-1-foundation`

### `src/solace_agent_mesh/workflow/app.py`

Complete workflow definition model hierarchy:

#### Retry & Exit Handlers

| Model | Purpose |
|-------|---------|
| `BackoffStrategy` | Exponential backoff configuration for retries |
| `RetryStrategy` | Retry configuration (limit, policy, backoff) |
| `ExitHandler` | Cleanup/notification on workflow completion |

#### Node Types

| Model | Purpose |
|-------|---------|
| `WorkflowNode` | Base node class with id, type, depends_on |
| `AgentNode` | Invoke an agent with input/output schemas |
| `ConditionalNode` | Binary if/else branching |
| `SwitchNode` | Multi-way branching (first match wins) |
| `SwitchCase` | Individual case in switch node |
| `LoopNode` | While-loop iteration until condition false |
| `MapNode` | Parallel for-each iteration over items |

#### Workflow Definition & Configuration

| Model | Purpose |
|-------|---------|
| `WorkflowDefinition` | Complete workflow DAG with nodes, schemas, exit handlers |
| `WorkflowAppConfig` | App configuration extending SamAgentAppConfig |
| `WorkflowApp` | Custom App class for workflow orchestration |

### `src/solace_agent_mesh/workflow/__init__.py`

Package initialization and exports.

### Argo Workflows Compatibility

The models support Argo-style field names via aliases:

```yaml
# Both syntaxes work:
depends_on: [step1]      # SAM style
dependencies: [step1]     # Argo style

true_branch: step2        # SAM style
trueBranch: step2         # Argo style
```

### Node Type Summary

| Type | Use Case | Key Fields |
|------|----------|------------|
| `agent` | Call an agent | `agent_name`, `input`, `when`, `retry_strategy` |
| `conditional` | Binary branch | `condition`, `true_branch`, `false_branch` |
| `switch` | Multi-way branch | `cases[]`, `default` |
| `loop` | While iteration | `node`, `condition`, `max_iterations` |
| `map` | For-each parallel | `items`, `node`, `concurrency_limit` |

### DAG Validation

`WorkflowDefinition.validate_dag_structure()` ensures:
- All node references are valid
- Branch targets depend on their parent node
- Exit handler references exist
- No configuration contradictions

### Template Expression Syntax

Input mappings use template expressions:

```yaml
input:
  query: "{{workflow.input.search_term}}"
  data: "{{previous_node.output.results}}"
```

### DIFF
--- a/src/solace_agent_mesh/workflow/app.py
+++ b/src/solace_agent_mesh/workflow/app.py
@@ -1,0 +1,622 @@
+"""
+WorkflowApp class and configuration models for Prescriptive Workflows.
+
+Supports Argo Workflows-compatible syntax with SAM extensions.
+"""
+
+import logging
+from typing import Any, Dict, List, Literal, Optional, Union
+from pydantic import BaseModel, Field, model_validator
+
+from solace_ai_connector.flow.app import App
+from ..common import a2a
+from ..agent.sac.app import SamAgentAppConfig, AgentCardConfig, AgentCardPublishingConfig
+
+log = logging.getLogger(__name__)
+
+# --- Retry Strategy Models (Argo-compatible) ---
+
+
+class BackoffStrategy(BaseModel):
+    """Exponential backoff configuration for retries."""
+
+    duration: str = Field(
+        default="1s",
+        description="Initial backoff duration. Supports: '5s', '1m', '1h'.",
+    )
+    factor: float = Field(
+        default=2.0,
+        description="Multiplier for exponential backoff.",
+    )
+    max_duration: Optional[str] = Field(
+        default=None,
+        description="Maximum backoff duration cap.",
+        alias="maxDuration",
+    )
+
+    class Config:
+        populate_by_name = True
+
+
+class RetryStrategy(BaseModel):
+    """
+    Retry configuration for workflow nodes.
+    Argo-compatible with extensions.
+    """
+
+    limit: int = Field(
+        default=3,
+        description="Maximum number of retry attempts.",
+    )
+    retry_policy: Literal["Always", "OnFailure", "OnError"] = Field(
+        default="OnFailure",
+        description="When to retry: Always, OnFailure, OnError.",
+        alias="retryPolicy",
+    )
+    backoff: Optional[BackoffStrategy] = Field(
+        default=None,
+        description="Exponential backoff configuration.",
+    )
+
+    class Config:
+        populate_by_name = True
+
+
+# --- Exit Handler Model ---
+
+
+class ExitHandler(BaseModel):
+    """
+    Exit handler configuration for cleanup/notification on workflow completion.
+
+    Supports conditional handlers for different outcomes.
+    """
+
+    always: Optional[str] = Field(
+        default=None,
+        description="Node ID to execute regardless of workflow outcome.",
+    )
+    on_success: Optional[str] = Field(
+        default=None,
+        description="Node ID to execute only on successful completion.",
+        alias="onSuccess",
+    )
+    on_failure: Optional[str] = Field(
+        default=None,
+        description="Node ID to execute only on failure.",
+        alias="onFailure",
+    )
+
+    class Config:
+        populate_by_name = True
+
+
+# --- Workflow Node Models ---
+
+
+class WorkflowNode(BaseModel):
+    """
+    Base workflow node.
+
+    Supports both SAM and Argo field names:
+    - depends_on / dependencies (Argo alias)
+    """
+
+    id: str = Field(..., description="Unique node identifier")
+    type: str = Field(..., description="Node type")
+    depends_on: Optional[List[str]] = Field(
+        default=None,
+        description="List of node IDs this node depends on.",
+        alias="dependencies",
+    )
+
+    class Config:
+        populate_by_name = True
+
+
+class AgentNode(WorkflowNode):
+    """
+    Agent invocation node.
+
+    Argo-aligned features:
+    - `when`: Conditional execution clause (Argo-style)
+    - `retryStrategy`: Retry configuration
+    - `timeout`: Node-specific timeout override
+    """
+
+    type: Literal["agent"] = "agent"
+    agent_name: str = Field(..., description="Name of agent to invoke")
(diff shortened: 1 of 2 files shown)


src/solace_agent_mesh/workflow/app.py
```diff
@@ -0,0 +1,622 @@
...
+
+class ConditionalNode(WorkflowNode):
+    """
+    Conditional branching node (binary true/false).
+
+    For multi-way branching, use SwitchNode instead.
+    """
+
+    type: Literal["conditional"] = "conditional"
+    condition: str = Field(..., description="Expression to evaluate")
```
alimosaed, line 168: I assume that these expressions are validated in the next PRs. It was better to validate them in their own classes.

src/solace_agent_mesh/workflow/app.py
```diff
@@ -0,0 +1,622 @@
...
+    LoopNode repeats a node until a condition becomes false.
+    """
+
+    type: Literal["loop"] = "loop"
+    node: str = Field(..., description="Node ID to execute repeatedly")
+    condition: str = Field(
+        ...,
+        description="Continue looping while this expression is true.",
+    )
+    max_iterations: int = Field(
```
alimosaed, line 235: Should not this filed be optional? When the condition is data driven, this cap reduces the flexibility of the condition.

src/solace_agent_mesh/workflow/app.py
```diff
@@ -0,0 +1,622 @@
...
+    )
+
+    class Config:
+        populate_by_name = True
+
+
+# --- Exit Handler Model ---
+
+
+class ExitHandler(BaseModel):
```
alimosaed, line 68: Do not you need to handle cancellation?

src/solace_agent_mesh/workflow/app.py
```diff
@@ -0,0 +1,622 @@
...
+            if isinstance(self.on_exit, str):
+                if self.on_exit not in node_map:
+                    raise ValueError(
+                        f"onExit references non-existent node '{self.on_exit}'"
+                    )
+            else:
+                for field, node_id in [
+                    ("always", self.on_exit.always),
+                    ("on_success", self.on_exit.on_success),
+                    ("on_failure", self.on_exit.on_failure),
```
alimosaed, line 456: If you add on_cancel, please update here as well

src/solace_agent_mesh/workflow/app.py
```diff
@@ -0,0 +1,622 @@
...
+        default=None,
+        description="Default retry strategy for all nodes (can be overridden per-node).",
+        alias="retryStrategy",
+    )
+
+    class Config:
+        populate_by_name = True
+
+    @model_validator(mode="after")
+    def validate_dag_structure(self) -> "WorkflowDefinition":
```
alimosaed, line 397: The method does not validate if there is only and only one initial node.

src/solace_agent_mesh/workflow/app.py
efunneko: Expression validation happens at runtime in the DAG executor. Static validation of expression syntax in Pydantic models would be complex and couldn't catch all runtime issues (e.g., references to nodes that don't exist in context). The current approach validates structural correctness at parse time and expression correctness at execution time. Tracked in DATAGO-121942 - no changes needed.

src/solace_agent_mesh/workflow/app.py
efunneko: The `max_iterations` field has `default=100`, so it's effectively optional - users don't need to specify it. The safety cap is intentional to prevent runaway loops. Users can set a higher value (e.g., `max_iterations: 10000`) if their use case requires more iterations. Tracked in DATAGO-121942 - no changes needed.

src/solace_agent_mesh/workflow/app.py
efunneko: Good suggestion for a future enhancement. For the MVP, we handle `on_success` and `on_failure` which covers the main use cases. Cancellation handling (`on_cancel`) could be added later when we implement more sophisticated workflow lifecycle management. Tracked in DATAGO-121942 - deferred to future enhancement.

src/solace_agent_mesh/workflow/app.py
efunneko: Acknowledged - if `on_cancel` is added to `ExitHandler` in the future, the validation logic will need to be updated accordingly. Tracked in DATAGO-121942.

src/solace_agent_mesh/workflow/app.py
efunneko: Multiple initial nodes (nodes without dependencies) is a valid and useful pattern for parallel execution - the workflow can start multiple branches simultaneously. Having zero initial nodes would result in nothing executing, which is caught at runtime. Adding validation for exactly one initial node would be overly restrictive. Tracked in DATAGO-121942 - no changes needed.

src/solace_agent_mesh/workflow/app.py
efunneko: Update: After further review, we will implement `on_cancel` as part of this PR stack. The fix is tracked in DATAGO-121943 and will be applied in a new PR at the top of the stack.

src/solace_agent_mesh/workflow/app.py
efunneko: Will do - updating validation logic is included in the fix. Tracked in DATAGO-121943.

efunneko:
Closing as part of stacked PR cleanup. This PR has been reviewed and approved. Any follow-up items have been tracked in Jira.

PR #708, opened 2025-12-21
feat(DATAGO-114389): PR 3: Structured Invocation Support

efunneko:
# PR 3: Structured Invocation Support

## Overview

This PR enables agents to be invoked with schema-validated input/output, functioning as a "structured function call" pattern. This is used by workflows and other programmatic callers that need predictable, validated responses from agents.

## Branch Information

- **Branch Name:** `pr/workflows-3-agent-support`
- **Target:** `pr/workflows-2-models`

### `src/solace_agent_mesh/agent/sac/structured_invocation/handler.py`

The `StructuredInvocationHandler` class (~1,100 lines) provides:

| Method | Purpose |
|--------|---------|
| `extract_structured_invocation_context()` | Detect if message is a structured invocation |
| `execute_structured_invocation()` | Main entry point for structured execution |
| `_extract_input_data()` | Extract input from message parts |
| `_execute_with_output_validation()` | Run agent and validate output |
| `_parse_result_embed()` | Extract result artifact from agent output |
| `_return_structured_result()` | Send StructuredInvocationResult response |

#### Key Features

- **Input extraction**: Handles text input, JSON/YAML/CSV file parts, and artifact URIs
- **Schema validation**: Validates input before execution, output after
- **Output retry**: Re-prompts agent if output fails validation (configurable retries)
- **Result embed pattern**: Agents signal completion via `«result:artifact=output.json»` embed
- **Error handling**: Catches exceptions and returns structured failure result

### `src/solace_agent_mesh/agent/sac/structured_invocation/validator.py`

Simple JSON Schema validation using `jsonschema` library:

```python
def validate_against_schema(data: Any, schema: Dict[str, Any]) -> Optional[List[str]]:
    """Validate data against JSON schema. Returns error list or None."""
```

### `src/solace_agent_mesh/agent/sac/component.py` (modifications)

Integration of structured invocation into the agent component:

- Check incoming messages for `StructuredInvocationRequest` data part
- Route to handler if detected
- Fall back to normal agent processing otherwise

### `src/solace_agent_mesh/agent/sac/app.py` (modifications)

New configuration fields:

- `input_schema`: Default input schema for agent
- `output_schema`: Default output schema for agent
- `validation_max_retries`: Max retries on output validation failure

### Structured Invocation Flow

```
1. Caller sends A2A message with StructuredInvocationRequest data part
2. Agent extracts invocation context
3. Agent validates input against schema
4. Agent executes with structured prompt
5. Agent validates output against schema
6. If output invalid, retry with correction prompt (up to max_retries)
7. Agent returns StructuredInvocationResult with artifact reference
```

### Result Embed Pattern

Agents signal their structured output via a special embed format:

```
«result:artifact=output.json status=success»
```

The handler parses this to locate the output artifact for validation.

### Schema Override Hierarchy

1. Schema in `StructuredInvocationRequest` (highest priority)
2. Schema in agent card extensions
3. Schema in agent configuration
4. Default text schema (fallback)

### DIFF
--- a/src/solace_agent_mesh/agent/sac/structured_invocation/handler.py
+++ b/src/solace_agent_mesh/agent/sac/structured_invocation/handler.py
@@ -1,0 +1,1161 @@
+"""
+StructuredInvocationHandler implementation.
+
+Enables agents to be invoked with schema-validated input/output,
+functioning as a "structured function call" pattern. Used by workflows
+and other programmatic callers that require predictable, validated responses.
+"""
+
+import logging
+import json
+import asyncio
+import re
+import yaml
+import csv
+import io
+from datetime import datetime, timezone
+from typing import Any, Callable, Dict, List, Optional, TYPE_CHECKING
+
+from pydantic import ValidationError
+from google.adk.models.llm_request import LlmRequest
+from google.adk.models.llm_response import LlmResponse
+from google.adk.agents.callback_context import CallbackContext
+from google.adk.events import Event as ADKEvent
+from google.genai import types as adk_types
+from google.adk.agents import RunConfig
+from google.adk.agents.run_config import StreamingMode
+
+from a2a.types import (
+    Message as A2AMessage,
+    FilePart,
+    FileWithBytes,
+    FileWithUri,
+    TaskState,
+)
+
+from ....common import a2a
+from ....common.data_parts import (
+    StructuredInvocationRequest,
+    StructuredInvocationResult,
+)
+from ....agent.adk.runner import run_adk_async_task_thread_wrapper
+from ....common.utils.embeds.constants import EMBED_REGEX
+from ....agent.utils.artifact_helpers import parse_artifact_uri
+
+if TYPE_CHECKING:
+    from ..component import SamAgentComponent
+
+log = logging.getLogger(__name__)
+
+
+class ResultEmbed:
+    """Parsed result embed from agent output."""
+
+    def __init__(
+        self,
+        artifact_name: Optional[str] = None,
+        version: Optional[int] = None,
+        status: str = "success",
+        message: Optional[str] = None,
+    ):
+        self.artifact_name = artifact_name
+        self.version = version
+        self.status = status
+        self.message = message
+
+
+class StructuredInvocationHandler:
+    """
+    Handles structured invocation logic for an agent.
+
+    Enables agents to be invoked with schema-validated input and output,
+    supporting retry on validation failure. Used by workflows and other
+    programmatic callers that need predictable, validated responses.
+    """
+
+    def __init__(self, host_component: "SamAgentComponent"):
+        self.host = host_component
+        self.input_schema = host_component.get_config("input_schema")
+        self.output_schema = host_component.get_config("output_schema")
+        self.max_validation_retries = host_component.get_config(
+            "validation_max_retries", 2
+        )
+
+    def extract_structured_invocation_context(
+        self, message: A2AMessage
+    ) -> Optional[StructuredInvocationRequest]:
+        """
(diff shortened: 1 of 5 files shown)


src/solace_agent_mesh/agent/sac/structured_invocation/handler.py
```diff
@@ -0,0 +1,1161 @@
...
+        # This prevents parsing example embeds from the workflow instructions
+        if adk_event.content.role != "model":
+            log.debug(f"Result embed parse: Event role is {adk_event.content.role}, skipping.")
+            return None
+
+        # Extract text from last event
+        text_content = ""
+        for part in adk_event.content.parts:
+            if part.text:
+                text_content += part.text
```
alimosaed, line 846: You can add a splitter like \n within texts.

src/solace_agent_mesh/agent/sac/structured_invocation/handler.py
```diff
@@ -0,0 +1,1161 @@
...
+                if version_str.startswith("v"):
+                    version_str = version_str[1:]
+                version = int(version_str)
+            except (ValueError, TypeError):
+                pass
+
+        return ResultEmbed(
+            artifact_name=artifact_name,
+            version=version,
+            status=params.get("status", "success"),
```
alimosaed, line 913: if the param_pattern Regex cannot extract 'status' param, this line set the default value 'success', which seems incorrect. The method does not check if an artifact result is in "«result:artifact=<name>:v<version> status=<success|failure> message="<text>"»" format

src/solace_agent_mesh/agent/sac/structured_invocation/handler.py
efunneko: The text parts are from a single message content and are concatenated to form a searchable string for embed pattern matching. Adding newlines between parts could break the regex matching if an embed spans multiple parts. The current implementation is intentional. Tracked in DATAGO-121944 - no changes needed.

src/solace_agent_mesh/agent/sac/structured_invocation/handler.py
efunneko: The default to `status="success"` is intentional design. If an agent outputs a result embed with an artifact reference, it implicitly means the agent successfully produced output. The `status` field exists for agents to *explicitly* report failure (e.g., `status="error"`). This follows a "success unless explicitly marked as failure" pattern which is common in similar protocols. The format comment in line 862 documents the expected format; malformed embeds would fail earlier in parsing (no artifact match). Tracked in DATAGO-121944 - no changes needed.

src/solace_agent_mesh/agent/sac/structured_invocation/handler.py
efunneko: The text parts are from a single message content and are concatenated to form a searchable string for embed pattern matching. Adding newlines between parts could break the regex matching if an embed somehow spans multiple parts. The current implementation is intentional. Tracked in DATAGO-121944 - no changes needed.

src/solace_agent_mesh/agent/sac/structured_invocation/handler.py
efunneko: Good catch. You are right that defaulting to `status="success"` for malformed embeds is sloppy - it would just fail later at artifact load time with a confusing error. We will add validation to ensure either `artifact` is present OR `status=error` is explicitly set. If neither, return `None` to trigger proper error handling with a clear message. Tracked in DATAGO-121944.

efunneko:
Closing as part of stacked PR cleanup. This PR has been reviewed and approved. Any follow-up items have been tracked in Jira.

PR #731, opened 2025-12-31
fix(DATAGO-120524) Prompt Templates "Discard Changes" button should not ask for confirmation

amir-ghasemi:
This pull request makes a minor adjustment to the `PromptTemplateBuilder` component to improve the handling of the close action when editing a prompt template.

* The `handleClose` function is now called with the `isEditing` flag as an argument, allowing it to distinguish between discarding changes and simply canceling.


client/webui/frontend/src/lib/components/prompts/PromptTemplateBuilder.tsx
```diff
@@ -339,7 +339,7 @@ export const PromptTemplateBuilder: React.FC<PromptTemplateBuilderProps> = ({ on
                 <NavigationBlocker />
                 {/* Footer Actions */}
                 <div className="flex justify-end gap-2 border-t p-4">
-                    <Button variant="ghost" onClick={() => handleClose()} disabled={isLoading}>
+                    <Button variant="ghost" onClick={() => handleClose(isEditing)} disabled={isLoading}>
```
lgh-solace, line 342: We can deliver this change, but I think there might be some clean up to do around navigation - it seems like there are flows where we call `allowNavigation` twice.  We can go over it in the next iteration. :)

PR #736, opened 2026-01-05
feat(DATAGO-118201): Add React Query

JKaram:

    Consolidating api call usage to one library, removing the need for repeated logic, and having a powerful caching library.

- Install React Query
- Wrap app in Query provider
- Export `queryClient` for use in enterprise.
-  Write initial api service/keys/hooks layer as an example of the new pattern

   The api calls that were created from projects are NOT being used in the app. The `queryClient` must be consumed by enterprise before using `react-query`.

### DIFF
--- a/client/webui/frontend/src/lib/providers/index.ts
+++ b/client/webui/frontend/src/lib/providers/index.ts
@@ -7,3 +7,4 @@ export * from "./CsrfProvider";
 export * from "./ThemeProvider";
 export * from "./ProjectProvider";
 export * from "./QueryProvider";
+export * from "./QueryClient";
--- a/client/webui/frontend/src/lib/providers/QueryClient.ts
+++ b/client/webui/frontend/src/lib/providers/QueryClient.ts
@@ -1,0 +1,13 @@
+import { QueryClient } from "@tanstack/react-query";
+
+/**
+ * Shared QueryClient instance used across the application.
+ * Export this to allow manual query invalidation or cache manipulation.
+ */
+export const queryClient = new QueryClient({
+    defaultOptions: {
+        queries: {
+            staleTime: 1000 * 60 * 5,
+        },
+    },
+});
--- a/client/webui/frontend/src/lib/providers/QueryProvider.tsx
+++ b/client/webui/frontend/src/lib/providers/QueryProvider.tsx
@@ -1,17 +1,6 @@
 import React from "react";
-import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
-
-/**
- * Shared QueryClient instance used across the application.
- * Export this to allow manual query invalidation or cache manipulation.
- */
-export const queryClient = new QueryClient({
-    defaultOptions: {
-        queries: {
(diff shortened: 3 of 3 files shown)


client/webui/frontend/src/lib/api/projects/keys.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+export const projectKeys = {
+    all: ["projects"] as const,
+    lists: () => [...projectKeys.all, "list"] as const,
+    list: (filters?: Record<string, unknown>) => [...projectKeys.lists(), { filters }] as const,
+    details: () => [...projectKeys.all, "detail"] as const,
+    detail: (id: string) => [...projectKeys.details(), id] as const,
+    artifacts: (id: string) => [...projectKeys.detail(id), "artifacts"] as const,
+    sessions: (id: string) => [...projectKeys.detail(id), "sessions"] as const,
+};
```
JKaram, line 13: Initial keying strategy. Forgoing the use of `https://github.com/lukemorales/query-key-factory` library. 

I am open to alternatives.

client/webui/frontend/src/lib/providers/index.ts
```diff
@@ -6,3 +6,4 @@ export * from "./AuthProvider";
 export * from "./CsrfProvider";
 export * from "./ThemeProvider";
 export * from "./ProjectProvider";
+export * from "../queryClient";
```
JKaram: Exporting the `queryClient` from `community`, to be consumed by `enterprise`.

Enterprise will have `react-query` installed and will share the same `queryClient` (otherwise it will not work)

client/webui/frontend/package.json
```diff
@@ -79,6 +79,7 @@
         "@radix-ui/react-tabs": "^1.1.12",
         "@radix-ui/react-tooltip": "^1.2.7",
         "@tailwindcss/vite": "^4.1.10",
+        "@tanstack/react-query": "5.90.16",
```
JKaram, line 83: Locking in the version to be the same as in `enterprise`

client/webui/frontend/src/lib/providers/index.ts
lgh-solace: Optionally, I'm wondering if we could have our own `QueryProvider` that wraps the tanstack one - it's only responsibility would be to instantiate the query client and provide it via its context. Thus, the current pattern might be maintainable and we could still insure only one queryClient is created?

JKaram: I also thought this. My only concern is exporting too much of React Query through our own library. I will do a test to see how much this effects the bundle of `solace-agent-mesh-ui-library-whatever-its-called`

JKaram: Ok all good the size difference is negligible.

client/webui/frontend/src/lib/api/projects/keys.ts
lgh-solace: Ok, I think we can iterate on these, looks good to start!

PR #739, opened 2026-01-06
fix(DATAGO-120206): Get agent card from remote agent

alimosaed:

1. **First:** A SAM application cannot retrieve an agent card of a remote agent that is connected via an A2A proxy from a customized path. It always uses the default agent card path, while the agent card path on the remote agent may differ.  
2. **Second:** If the remote agent requires authentication to retrieve an agent card, a SAM application fails to authenticate.

- For the first problem, the agent card path has been appended to the agent URL. This is accomplished by passing `agent_card_path` to the `A2ACardResolver` method. This method already supported a customized agent card path; however, the correct value was not being passed to it.

- For the second problem, synchronous authentication was added to the base class of the A2A proxy. This authentication is necessary to retrieve the agent card immediately when the proxy starts.

### Key Design Decisions

Both synchronous and asynchronous authentication methods are required because the proxy has two distinct discovery phases operating in different execution contexts. The base component’s `_perform_initial_discovery()` runs synchronously during `__init__()` to quickly populate the agent registry at startup using `httpx.Client()`, while the A2A component’s `_fetch_agent_card()` runs asynchronously on a recurring timer using `httpx.AsyncClient()` to refresh agent cards and support OAuth2 authentication (which requires awaiting token retrieval). Since async functions cannot be called from synchronous code without significant architectural changes, `build_static_auth_headers()` is used for the synchronous startup path (supporting only bearer/API key authentication), and `build_auth_headers_async()` is used for the asynchronous runtime path (supporting all authentication types, including OAuth2). Both methods share the same core authentication logic to avoid code duplication while respecting the fundamental synchronous/asynchronous boundary imposed by Python’s execution model.

Does the design make sense and if there is any edge cases.

### DIFF
--- a/src/solace_agent_mesh/common/auth_headers.py
+++ b/src/solace_agent_mesh/common/auth_headers.py
@@ -23,7 +23,7 @@ def build_static_auth_headers(
     - static_bearer: Bearer token authentication
     - static_apikey: API key authentication
 
-    For OAuth2 authentication, use build_auth_headers_async() instead.
+    For OAuth2 authentication, use build_full_auth_headers() instead.
 
     Args:
         agent_name: The name of the agent
@@ -42,7 +42,7 @@ def build_static_auth_headers(
     Note:
         OAuth2 authentication types (oauth2_client_credentials,
         oauth2_authorization_code) will be skipped with a warning.
-        These require async context - use build_auth_headers_async().
+        These require async context - use build_full_auth_headers().
 
     Example:
         >>> headers = build_static_auth_headers(
@@ -90,7 +90,7 @@ def build_static_auth_headers(
                     log.warning(
                         "%s OAuth2 authentication (%s) is not supported in synchronous context. "
                         "Agent '%s' headers will not include authentication. "
-                        "Use build_auth_headers_async() for OAuth2 support.",
+                        "Use build_full_auth_headers() for OAuth2 support.",
                         log_identifier,
                         auth_type,
                         agent_name,
@@ -108,7 +108,7 @@ def build_static_auth_headers(
     return headers
 
 
-async def build_auth_headers_async(
+async def build_full_auth_headers(
     agent_name: str,
     agent_config: Dict[str, Any],
     custom_headers_key: str,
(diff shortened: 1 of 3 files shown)


src/solace_agent_mesh/common/auth_headers.py
```diff
@@ -0,0 +1,214 @@
...
+        for header_config in custom_headers_list:
+            header_name = header_config.get("name")
+            header_value = header_config.get("value")
+            if header_name and header_value:
+                headers[header_name] = header_value
+
+    return headers
+
+
+async def build_auth_headers_async(
```
enavitan: this method implies resemblance to the one above except its async - however in reality:
 - uses the static one to  build static headers
 - performs async calls to the oauth_toke_fetcher
 - pulls custom headers from the config
 - aggregates the results of all of them
 
So in short name doesn't reflect what exactly method does

this does feel more like some sort of a builder pattern however caller must be in charge of deciding what to pass in for final aggregation


```
HeaderBuilder
  .with(static_headers) <-- call to the static headers function 
  .with(await fetch_oauth_headers(...)) <-- call to oauth2 Auth headers
  .with(custom_headers) <-- call to the custom headers method 
  .build()
```
so in short - aggregator knows how to assemble these headers, while consumer decides what kind of header he interested in, WDYT?

src/solace_agent_mesh/agent/proxies/base/component.py
```diff
@@ -420,8 +428,49 @@ def _initial_discovery_sync(self):
...
+
+                            if auth_type in ("oauth2_client_credentials", "oauth2_authorization_code"):
+                                # Skip this agent - OAuth2 not supported in sync context
+                                log.info(
+                                    "%s Skipping agent '%s' in synchronous discovery (OAuth2 configured). "
+                                    "Will be discovered during periodic async discovery.",
+                                    self.log_identifier,
+                                    agent_alias,
+                                )
+                                continue
```
enavitan, line 453: lets improve flow a little:
 - we build static headers but later on  we might not even fetch anything if oauth_type is of certain type- thus we continue.

how about instead:
check if auth_type is of the type we need and if o so go ahead with static headers calcualtion, otherwise continue - not spend time building static_headers?

WDYT?

src/solace_agent_mesh/common/auth_headers.py
alimosaed: Renamed the methods.

About the builder pattern, since we need only two scenarios of building a header, the simplest way is to have two separate methods. The builder pattern is a good option for scenarios when various scenarios can be generated using small methods.
So, definitely we can use the builder method but it does not add enough value. On the other hand, we have to test different authentication scenarios.

PR #743, opened 2026-01-06
fix(DATAGO-119380): Update Context7 URL

alimosaed:

- The context7 url was configured wrongly.
- An irrelevant unit test was failing.
`-___ TestOAuth2ClientCredentialsTokenManager.test_ssl_certificate_validation ____
/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/unittest/mock.py:929: in assert_called_with
    raise AssertionError(_error_message()) from cause
E   AssertionError: expected call not found.
E   Expected: AsyncClient(verify='/path/to/ca.crt')
E   Actual: AsyncClient(headers={})`

- Updated the url of SAM library in the `context7.json` file.
- Updated the unit test to ensure that AsyncClient is called once.

Not Applicable

Nothing

### DIFF
--- a/tests/unit/agent/adk/models/test_oauth2_token_manager.py
+++ b/tests/unit/agent/adk/models/test_oauth2_token_manager.py
@@ -234,7 +234,7 @@ class TestOAuth2ClientCredentialsTokenManager:
             await manager.get_token()
 
             # Verify AsyncClient was called with custom CA cert
-            mock_client.assert_called_with(verify="/path/to/ca.crt")
+            mock_client.assert_called_once_with(verify="/path/to/ca.crt")
 
     @pytest.mark.asyncio
     async def test_concurrent_token_requests(self, token_manager, mock_token_response):


tests/unit/agent/adk/models/test_oauth2_token_manager.py
```diff
@@ -234,7 +234,7 @@ async def test_ssl_certificate_validation(self, mock_token_response):
             await manager.get_token()
 
             # Verify AsyncClient was called with custom CA cert
-            mock_client.assert_called_with(verify="/path/to/ca.crt")
+            mock_client.assert_called_once_with(verify="/path/to/ca.crt")
```
enavitan, line 237: I wonder how exactly this fix fixes:
```
Expected: AsyncClient(verify='/path/to/ca.crt') E   Actual: AsyncClient(headers={})
```
we are now asserting called_once_with vs called_with

enavitan (approved):
url fix makes sense - as for unit test looks like - there are multiple invocations now?, and we're just asserting at least one of those was with expected path.

tests/unit/agent/adk/models/test_oauth2_token_manager.py
alimosaed: This unit test is passed locally; however, it failed in Github with the below link:
`-___ TestOAuth2ClientCredentialsTokenManager.test_ssl_certificate_validation ____ /opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/unittest/mock.py:929: in assert_called_with raise AssertionError(_error_message()) from cause E   AssertionError: expected call not found. E   Expected: AsyncClient(verify='/path/to/ca.crt') E   Actual: AsyncClient(headers={})`

The error shows that AsyncClient is called by a wrong parameter (`headers={}`), whereas the test requires `verify="/path/to/ca.crt"`). It seems that there is a mismatch between Github environment and local system (maybe the `pytest-mock` version) that invokes `AsyncClient` multiple times with different parameters. `mock_client.assert_called_with` always choose the last call, which is `AsyncClient(headers={})`; however, `assert_called_once_with(verify="/path/to/ca.crt")` checks that AsyncClient is called once with `verify="/path/to/ca.crt"` parameter.  
So, `assert_called_once_with` covers the unit test goal while ignoring the mismatch.

PR #744, opened 2026-01-06
feat(DATAGO-120289): Add stepper ui component

ziyanwan:

    Add a UI component based on https://stepperize.vercel.app/docs/react/shadcn/introduction

    1. copy example code from https://stepperize.vercel.app/docs/react/shadcn/installation
    2. fix type issues and export style
    3. fix step button style (use outline button variant for inactive step, use cursor default)

- [✅ ] Manual testing: [Tested with npm link in SAM enterprise]

### Demos

https://github.com/user-attachments/assets/efc22603-5efd-480c-871c-e3b19194e83c

Screenshot 2026-01-06 at 12 28 23 PM
Screenshot 2026-01-06 at 12 28 34 PM
Screenshot 2026-01-06 at 12 27 45 PM
Screenshot 2026-01-06 at 12 28 00 PM

### DIFF
--- a/client/webui/frontend/src/stories/Stepper.stories.tsx
+++ b/client/webui/frontend/src/stories/Stepper.stories.tsx
@@ -1,0 +1,174 @@
+import { defineStepper, type StepperVariant, type StepperLabelOrientation } from "@/lib/components/ui/stepper";
+import { Button } from "@/lib/components/ui/button";
+import type { Meta, StoryContext, StoryFn, StoryObj } from "@storybook/react-vite";
+import { expect, within } from "storybook/test";
+
+// Define the steps for the stepper
+const { Stepper } = defineStepper({ id: "step1", title: "Step 1", description: "First step" }, { id: "step2", title: "Step 2", description: "Second step" }, { id: "step3", title: "Step 3", description: "Third step" });
+
+// Example Stepper Implementation Component
+const StepperExample = ({ variant = "horizontal", labelOrientation = "horizontal" }: { variant?: StepperVariant; labelOrientation?: StepperLabelOrientation }) => {
+    return (
+        <Stepper.Provider variant={variant} labelOrientation={labelOrientation}>
+            {({ methods }) => (
+                <>
+                    <Stepper.Navigation>
+                        <Stepper.Step of="step1">
+                            <Stepper.Title>Account</Stepper.Title>
+                            <Stepper.Description>Create your account</Stepper.Description>
+                        </Stepper.Step>
+                        <Stepper.Step of="step2">
+                            <Stepper.Title>Profile</Stepper.Title>
+                            <Stepper.Description>Set up your profile</Stepper.Description>
+                        </Stepper.Step>
+                        <Stepper.Step of="step3">
+                            <Stepper.Title>Complete</Stepper.Title>
+                            <Stepper.Description>Review and finish</Stepper.Description>
+                        </Stepper.Step>
+                    </Stepper.Navigation>
+
+                    <div style={{ marginTop: "2rem" }}>
+                        <Stepper.Panel>
+                            <div style={{ padding: "1rem", border: "1px solid #ccc", borderRadius: "8px" }}>
+                                <h3>Step Content: {methods.current.title}</h3>
+                                <p>{methods.current.description}</p>
+                            </div>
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/components/ui/stepper.tsx
```diff
@@ -0,0 +1,377 @@
...
+                {typeof children === "function" ? children({ methods }) : children}
+            </div>
+        );
+    };
+
+    return {
+        ...rest,
+        useStepper,
+        Stepper: {
+            Provider: ({ variant = "horizontal", labelOrientation = "horizontal", tracking = false, children, className, ...props }) => {
```
bwiebe-solace, line 42: There's a lot going on in this file which makes it a little hard to follow. Notably, there's a lot of sub-components, and we're also mixing having some defined here vs. as top-level constants in the file. For maintainability, we should probably:

1. Use a single pattern for these components on Stepper, if possible. My preference would be extracting these into separate constants to make the separation of responsibilities clear.
2. Consider splitting-up this file to better divide the responsibilities (eg. creating a "stepper" directory with a few files). Maybe the hook(s) could be in one, the sub-components in another (or even individual files if we think they might grow in complexity over time), reusable types in one, etc. This will also have the benefit of fixing the Copilot-flagged fast refresh warnings.

client/webui/frontend/src/lib/components/ui/stepper.tsx
```diff
@@ -0,0 +1,377 @@
+import { Slot } from "@radix-ui/react-slot";
+import * as Stepperize from "@stepperize/react";
```
bwiebe-solace, line 2: Consider scoping this import more narrowly to help with tree-shaking. If we're concerned about name collisions, we can do something like `import { Stepper as StepperizeStepper }` as needed.

bwiebe-solace:
One more thought: we have a stepper in the config portal UI (ie. when running `sam init --gui`) as well. In the future, we may want to replace that implementation with this common one for consistency and maintainability.

client/webui/frontend/package.json
```diff
@@ -78,6 +78,7 @@
         "@radix-ui/react-slot": "^1.2.3",
         "@radix-ui/react-tabs": "^1.1.12",
         "@radix-ui/react-tooltip": "^1.2.7",
+        "@stepperize/react": "^5.1.9",
```
lgh-solace, line 81: Just curious, do you have any idea of the size of this new import?

client/webui/frontend/src/lib/components/ui/stepper.tsx
```diff
@@ -0,0 +1,377 @@
...
+                const { variant } = useStepperProvider();
+                return (
+                    <nav date-component="stepper-navigation" aria-label={ariaLabel} role="tablist" {...props}>
+                        <ol date-component="stepper-navigation-list" className={classForNavigationList({ variant: variant })}>
+                            {children}
+                        </ol>
+                    </nav>
+                );
+            },
+            Step: ({ children, className, icon, ...props }) => {
```
lgh-solace, line 63: We could consider adding test ids here and there now to facilitate testing.

lgh-solace:
And perhaps a storybook test, if not now then later?

client/webui/frontend/src/lib/components/ui/stepper.tsx
ziyanwan: Thank you, Ben.
We are likely to switch to MRC components in the future. For now, I wouldn't spend too much time on the refactoring.

client/webui/frontend/src/lib/components/ui/stepper.tsx
ziyanwan: Good point, thank you, Ben.
@stepperize/react has 6 outputs, Get, Step, Stepper is from @stepperize/core, and then {type ScopedProps, type StepperReturn, defineStepper}. All of these 6 exports are used in this component. Tree-shaking won't help here.

client/webui/frontend/package.json
ziyanwan: NPM says it has 7.95 kB unpacked. 
As a comparison, the one above it, @radix-ui/react-tooltip has 135 kB, 17 times more.
So I wouldn't concern about the size.

PR #748, opened 2026-01-07
feat(DATAGO-121076): Support @mention in chat

efunneko:
# @Mentions Feature for Chat UI

Video: https://solacedotcom.slack.com/files/U64NFU43C/F0A7B5S2YKH/screen_recording_2026-01-02_at_7.42.55___am.mov

## Summary

This PR adds the ability to @mention users in the chat interface. Users can type `@` followed by a name to search for and mention colleagues, with the mentions displayed as styled chips and sent to the backend with user identification.

### Core Functionality
- **@mention trigger**: Type `@` after a space to open the mention picker
- **Search**: Real-time search across users by name or email
- **Recent mentions**: Quick access to recently mentioned people
- **Visual chips**: Mentions displayed as styled, non-editable chips in the input
- **Duplicate name handling**: Automatic disambiguation when multiple users share the same name (e.g., `@John Doe [john.doe@sales.com]`)

### Backend Integration
- **Identity service integration**: Searches users via configured identity service
- **Feature flag**: Mentions automatically disabled when no identity service is configured
- **Backend format**: Mentions sent as `Name <user_id:email>` format for agent processing

### User Experience
- Keyboard navigation (arrow keys, Enter, Escape)
- Copy/paste support preserving mention data
- Works alongside existing `/` command picker

### Internal Mention Format
Mentions use an internal format `@[Name](id)` that embeds the user's unique ID:
- **Internal**: `@[John Doe](john.doe@solace.com)`
- **Display**: `@John Doe` (or `@John Doe [email]` when disambiguated)
- **Backend**: `John Doe <user_id:john.doe@solace.com>`

This ensures unique identification even when multiple users share the same name.

### Key Components

| File | Purpose |
|------|---------|
| `MentionsCommand.tsx` | Mention picker popup with search and recent mentions |
| `mention-contenteditable.tsx` | ContentEditable input that renders mention chips |
| `mentionUtils.ts` | Utility functions for mention parsing/formatting |
| `recentMentions.ts` | Local storage for recent mentions |
| `ChatInputArea.tsx` | Integration of mentions with chat input |
| `routers/config.py` | Feature flag based on identity_service config |

### State Management
- `mentionMap: Map<string, Person>` - Keyed by person ID for unique identification
- `disambiguatedIds: Set<string>` - Tracks which IDs need disambiguation display

## Configuration

To enable mentions, configure an identity service in your gateway YAML:

```yaml
identity_service:
  type: local_file
  file_path: ./path/to/employees.json
  cache_ttl_seconds: 300
```

When no identity service is configured, the `@` character behaves normally and no mention picker appears.


efunneko:
Fixed the storybook and linting errors

client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
```diff
@@ -451,28 +468,102 @@ export const ChatInputArea: React.FC<{ agents: AgentCardInfo[]; scrollToBottom?:
...
+        const selection = window.getSelection();
+        if (!selection || selection.rangeCount === 0) return 0;
+        if (!chatInputRef.current) return 0;
+
+        const range = selection.getRangeAt(0);
+        let position = 0;
+        let found = false;
+
+        // Walk through all nodes to calculate position in internal format
+        const walker = document.createTreeWalker(
```
lgh-solace, line 483: Do you have any concerns about a lag in the case of large input content when walking the tree on each stroke?

client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
```diff
@@ -451,28 +468,102 @@ export const ChatInputArea: React.FC<{ agents: AgentCardInfo[]; scrollToBottom?:
...
+                    position += range.startOffset;
+                    found = true;
+                } else {
+                    position += node.textContent?.length || 0;
+                }
+            } else if (node.nodeType === Node.ELEMENT_NODE) {
+                const el = node as HTMLElement;
+                if (el.classList.contains("mention-chip")) {
+                    // Check if cursor is inside or after this chip
+                    if (range.startContainer === el || el.contains(range.startContainer)) {
```
lgh-solace, line 516: A bit of repetition here? These lines are called in both conditions.

```
               const internal = el.getAttribute("data-internal") || "";
                position += internal.length;
```

client/webui/frontend/src/lib/components/chat/MessageHoverButtons.tsx
```diff
@@ -15,6 +16,8 @@ interface MessageHoverButtonsProps {
 
 export const MessageHoverButtons: React.FC<MessageHoverButtonsProps> = ({ message, className, textContentOverride }) => {
     const { addNotification } = useChatContext();
+    const { configFeatureEnablement } = useConfigContext();
+    const mentionsEnabled = configFeatureEnablement?.mentions ?? true;
```
lgh-solace: Default should be `false`, yes?

client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
```diff
@@ -771,6 +917,20 @@ export const ChatInputArea: React.FC<{ agents: AgentCardInfo[]; scrollToBottom?:
                 onReservedCommand={handleChatCommand}
             />
 
+            {/* Mentions Command Popover */}
+            {mentionsEnabled && (
```
lgh-solace, line 917: I have been wondering if this feature should be enterprise only.... it would be some work to do so though, so I understand keeping it in community.

client/webui/frontend/src/lib/components/chat/MentionsCommand.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+interface MentionsCommandProps {
+    isOpen: boolean;
+    onClose: () => void;
+    textAreaRef: React.RefObject<HTMLDivElement | HTMLTextAreaElement | null>;
+    onPersonSelect: (person: Person) => void;
+    searchQuery: string;
+}
+
+export const MentionsCommand: React.FC<MentionsCommandProps> = ({ isOpen, onClose, textAreaRef, onPersonSelect, searchQuery }) => {
```
lgh-solace, line 19: I realize it's nice to have the UI line up with the `@` but it feels like it would be simpler if we centered  the command in the view. Thoughts?

If we always did something like this:
image

Then we would avoid the UI from transitioning from connected to unattached on searching.
image

client/webui/frontend/src/lib/components/ui/chat/MentionContentEditable.tsx
```diff
@@ -0,0 +1,413 @@
...
+    mentionMap?: Map<string, Person>; // Map of person ID to Person object
+    disambiguatedIds?: Set<string>; // IDs that need disambiguation display
+}
+
+/**
+ * ContentEditable input with visual mention chips.
+ * Internal format: @[Name](id) - stores person ID for unique identification
+ * Display format: @Name or @Name [email] (when disambiguated)
+ */
+const MentionContentEditable = React.forwardRef<HTMLDivElement, MentionContentEditableProps>(
```
lgh-solace, line 25: Could you rename the file to be camelCase?

client/webui/frontend/src/lib/components/chat/MentionsCommand.tsx
```diff
@@ -0,0 +1,303 @@
...
+            activeElement.scrollIntoView({ behavior: "smooth", block: "nearest" });
+        }
+    }, [activeIndex]);
+
+    if (!isOpen || !popupPosition) return null;
+
+    return (
+        <>
+            {/* Backdrop */}
+            <div ref={backdropRef} className="fixed inset-0 z-40 bg-transparent" onClick={onClose} />
```
lgh-solace, line 196: We do have a popover component that may make a lot of this easier? Did you take a look at it?

client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
efunneko: I don't think that this will be too bad with the expected amount of input content. If you paste a massive amount of content, then it gets converted to an artifact, so we generally are only having 20 or fewer lines of text.

I will say that getting the cursor position working was the hardest part of this change...

client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
efunneko: Yes - that is sloppy. I will fix

client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
efunneko: In community, we don't have all the fancy identity provider stuff and you need to just provide a file of users to make it work. If you don't have an identity provider configured (which to be properly useful is only Enterprise) then this feature is feature flagged off.

client/webui/frontend/src/lib/components/chat/MentionsCommand.tsx
efunneko: I am not too opinionated about this. I thought it was a bit better to be close to the cursor so your eyes didn't have to move too far on a big screen, but could be convinced otherwise. 

I agree that there is a pile of code to try to figure out where to place it but in the days of AI coding, I don't think people will be manually maintaining it anymore

client/webui/frontend/src/lib/components/chat/MessageHoverButtons.tsx
efunneko: For sure. As long as we are in sync with the backend, this should always be there, but we should default to false

client/webui/frontend/src/lib/components/ui/chat/MentionContentEditable.tsx
efunneko: Yes - this was one of the ones I asked you about. This directory has a bunch of kebab cased files and I wasn't sure if we should follow the group, fix them all or be the lone difference.

efunneko:
> Approving now and assuming you will make the change to the default enablement? I believe `const mentionsEnabled = configFeatureEnablement?.mentions ?? true;` is in a couple of places... thanks!

Yup - I have done that, fixed the redundant code and renamed the file

PR #750, opened 2026-01-07
chore: update module path in launch configurations to use relative import

cyrus2281:

Update the module import paths in VS Code launch configurations to use relative imports instead of absolute imports, making the project more modular and development environment more flexible.

## How is this accomplished?

- fix: update module path in launch configurations to use relative import
  - Changed all module paths from `solace_agent_mesh.cli.main` to `cli.main` in the launch.json file

Ensure this change doesn't break any local development environments. The relative imports should work as long as the Python module resolution is properly set up.


enavitan (approved):
changes looks good, however jira task points to a diff description, mind pls addressing it.

PR #756, opened 2026-01-07
fix(DATAGO-121140): Fix loss of embedded templates in sub-agent responses

efunneko:
## Summary

Fixes an issue where `template_liquid` blocks in LLM responses were being stripped when responses were sent back to calling agents in peer-to-peer communication. This caused confusion in the calling agent as they couldn't see or forward the template blocks to the gateway for resolution.

**Root Cause:** When an LLM generates a response containing `template_liquid` blocks, the stream parser extracts the template content and emits a `TemplateBlockData` signal for the gateway to resolve. However, for RUN_BASED sessions (peer-to-peer agent requests), the calling agent only receives the aggregated response text—which had the template blocks stripped out.

**Fix:** For RUN_BASED sessions, when a `TemplateBlockCompletedEvent` is processed, we now preserve the original template block text in the response at its original position. This allows calling agents to forward templates to the gateway for resolution.

### Changes

**`src/solace_agent_mesh/agent/adk/callbacks.py`:**
- Reconstruct and preserve `template_liquid` blocks in the response for RUN_BASED (peer-to-peer) sessions
- Store the original template text in `completed_template_blocks_list` for potential future use
- Add warning log when `TaskExecutionContext` is not found during inline artifact registration

**`src/solace_agent_mesh/agent/sac/component.py`:**
- Add diagnostic logging when no produced artifacts are attached during task finalization
- Add warning when `TaskExecutionContext` is not found during finalization
- Minor formatting fixes

### DIFF
--- a/src/solace_agent_mesh/agent/adk/callbacks.py
+++ b/src/solace_agent_mesh/agent/adk/callbacks.py
@@ -75,6 +75,7 @@ from ...agent.adk.stream_parser import (
     TemplateBlockCompletedEvent,
     ARTIFACT_BLOCK_DELIMITER_OPEN,
     ARTIFACT_BLOCK_DELIMITER_CLOSE,
+    TEMPLATE_LIQUID_START_SEQUENCE,
 )
 
 log = logging.getLogger(__name__)
@@ -583,7 +584,7 @@ async def process_artifact_blocks_callback(
                             [f'{k}="{v}"' for k, v in params.items()]
                         )
                         original_template_text = (
-                            f"{ARTIFACT_BLOCK_DELIMITER_OPEN}template_liquid: {params_str}\n"
+                            f"{TEMPLATE_LIQUID_START_SEQUENCE} {params_str}\n"
                             f"{event.template_content}"
                             f"{ARTIFACT_BLOCK_DELIMITER_CLOSE}"
                         )


src/solace_agent_mesh/agent/adk/callbacks.py
```diff
@@ -569,8 +576,37 @@ async def process_artifact_blocks_callback(
...
-                        # Store template_id in session for potential future use
-                        # (Gateway will handle the actual resolution)
+                        # Reconstruct the original template block text for peer-to-peer responses
+                        # Peer agents don't receive TemplateBlockData signals, so they need
+                        # the original block text to pass templates through to the gateway
+                        params_str = " ".join(
+                            [f'{k}="{v}"' for k, v in params.items()]
+                        )
+                        original_template_text = (
+                            f"{ARTIFACT_BLOCK_DELIMITER_OPEN}template_liquid: {params_str}\n"
```
amir-ghasemi: nitpick: `TEMPLATE_LIQUID_START_SEQUENCE` is already defined in `stream_parser.py` `(TEMPLATE_LIQUID_START_SEQUENCE = f"{ARTIFACT_BLOCK_DELIMITER_OPEN}template_liquid:")` and could be reused here.

efunneko: Good catch! Fixed in bded51c61.

PR #757, opened 2026-01-07
fix(DATAGO-121102): remove all meta versions when deleting artifacts

zhenyu369:

    Brief summary - what problem does this solve?


efunneko (changes requested):
The changes that you have made are at the wrong level. We need to move them to the place where we create the .metadata.json file, which I am pretty sure is in the artifact helpers. 

.metadata.json is a SAM specific file and the ADK doesn't know anything about it. The filesystem_artifact_service.py and s3_artifact_service.py are custom files for the ADK. There are other artifact service files in the ADK that we can't change, but would need this change if we coded it this way.

tests/unit/agent/utils/test_artifact_helpers.py
```diff
@@ -1476,4 +1478,367 @@ def test_filename_operations_with_fixtures(self, sample_csv_metadata):
         # Test with wrong extension
         wrong_ext_filename = filename.replace(".csv", ".txt")
         corrected = ensure_correct_extension(wrong_ext_filename, "csv")
-        assert corrected == filename
\ No newline at end of file
+        assert corrected == filename
+
+
+class TestDeleteArtifactWithMetadata:
+    """Test the delete_artifact_with_metadata function."""
```
efunneko, line 1485: I am really not a fan of these tests. They are heavily mocked and force side effects and then just check those side-effects. It looks like a lot of AI generated tests that verify very little and create a lot of noise. I feel that we should have something that is really testing the behaviour at the integration level that adds an artifact and actually makes sure it is fully deleted. I am not confident that all these many tests actually do that

tests/unit/agent/utils/test_artifact_helpers.py
```diff
@@ -1476,4 +1478,367 @@ def test_filename_operations_with_fixtures(self, sample_csv_metadata):
...
+        # Test success message
+        mock_artifact_service.delete_artifact.side_effect = [None, None]
+        result = await delete_artifact_with_metadata(
+            artifact_service=mock_artifact_service,
+            app_name="testapp",
+            user_id="user123",
+            session_id="session456",
+            filename="test.txt"
+        )
+        assert "Both data artifact 'test.txt' and metadata artifact" in result["message"]
```
efunneko, line 1828: This is a clear example of what I was talking about earlier. We are just checking that the message that is returned says the artifact was removed, but where do we check that the artifact is ACTUALLY removed

zhenyu369: Hi Ed. I added a new integration test in the PR, which should cover the actual artifact creation / deletion scenario

tests/unit/agent/utils/test_artifact_helpers.py
zhenyu369: Hi Ed, I added a new integration test in the PR, which should cover the actual artifact creation / deletion scenario

src/solace_agent_mesh/gateway/http_sse/routers/projects.py
```diff
@@ -465,29 +465,30 @@ async def delete_project_artifact(
...
     user_id = user.get("id")
-    log.info(f"User {user_id} attempting to delete artifact '{filename}' from project {project_id}")
+    log.info(
+        f"User {user_id} attempting to delete artifact '{filename}' "
+        f"from project {project_id}"
+    )
 
     try:
-        success = await project_service.delete_artifact_from_project(
+        result = await project_service.delete_artifact_from_project(
```
lgh-solace: `result` seems unused?

src/solace_agent_mesh/gateway/http_sse/routers/projects.py
```diff
@@ -465,29 +465,30 @@ async def delete_project_artifact(
...
-        log.warning(f"Validation error deleting artifact from project {project_id}: {e}")
-        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
-    except HTTPException:
-        raise
+        log.warning(
+            f"Validation error deleting artifact '{filename}' "
+            f"from project {project_id}: {e}"
+        )
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
```
lgh-solace, line 489: There might be some oddness in the error handling? I think the delete_artifact returns silently if the artifact is not found (which it correct for a REST deletion, I think) so this error should still be a 400. Perhaps just trace through the error handling again?

I will leave Ed to comment on the tests, I think he is back today. Cheers!

src/solace_agent_mesh/gateway/http_sse/routers/projects.py
zhenyu369: I removed the un-used variable

src/solace_agent_mesh/gateway/http_sse/routers/projects.py
zhenyu369: The 404 error is raised against ValueError if:
1. Project not found or access denied
2. Artifact service is not configured

PR #759, opened 2026-01-08
chore(DATAGO-108909): Add adaptive stream rendering of chat result text to smooth it out 

efunneko:

Smooth out the rendering of streaming chunks of text from the backend. This is done to improve the user experience in cases where the chunk size is large and the rendering becomes very stuttery.

With this change, we are now able to greatly increase the batching size in the backend to reduce the large number of status update messages we are getting today, which are only there to improve the smooth rendering.

A new "StreamingMarkdown" component is created that contains adaptive software that measures the rate of responses and chunk size and will automatically adapt to attempt to keep the text rendering running at a constant rate.

- [x] Manual testing: Lot's of different scenarios with straight text output, streaming artifacts, multi-agent, etc. I made heavy use of SAM to create tests in my chat session.

Just the software layout and how react is used. I wouldn't bother too much trying to understand the adaptive algorithm for managing the character rate. That was devised with a lot of manual testing and trying to find a sweet spot of constant rate while still avoiding pauses and too much up front latency


client/webui/frontend/src/lib/components/chat/ChatMessage.tsx
```diff
@@ -108,7 +108,7 @@ const MessageContent = React.memo<{ message: MessageFE }>(({ message }) => {
 
         const embeddedContent = extractEmbeddedContent(displayText);
         if (embeddedContent.length === 0) {
-            return <MarkdownHTMLConverter>{displayText}</MarkdownHTMLConverter>;
+            return isStreaming ? <StreamingMarkdown content={displayText} isStreaming={true} /> : <MarkdownHTMLConverter>{displayText}</MarkdownHTMLConverter>;
```
lgh-solace: I'm wondering if this `isStreaming` prop is redundant? Would you ever use `StreamingMarkdown` when it was false?

lgh-solace: Further, I'm wondering if it might be worth having a small `MarkdownWrapper` (or some better name) with props for `content`,  `isStreaming` and `className` that will  do this repeated logic switch of which one to use?

lgh-solace: Also, looks like tests need updating?

src/solace_agent_mesh/agent/sac/app.py
```diff
@@ -384,7 +384,7 @@ class SamAgentAppConfig(SamConfigBase):
         description="If true, automatically attempts to continue LLM generation if it is interrupted by a token limit.",
     )
     stream_batching_threshold_bytes: int = Field(
-        default=0,
+        default=100,
```
JKaram, line 387: Looks like you're linter made a bunch of changes in this file. Unrelated to this change.

I notice the default is different here then in the `orchestrator_example`. Not sure if that is on purpose or not.

client/webui/frontend/src/lib/components/common/StreamingMarkdown.tsx
```diff
@@ -0,0 +1,117 @@
+import React, { useState, useEffect, useRef } from "react";
+import { MarkdownHTMLConverter } from "./MarkdownHTMLConverter";
+
+interface StreamingMarkdownProps {
+    content: string;
+    isStreaming?: boolean;
+    className?: string;
+}
+
+export const StreamingMarkdown: React.FC<StreamingMarkdownProps> = ({ content, isStreaming, className }) => {
```
JKaram: Nitpick: This is a standalone component file, so convention is to use `export default StreamingMarkdown`

client/webui/frontend/src/lib/components/common/index.ts
```diff
@@ -7,3 +7,4 @@ export { LoadingBlocker } from "./LoadingBlocker";
 export { MarkdownHTMLConverter } from "./MarkdownHTMLConverter";
 export { MessageBanner } from "./MessageBanner";
 export * from "./messageColourVariants";
+export { StreamingMarkdown } from "./StreamingMarkdown";
```
JKaram: Nitpick: There is only one export so it would be better to use `export *`

client/webui/frontend/src/lib/components/common/StreamingMarkdown.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const [displayedContent, setDisplayedContent] = useState(isStreaming ? "" : content);
+    
+    // Use refs for mutable state to avoid re-renders during calculations
+    const state = useRef({
+        cursor: isStreaming ? 0 : content.length,
+        speed: 0.03, // Initial conservative speed (30 chars/sec)
+        lastArrivalTime: 0, // Initialize to 0 to detect first chunk
+        avgInterval: 500, // Estimate 500ms between chunks initially
+        lastLen: isStreaming ? 0 : content.length,
+    });
```
JKaram: I see a lot of "magic numbers" in this file.

We have a constants directory. I think you should move these numbers to be named variables. I would create a file called `/constants/chat.ts` (or something like that) and import all these numbers into this file. That way we can maintain them better if they are used in other spots in the UI or need to be synced with the BE.

 By magic numbers I mean these: 
 

-  0 — Used several times to initialize values such as lastArrivalTime, check cursor/index values, and detect first chunks.
- 0.03 — Initial value for speed (30 chars/sec).
- 500 — Initial value for avgInterval (500ms between chunks).
- 0.05 — Alpha value for moving average update (when dt < 200 ms).
- 0.2 — Alpha value for moving average update (when dt >= 200 ms).
- 200 — Threshold in ms to decide which alpha to use.
- 5000 — The max allowed clamp value for dt (to cap extreme time gaps).
- 20 — The minimum allowed clamp value for dt.
- 1.5 — Safety factor (adds "buffer" time on interval calculation).
- 0.5 — Momentum factor used when increasing speed.
- 0.8 — Momentum factor used when decreasing speed.
- 0.005 — Minimum possible speed value.
- 3.0 — Maximum possible speed value.

client/webui/frontend/src/lib/components/common/StreamingMarkdown.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+            animationFrameId = requestAnimationFrame(animate);
+        };
+
+        animationFrameId = requestAnimationFrame(animate);
+
+        return () => {
+            cancelAnimationFrame(animationFrameId);
+        };
+    }, [isStreaming]); // content dependency removed from effect, accessed via ref
```
JKaram: Logic looks good on these effects . 

I would abstract these into hooks like `useStreamingSpeed` and `useStreamingAnimation` for reusability. AI will do this really quickly for you.

JKaram (commented):
Nice improvement! A few comments. The nitpicks are up to you.

src/solace_agent_mesh/agent/sac/app.py
efunneko: This is intentional. With this smoothing in place, we are able to raise the default streaming batch amount which increases efficiency. Me wanting to raise this was what drove the smooth rendering.

The example is even higher than the default, which is okay.

client/webui/frontend/src/lib/components/chat/ChatMessage.tsx
efunneko: Thanks - I have applied the first two comments.

For tests, do you mean that there are a few tests that are failing (I have fixed these) or that we should add tests for the UI for this?

client/webui/frontend/src/lib/components/common/index.ts
efunneko: I guess you can't do both `export default` and do `export *` like this. I just fixed this one and not the previous

lgh-solace (approved):

PR #767, opened 2026-01-09
fix(DATAGO-121333): Change MCP results to 'all or none' and allow full configuration of schema discovery depth

efunneko:
## Summary

   - **All-or-nothing approach for large MCP responses**: When MCP tool responses exceed the LLM return size limit, the data is now completely withheld
   rather than being truncated. This prevents LLM hallucination caused by partial data - LLMs tend to confidently fill in gaps when given truncated
   content.
   - **Configurable schema inference depth**: Added `schema_inference_depth` configuration (default: 4, range: 1-10) that can be set at the agent level or
   per-tool level via `tool_config`. This controls how deeply nested structures are inspected when generating artifact metadata.
   - **Improved LLM guidance**: When data is withheld, the LLM receives explicit instructions to use `template_liquid` (for displaying to users) or
   `load_artifact` (for processing) to access the full data from the saved artifact.

   ### Key Changes

   | File | Change |
   |------|--------|
   | `callbacks.py` | Removed truncation logic; now returns `data_in_artifact_only` type with guidance message |
   | `intelligent_mcp_callbacks.py` | Added `_get_schema_config_from_tool_or_agent()` helper for per-tool config override; passes `schema_inference_depth`
   to artifact saving |
   | `app.py` | Added `schema_inference_depth` field to `SamAgentAppConfig` |
   | `artifact_helpers.py` | Enhanced `save_artifact_with_metadata()` to resolve depth from wrapper config if not provided |

### DIFF
--- a/src/solace_agent_mesh/agent/adk/intelligent_mcp_callbacks.py
+++ b/src/solace_agent_mesh/agent/adk/intelligent_mcp_callbacks.py
@@ -19,6 +19,7 @@ from .mcp_content_processor import MCPContentProcessor, MCPContentProcessorConfi
 from ...agent.utils.artifact_helpers import (
     save_artifact_with_metadata,
     DEFAULT_SCHEMA_MAX_KEYS,
+    DEFAULT_SCHEMA_INFERENCE_DEPTH,
 )
 from ...agent.utils.context_helpers import get_original_session_id
 
@@ -335,7 +336,7 @@ async def _save_content_item_as_artifact(
             tool, host_component, "schema_max_keys", DEFAULT_SCHEMA_MAX_KEYS
         )
         schema_inference_depth = _get_schema_config_from_tool_or_agent(
-            tool, host_component, "schema_inference_depth", 4
+            tool, host_component, "schema_inference_depth", DEFAULT_SCHEMA_INFERENCE_DEPTH
         )
         artifact_timestamp = datetime.now(timezone.utc)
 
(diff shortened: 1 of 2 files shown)


tests/unit/agent/adk/test_intelligent_mcp_callbacks.py
```diff
@@ -0,0 +1,159 @@
+"""
+Unit tests for intelligent MCP callback functions.
+Tests cover schema configuration resolution from tool config and agent config.
+"""
+
+from unittest.mock import Mock
+
+from solace_agent_mesh.agent.adk.intelligent_mcp_callbacks import (
+    _get_schema_config_from_tool_or_agent,
+    DEFAULT_SCHEMA_MAX_KEYS,
```
amir-ghasemi: Consider importing directly from `artifact_helpers` where `DEFAULT_SCHEMA_MAX_KEYS`  is originally defined.

src/solace_agent_mesh/agent/adk/intelligent_mcp_callbacks.py
```diff
@@ -294,8 +330,12 @@ async def _save_content_item_as_artifact(
...
         user_id = tool_context._invocation_context.user_id
         session_id = get_original_session_id(tool_context._invocation_context)
-        schema_max_keys = host_component.get_config(
-            "schema_max_keys", DEFAULT_SCHEMA_MAX_KEYS
+        # Get schema config from tool_config (per-tool) or agent config (fallback)
+        schema_max_keys = _get_schema_config_from_tool_or_agent(
+            tool, host_component, "schema_max_keys", DEFAULT_SCHEMA_MAX_KEYS
+        )
+        schema_inference_depth = _get_schema_config_from_tool_or_agent(
+            tool, host_component, "schema_inference_depth", 4
```
amir-ghasemi: Import and use `DEFAULT_SCHEMA_INFERENCE_DEPTH` instead of the hardcoded '4' constant.

PR #786, opened 2026-01-12
feat(DATAGO-119365): Update S3 bucket documentation for connector specs and artifacts

bwiebe-solace:

This PR updates the documentation around the new S3 connector specs bucket.

Updated artifact-storage.md file to have a description of the buckets and their purposes, aligned with other docs in other repos.

n/a

n/a

### DIFF
--- a/docs/docs/documentation/enterprise/connectors/openapi-connectors.md
+++ b/docs/docs/documentation/enterprise/connectors/openapi-connectors.md
@@ -33,9 +33,11 @@ You need an OpenAPI specification file in JSON or YAML format that describes the
 
 ### Public Storage Bucket Configuration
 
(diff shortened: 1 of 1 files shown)


docs/docs/documentation/installing-and-configuring/artifact-storage.md
```diff
@@ -42,6 +42,34 @@ Contrast with session storage:
 
 For session storage configuration, see [Session Storage](./session-storage.md).
 
+### Multiple S3 Buckets for OpenAPI Connector Feature
```
mo-radwan1, line 45: It looks good to me the only thing for me is the name of the file `artifact-storage` I don't think this would fit under artifact storage ? correct me if I am wrong. I will approve the PR but please give it some thought if we need to rename the tab or fit it under a different section ? i am not sure if artifact storage still applies

PR #794, opened 2026-01-13
chore(DATAGO-121446): WF PR 13 - Workflow node detail panel and diagram refinements

efunneko:
## Summary
- Enhances the workflow node detail panel with comprehensive property display
- Aligns workflow diagram visualization with HLM design specifications
- Adds expression parser for highlighting referenced nodes on hover
- Improves Switch, Map, and Loop node rendering

### Node Detail Panel
- Expanded `WorkflowNodeDetailPanel` with detailed property views for all node types
- Added `InputMappingViewer` component for displaying input/output mappings
- Added `WorkflowDetailsSidePanel` for workflow-level information
- Expression references highlight source nodes on hover

### Diagram Refinements (HLM Alignment)
- **SwitchNode**: Shows numbered case rows inside the node with conditions
- **ConditionPillNode**: Added case number prefix, more rectangular shape
- **MapNode/LoopNode**: Solid header box straddling dotted children container
- Collapsed Map/Loop nodes render as simple nodes without dotted container
- Removed redundant type badges from Map/Loop nodes
- Fixed switch node height calculation based on number of cases

### Other
- Added expression parser utility for extracting node references from template strings
- Updated layout engine for proper switch node sizing
- Various styling improvements across node components

## Test plan
- [ ] Click on different node types and verify detail panel shows correct properties
- [ ] Hover over expressions in detail panel - verify referenced nodes highlight
- [ ] Verify Switch nodes show case rows inside with proper spacing
- [ ] Test Map/Loop expand/collapse - verify edges connect correctly
- [ ] Verify condition pills show case numbers (1, 2, etc.) and "Default"

### DIFF
--- a/client/webui/frontend/src/lib/components/workflowVisualization/CanvasControls.tsx
+++ b/client/webui/frontend/src/lib/components/workflowVisualization/CanvasControls.tsx
@@ -36,7 +36,7 @@ export const CanvasControls: React.FC<CanvasControlsProps> = ({
     const isAtMaxZoom = zoomLevel >= maxZoom;
 
     return (
-        <div className="flex items-center gap-2 border-b border-gray-200 bg-white px-4 py-2 dark:border-gray-700 dark:bg-gray-800">
+        <div className="flex items-center justify-end gap-2 border-b border-gray-200 bg-white px-4 py-2 dark:border-gray-700 dark:bg-gray-800">
             {/* Fit to view / Center button */}
             <button
                 onClick={onFitToView}
--- a/client/webui/frontend/src/lib/components/activities/FlowChart/PanZoomCanvas.tsx
+++ b/client/webui/frontend/src/lib/components/activities/FlowChart/PanZoomCanvas.tsx
@@ -411,8 +411,20 @@ const PanZoomCanvas = React.forwardRef<PanZoomCanvasRef, PanZoomCanvasProps>(
 
         // Handle double-click to zoom in at click location
         const handleDoubleClick = useCallback((e: React.MouseEvent) => {
-            // Only handle if clicking on the container background (not on nodes)
-            if (e.target !== containerRef.current) return;
+            // Prevent text selection on double-click
+            e.preventDefault();
+
+            // Only handle if clicking on the container background (not on interactive nodes)
+            // Check if the click target is a button, link, or has data-no-zoom attribute
+            const target = e.target as HTMLElement;
+            if (
+                target.closest("button") ||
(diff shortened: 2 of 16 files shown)


client/webui/frontend/src/lib/components/workflowVisualization/CanvasControls.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const zoomPercentage = Math.round(zoomLevel * 100);
+
+    // Determine if buttons should be disabled
+    const isAtMinZoom = zoomLevel <= minZoom;
+    const isAtMaxZoom = zoomLevel >= maxZoom;
+
+    return (
+        <div className="flex items-center justify-end gap-2 border-b border-gray-200 bg-white px-4 py-2 dark:border-gray-700 dark:bg-gray-800">
+            {/* Fit to view / Center button */}
+            <button
```
lgh-solace, line 41: Do we need to use the native `button` here and not the `Button` component which brings some styles with it? It should be updated to use the standard component if possible.

client/webui/frontend/src/lib/components/workflowVisualization/IMPLEMENTATION_STATUS.md
```diff
@@ -0,0 +1,131 @@
+# Workflow Visualization - Expression Highlighting Feature
+
+## Overview
```
lgh-solace, line 3: Do we want to include this document? It contains a suggestion below for hard-coded colours so if we want to deliver it, it should be updated.

lgh-solace (approved):
For this PR, there are a number of places the native `button` was used over the standard `Button` component. There are times when this is necessary, but I suspect some of these could use the standard component (with its standard styles).

  Priority 1 - Replace native buttons:
  1. ✏️ CanvasControls.tsx - 3 buttons
  2. ✏️ WorkflowNodeDetailPanel.tsx - 5 buttons
  3. ✏️ InputMappingViewer.tsx - 1 button
  4. ✏️ WorkflowVisualizationPage.tsx - 1 button
  5. ✏️ LoopNode.tsx - 2 buttons
  6. ✏️ MapNode.tsx - 2 buttons
  7. ✏️ WorkflowRefNode.tsx - 1 button
  
  Outside of these buttons, there is a committed `md` file that seems out of place?
  
  CC: @solace-pjones

efunneko:
This PR has been merged into `feat/workflows` as part of the consolidated Workflows feature branch. Closing this stacked PR.

PR #796, opened 2026-01-13
ci: add ability to release exact version

artyom-morozov:

 Need ability to release exact version to PYPI to fix a release workflow

Pass the exact version to the release workflow 

can only validate during release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -184,7 +184,7 @@ jobs:
         id: prep
         uses: SolaceDev/solace-public-workflows/.github/actions/hatch-release-prep@main
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -171,7 +184,7 @@ jobs:
         id: prep
         uses: SolaceDev/solace-public-workflows/.github/actions/hatch-release-prep@main
         with:
-          version: ${{ github.event.inputs.version }}
+          version: ${{ github.event.inputs.exact_version != '' && github.event.inputs.exact_version || github.event.inputs.version }}
```
johnvincentcorpuz: I think this is sufficient: version: ${{ github.event.inputs.exact_version || github.event.inputs.version }}

johnvincentcorpuz: exact version takes precedence over inputs.version

PR #800, opened 2026-01-13
fix(DATAGO-121281): Fix Pydantic Schema Validation for Enum values

Hugo-Pare:

Fixes Pydantic validation errors when MCP (Model Context Protocol) servers provide tool schemas with integer enum values. Previously, the schema conversion would fail with `ValidationError: Input should be a valid string` when encountering enums like `{"enum": [3, 7, 14]}`, which are valid per JSON Schema specification but not accepted by Pydantic's string-only enum validation.

Added a `_normalize_schema_dict()` helper function in [src/solace_agent_mesh/agent/adk/models/lite_llm.py](src/solace_agent_mesh/agent/adk/models/lite_llm.py) that:
1. Recursively traverses schema dictionaries (properties, items, nested structures)
2. Converts any enum values to strings before Pydantic validation
3. Is called before `types.Schema.model_validate()` at two locations in `_schema_to_dict()` (lines 348 and 361)

This normalization layer sits between MCP server schemas and Pydantic validation, ensuring compatibility without modifying the core schema conversion logic.

### Key Design Decisions

**Why normalize instead of modifying Pydantic models?**
- JSON Schema spec allows enums of any type (string, number, integer, boolean, null)
- Converting to strings is safe for LLM tool parameters since most LLMs accept string-typed parameters
- This approach maintains backward compatibility while supporting spec-compliant MCP servers
- Keeps changes minimal and localized to the conversion layer

**Why recursive normalization?**
- MCP tools often have deeply nested schemas with enums at various levels
- Handles complex real-world scenarios like status codes in nested filters or priority levels in array items

- [x] Manual testing: Verified Python syntax compilation passes
- [x] Unit tests: Added `test_mcp_tool_with_integer_enums()` in [tests/unit/agent/adk/models/test_lite_llm_schema_serialization.py](tests/unit/agent/adk/models/test_lite_llm_schema_serialization.py#L266-L303)
  - Tests integer enums in simple properties
  - Tests integer enums in nested array items
  - Verifies no Pydantic validation errors
  - Confirms enum values are stringified in output
- [ ] Integration tests: Should be tested with actual MCP server providing integer enums
- [ ] Known limitations: Only tested with integer enums; boolean/null enums not explicitly tested but should work via `str()` conversion

1. **Mutation vs immutability**: The `_normalize_schema_dict()` function mutates the input dictionary. This is acceptable since it's called on temporary dictionaries during conversion, but reviewers should verify no shared references could cause issues.

2. **Performance**: Recursive normalization adds minimal overhead, but reviewers should consider if this impacts schema conversion performance for deeply nested structures (though these are rare).

3. **Edge cases**: Review whether other enum types (boolean, null, mixed types) would be handled correctly by the `str()` conversion approach.

4. **Test coverage**: The new test uses `_normalize_schema_dict` directly. Reviewers may want to verify if additional integration testing with real MCP servers is needed before merging.

### DIFF
--- a/tests/unit/agent/adk/models/test_lite_llm_schema_serialization.py
+++ b/tests/unit/agent/adk/models/test_lite_llm_schema_serialization.py
@@ -263,6 +263,45 @@ class TestSchemaToDict:
         assert result["properties"]["name"]["type"] == "string"
         assert result["properties"]["email"]["type"] == "string"
 
+    def test_mcp_tool_with_integer_enums(self):
+        """Test MCP tool schema with integer enums are properly handled.
+
+        MCP servers can provide integer enums which is valid per JSON Schema spec.
+        This test ensures the schema normalization handles this without Pydantic errors.
+        """
+        from solace_agent_mesh.agent.adk.models.lite_llm import _normalize_schema_dict
+
+        # Simulate MCP tool schema with integer enums in nested structure
+        mcp_tool_params = types.Schema(
+            type=types.Type.OBJECT,
+            properties={
+                "priority": types.Schema.model_validate(
+                    _normalize_schema_dict({"type": "integer", "enum": [3, 7, 14]})
+                ),
+                "filters": types.Schema(
+                    type=types.Type.OBJECT,
+                    properties={
+                        "status_codes": types.Schema.model_validate(
+                            _normalize_schema_dict({"type": "array", "items": {"type": "integer", "enum": [200, 404, 500]}})
+                        )
+                    }
+                )
+            }
+        )
+
+        func = types.FunctionDeclaration(
+            name="mcp_test_tool",
+            description="Test MCP tool with integer enums",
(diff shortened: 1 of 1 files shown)


src/solace_agent_mesh/agent/adk/models/lite_llm.py
```diff
@@ -147,33 +150,33 @@ def _safe_json_serialize(obj) -> str:
...
     if prefix_length < 1:
         prefix_length = 1
-    
+
     prefix = tool_call_id[:prefix_length]
     # Use SHA256 and take first 32 hex characters for uniqueness
     hash_suffix = hashlib.sha256(tool_call_id.encode()).hexdigest()[:32]
-    
+
     return f"{prefix}_{hash_suffix}"
```
Hugo-Pare, line 182: Everything above is just cleanup

PR #806, opened 2026-01-14
chore(DATAGO-119380): Update the vibe coding document 

alimosaed:
The instruction was not straightforward.

Simplified the instruction.

Not applicable

Nothing


docs/docs/documentation/vibe_coding.md
```diff
@@ -37,14 +37,40 @@ Before you begin, ensure you have:
...
 
-Review the response from your coding assistant to confirm that it acknowledges the Context7 integration and recognizes the `solacelabs/solace-agent-mesh` library. Depending on the assistant, you might see confirmations such as:  
-`I've successfully configured the solacelabs/solace-agent-mesh context7 library ...`  
-or  `I'll use the solacelabs/solace-agent-mesh library ...`  
+Review the response from your coding assistant to confirm that it acknowledges the Context7 integration and recognizes the `solacelabs/solace-agent-mesh` library. Depending on the assistant, you might see different confirmation messages. The following shows example messages for when the MCP is connected and disconnected.
+
+<table>
+<tr>
+<td><span style="color:green">Connected</span></td>
+<td><img src="../../static/img/doc-correct-context7-connection.png" alt="Correct Context7 connection confirmation"/></td>
```
solace-mdupls: These images are missing from branch.

docs/docs/documentation/vibe_coding.md
```diff
@@ -37,14 +37,40 @@ Before you begin, ensure you have:
...
 This ensures your coding agent is correctly set up to leverage the Solace Agent Mesh knowledge base.
 
+:::note
+You need to specify the library at least once in your chat session to activate the Context7 integration.
+To avoid mentioning `solacelabs/solace-agent-mesh` context7 library in every prompt, configure your MCP client with a rule that automatically triggers the library for any code-related request.
+
+```
+Claude Code: CLAUDE.md
+Or the equivalent configuration in your MCP client.
+```
```
solace-mdupls: After further thought, you can remove this part that specifically talks about `CLAUDE.md` @alimosaed. CLAUDE.md files may be completely obsolete one day or be replaced by something else. We don't want to have to update our documentation because they've decided to introduce something different.

docs/docs/documentation/vibe_coding.md
```diff
@@ -37,14 +37,35 @@ Before you begin, ensure you have:
...
 
+:::note
+You need to specify the library at least once in your chat session to activate the Context7 integration.
+To avoid mentioning `solacelabs/solace-agent-mesh` context7 library in every prompt, configure your MCP client with a rule that automatically triggers the library for any code-related request.
+
+Example rule:
+```
+Automatically use `solacelabs/solace-agent-mesh` context7 library without requiring me to explicitly request it.
+```
+:::
```
solace-mdupls, line 51: Do you think we need this example given that we mention

> Use the `solacelabs/solace-agent-mesh` context7 library when answering questions in this chat session.

above? The note may be sufficient enough.

docs/docs/documentation/vibe_coding.md
```diff
@@ -37,14 +37,35 @@ Before you begin, ensure you have:
...
+<table>
+<tr>
+<td><span style="color:green">Connected</span></td>
+<td><img src="../../static/img/doc-correct-context7-connection.png" alt="Correct Context7 connection confirmation"/></td>
+</tr>
+<tr>
+<td><span style="color:red">Disconnected</span></td>
+<td><img src="../../static/img/doc-incorrect-context7-connection.png" alt="Incorrect Context7 connection confirmation"/></td>
+</tr>
+</table>
```
solace-mdupls: @alimosaed when running the docs locally, I get this error on the vibe coding page:

> The `style` prop expects a mapping from style properties to values, not a string. For example, style={{marginRight: spacing + 'em'}} when using JSX.

Use markdown table syntax unless html syntax is absolutely needed. If you're using html to simply specify the background color because that doesn't work in markdown, then let's pivot away from using background color. An option is to use a checkmark like so: :white_check_mark:

However, after some thought and going back to what we discussed, do these two images:
1. Provide more clarity than before? Will those not using Claude be more confused because they're not familiar with the response structure?
1. Add tech debt as they quickly become stale-over-time?
Image

Looking at what you have already, I think this alone is sufficient:

> Review the response from your coding assistant to confirm that it acknowledges the Context7 integration and recognizes the solacelabs/solace-agent-mesh library.

docs/docs/documentation/vibe_coding.md
solace-mdupls: > To avoid mentioning solacelabs/solace-agent-mesh context7 library in every prompt, configure your MCP client with a rule that automatically triggers the library for any code-related request.

I thought we discussed that in most cases this won't be required (ie. you are just creating a SAM project)? 

So this can be changed to a `:::tip` and reworded to mention that you can avoid having to prompt your coding agent to use the Context7 library by configuring your MCP client with a rule. Then specify the example.

docs/docs/documentation/vibe_coding.md
```diff
@@ -37,14 +37,35 @@ Before you begin, ensure you have:
...
+
+
 This ensures your coding agent is correctly set up to leverage the Solace Agent Mesh knowledge base.
```
solace-mdupls: Instead of this line, you can mention something along the lines of: "Once you've verified your coding agent is correctly using the Context7 library, you can move to the next step."

docs/docs/documentation/vibe_coding.md
alimosaed: I think mentioning the library name for verification is necessary to ensure that they connects to context7 and find the library.

docs/docs/documentation/vibe_coding.md
alimosaed: As we discussed in call, I deleted these sample images.

docs/docs/documentation/vibe_coding.md
alimosaed: Since we removed the images and cleaned up the text, I think we no longer need this line. Devs know that mandatory steps must be followed one by one.

PR #827, opened 2026-01-16
chore(DATAGO-117142):  moving token clearing to avoid race conditions for refreshing

lgh-solace:

    There was a race condition between logging out and refreshing the token

This pull request updates the logout flow in the `AuthProvider` component to improve reliability and prevent race conditions during token refresh. The most important changes are:

Authentication flow improvements:
* Moved the removal of `access_token` and `refresh_token` from after the logout API call to before it, ensuring tokens are cleared from `localStorage` immediately to prevent race conditions. (`client/webui/frontend/src/lib/providers/AuthProvider.tsx`, [client/webui/frontend/src/lib/providers/AuthProvider.tsxL77-R92](diffhunk://#diff-8b359133dc616535b6c6ac8f33cbfe43fc5ac561c6c7ecff9692c7adec44e13eL77-R92))
* Updated the logout process to always redirect to the login page in a `finally` block, guaranteeing users are redirected even if an error occurs during logout. (`client/webui/frontend/src/lib/providers/AuthProvider.tsx`, [client/webui/frontend/src/lib/providers/AuthProvider.tsxL77-R92](diffhunk://#diff-8b359133dc616535b6c6ac8f33cbfe43fc5ac561c6c7ecff9692c7adec44e13eL77-R92))
* Improved error logging by updating the error message to "Error during logout" for clarity. (`client/webui/frontend/src/lib/providers/AuthProvider.tsx`, [client/webui/frontend/src/lib/providers/AuthProvider.tsxL77-R92](diffhunk://#diff-8b359133dc616535b6c6ac8f33cbfe43fc5ac561c6c7ecff9692c7adec44e13eL77-R92))

Feature protected by feature flag - currently unused - requires manual testing in enterprise deployments.


client/webui/frontend/src/lib/providers/AuthProvider.tsx
```diff
@@ -74,20 +74,22 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
...
                 clearCsrfToken();
 
-                // Clear tokens from localStorage - set in authCallback.tsx
-                localStorage.removeItem("access_token");
-                localStorage.removeItem("refresh_token");
-
-                // Force redirect to login - prevents token refresh
-                window.location.href = configAuthLoginUrl;
+                // Call logout API - best effort only
+                await api.webui.post("/api/v1/auth/logout");
```
lgh-solace, line 87: Theoretically, we don't actually need this call - if the FE no longer has a token, then they can no longer exercise any api and are logged out. We are still calling the function though as a best effort. And, it may be useful for API users.

JKaram: Just my opinion but I think the finally block should handle most of the logic here. 

```
const logout = async () => {
      if (!configUseAuthorization) return;

      try {
          await api.webui.post("/api/v1/auth/logout");
      } catch (error) {
          console.error("Error calling logout endpoint:", error);
      } finally {
          setIsAuthenticated(false);
          setUserInfo(null);
          clearCsrfToken();
          localStorage.removeItem("access_token");
          localStorage.removeItem("refresh_token");
          window.location.href = configAuthLoginUrl;
      }
  };
  ```

lgh-solace: I've been playing with the ordering for a few tries. I used to have that clearing in the `finally`, but we need to clear the tokens first or the UI will automatically refresh... it's been "fun".

PR #830, opened 2026-01-16
fix(DATAGO-122082): Update concurrent session integration test

alimosaed:
There is an error in `tests/integration/scenarios_programmatic/gateway/test_http_sse_gateway.py` test.
**Details:** When authorization is disabled, all sessions are assigned the same hardcoded client ID "sam_dev_user" regardless of the actual user identity. This means that three concurrent test users with distinct email addresses (concurrent_user_0@example.com, concurrent_user_1@example.com, concurrent_user_2@example.com) all share the same session storage. When these users make simultaneous requests, a race condition occurs where session data gets overwritten, causing responses intended for one user to be delivered to another—hence why session 1 receives "concurrent session 2" instead of its expected "concurrent session 1" response.

Skipped this test and added a corresponding test to check concurrent users' requests.

### Key Design Decisions
This test is not working without authentication. Also, the gateway integration tests do not handle authentication. Adding a corresponding test can support the test scenario.

Nothing

### DIFF
--- a/tests/integration/apis/persistence/test_concurrent_sessions_with_auth.py
+++ b/tests/integration/apis/persistence/test_concurrent_sessions_with_auth.py
@@ -6,6 +6,7 @@ sessions from different users maintain proper isolation and don't interfere
 with each other.
 """
 
+import asyncio
 import pytest
 import uuid
 from fastapi.testclient import TestClient
@@ -25,10 +26,10 @@ async def test_concurrent_sessions_with_different_users(
     """
     Test that concurrent sessions from different users maintain proper isolation.
     
-    This test verifies that when multiple users make sequential requests,
+    This test verifies that when multiple users make CONCURRENT requests,
     each user receives their own response without cross-contamination.
     
-    Note: Uses gateway_adapter to directly create sessions and messages,
+    Note: Uses asyncio.gather() to create truly concurrent operations,
     simulating multiple users with proper database persistence.
     """
     scenario_id = "concurrent_sessions_with_auth_001"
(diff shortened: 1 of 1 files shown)


tests/integration/apis/persistence/test_concurrent_sessions_with_auth.py
```diff
@@ -0,0 +1,309 @@
...
+        print(f"✓ User {user_id} session has correct isolated history")
+    
+    print(f"\nScenario {scenario_id}: All concurrent sessions properly isolated ✓")
+
+
+@pytest.mark.all
+@pytest.mark.asyncio
+@pytest.mark.api
+@pytest.mark.gateway
+async def test_concurrent_session_creation(api_client: TestClient):
```
enavitan, line 152: this is not quite concurrent isn't it ? for it to be concurrent we'd need to spawn a thread per user and send requests concurrently, prob rename method to test_multi_session_creation

tests/integration/apis/persistence/test_concurrent_sessions_with_auth.py
```diff
@@ -0,0 +1,309 @@
...
+
+# Custom header for test user identification
+TEST_USER_HEADER = "X-Test-User-Id"
+
+
+@pytest.mark.all
+@pytest.mark.asyncio
+@pytest.mark.api
+@pytest.mark.gateway
+async def test_concurrent_sessions_with_different_users(
```
enavitan, line 23: this is not concurrent rather sequencial, since we block on each call to pull session info and print it, lets rename it accordingly, test_session_isolation_with_different_users ?

alimosaed: Right. I updated the method.

tests/integration/apis/persistence/test_concurrent_sessions_with_auth.py
alimosaed: Right. I updated the method.

PR #831, opened 2026-01-16
fix: update integration test for local file identity 

amir-ghasemi:
This pull request updates the integration test for the local file identity service to reflect changes in the user object format returned by `search_users`. The test now checks for the presence of `displayName` instead of `name`, ensuring compatibility with the canonical user format.

* Updated the assertion in `test_local_file_identity_service_search_users` to check for `displayName` instead of `name`, reflecting the new canonical user format returned by `search_users`.


efunneko (approved):
LG

PR #838, opened 2026-01-19
feat(DATAGO-122217): Add ssl_verify config option for A2A proxy

gregmeldrum:

Add a boolean configuration option (`ssl_verify`) to disable SSL certificate verification for individual A2A proxied agents. This allows connecting to remote A2A agents using self-signed certificates in development/testing environments.

- **config.py**: Added `ssl_verify` field to `A2AProxiedAgentConfig` with default value `True`
- **component.py**: Applied `ssl_verify` setting to httpx clients for both agent card fetching and task invocation
- **proxies.md**: Added documentation section on SSL verification with security warnings
- **a2a_proxy_example.yaml**: Added commented example showing `ssl_verify: false` usage

### Key Design Decisions

- Default is `True` (verify certificates) to maintain security by default
- Setting applies to both agent card fetching and task invocations for consistency
- Per-agent configuration allows mixed environments (some verified, some not)

- [x] Unit tests: Added 7 new tests in `tests/unit/agent/proxies/a2a/test_config.py`
- [x] Ruff linting: All checks pass
- [ ] Manual testing: Not yet tested with actual self-signed certificate

- Security documentation warns users about the risks of disabling SSL verification
- The setting only affects HTTPS connections; HTTP connections are unaffected

### DIFF
--- a/src/solace_agent_mesh/agent/proxies/a2a/component.py
+++ b/src/solace_agent_mesh/agent/proxies/a2a/component.py
@@ -628,6 +628,13 @@ class A2AProxyComponent(BaseProxyComponent):
                 log.debug("%s Fetching agent card without authentication", log_identifier)
 
             ssl_verify = agent_config.get("ssl_verify", True)
+            if not ssl_verify:
+                log.warning(
+                    "%s SSL verification disabled for agent '%s'. "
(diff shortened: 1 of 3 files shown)


RohanSreelesh (approved):
LG

PR #839, opened 2026-01-19
fix(DATAGO-122233): handle string indices in deep_research sources selection

amir-ghasemi:
This pull request improves the robustness of the `_select_sources_to_fetch` function by handling cases where the LLM returns indices as strings instead of integers. This change prevents errors and adds logging for invalid index values.

Error handling and logging improvements:

* Updated the loop that processes `selected_indices` to convert each index to an integer, handling cases where the LLM returns strings instead of integers, and logging a warning if the index is invalid.


gregmeldrum (approved):
LG

PR #850, opened 2026-01-21
feat(DATAGO-123135): Add ability to preview binary artifacts (PDF. PPTX, DOCX)

amir-ghasemi:
This pull request adds support for previewing Office documents (DOCX, PPTX) and PDF files in the web UI by integrating server-side LibreOffice conversion and a new PDF renderer. It introduces new frontend components and updates the artifact preview logic to handle these file types, including larger file size limits for URL-based previews. Additionally, it installs the necessary LibreOffice packages in the Docker image and updates dependencies for PDF rendering.

**Document Preview Enhancements:**

* Added `OfficeDocumentRenderer` React component to convert DOCX and PPTX files to PDF using a server-side LibreOffice service, with caching to avoid redundant conversions and graceful fallback if conversion is unavailable.
* Introduced `PdfRenderer` and integrated `react-pdf` for in-browser PDF previews, supporting both converted Office documents and direct PDF files. [[1]](diffhunk://#diff-7a1a6c384e0a090250fa886b2d9c70a107a16cc02d9bf53b492544ad575dc7dbL3-R17) [[2]](diffhunk://#diff-7a1a6c384e0a090250fa886b2d9c70a107a16cc02d9bf53b492544ad575dc7dbR34-R44) [[3]](diffhunk://#diff-59fc6bf7d75fdd9c6b95d53c759c680af3452422fa44da344768f04d213bcdadR18-R19) [[4]](diffhunk://#diff-2a0b1b97cd2a8a0f6ee43d592510c4b820af1558c2225e11df9b45533a36e36cR101) [[5]](diffhunk://#diff-2a0b1b97cd2a8a0f6ee43d592510c4b820af1558c2225e11df9b45533a36e36cR132)

**Preview Logic and File Type Detection:**

* Extended preview utility functions to detect DOCX, PPTX, and PDF files by filename or MIME type, and route them to the appropriate renderer. [[1]](diffhunk://#diff-5ef33e5331e80d169ca2dea093de2d7105cbd819b9d7e51706bc87d991d4e6e7R222-R272) [[2]](diffhunk://#diff-5ef33e5331e80d169ca2dea093de2d7105cbd819b9d7e51706bc87d991d4e6e7R313-R324)
* Updated artifact preview logic to support content and URL-based renderers, including larger size limits (up to 100MB) for URL-based previews like PDF. [[1]](diffhunk://#diff-5ef33e5331e80d169ca2dea093de2d7105cbd819b9d7e51706bc87d991d4e6e7L295-R363) [[2]](diffhunk://#diff-5ef33e5331e80d169ca2dea093de2d7105cbd819b9d7e51706bc87d991d4e6e7R374-R383) [[3]](diffhunk://#diff-5ef33e5331e80d169ca2dea093de2d7105cbd819b9d7e51706bc87d991d4e6e7L330-R407) [[4]](diffhunk://#diff-5ef33e5331e80d169ca2dea093de2d7105cbd819b9d7e51706bc87d991d4e6e7R420-R430)

**Backend and Docker Support:**

* Installed LibreOffice headless packages (`libreoffice-writer-nogui`, `libreoffice-impress-nogui`, `libreoffice-calc-nogui`) in the Docker image to enable server-side document conversion.

**Artifact URL Handling:**

* Updated artifact preview hook to provide direct artifact URLs for binary and large files, ensuring reliable fetching for PDF and Office document previews. [[1]](diffhunk://#diff-51a898d48b8b49417ed1ec45c15bf655afc3627962d1095f027951b8ac5ebb9aL62-R79) [[2]](diffhunk://#diff-51a898d48b8b49417ed1ec45c15bf655afc3627962d1095f027951b8ac5ebb9aL121-R129)

NOTE: LibreOffice is NOT installed by default unless the user sets the flag. When installed, it is used purely as a file conversion service with no modification. A licence notice and link to their source is added.

### DIFF
--- a/tests/unit/gateway/http_sse/routers/test_document_conversion.py
+++ b/tests/unit/gateway/http_sse/routers/test_document_conversion.py
@@ -202,7 +202,7 @@ class TestConvertToPdfEndpoint:
             result = await convert_to_pdf(
                 request=valid_request,
                 user_id="test-user-123",
-                user_config={"tool:artifact:load": True},
+                _={"tool:artifact:load": True},
             )
             
             assert isinstance(result, ConversionResponse)
@@ -221,7 +221,7 @@ class TestConvertToPdfEndpoint:
                 await convert_to_pdf(
                     request=valid_request,
                     user_id="test-user-123",
-                    user_config={"tool:artifact:load": True},
+                    _={"tool:artifact:load": True},
                 )
             
             assert exc_info.value.status_code == status.HTTP_503_SERVICE_UNAVAILABLE
@@ -244,7 +244,7 @@ class TestConvertToPdfEndpoint:
                 await convert_to_pdf(
                     request=unsupported_request,
                     user_id="test-user-123",
-                    user_config={"tool:artifact:load": True},
+                    _={"tool:artifact:load": True},
                 )
             
             assert exc_info.value.status_code == status.HTTP_400_BAD_REQUEST
@@ -263,7 +263,7 @@ class TestConvertToPdfEndpoint:
                 await convert_to_pdf(
                     request=invalid_request,
                     user_id="test-user-123",
-                    user_config={"tool:artifact:load": True},
+                    _={"tool:artifact:load": True},
                 )
             
             assert exc_info.value.status_code == status.HTTP_400_BAD_REQUEST
@@ -278,7 +278,7 @@ class TestConvertToPdfEndpoint:
             result = await convert_to_pdf(
                 request=valid_request,
                 user_id="test-user-123",
-                user_config={"tool:artifact:load": True},
+                _={"tool:artifact:load": True},
             )
             
             assert isinstance(result, ConversionResponse)
@@ -297,7 +297,7 @@ class TestConvertToPdfEndpoint:
                 await convert_to_pdf(
                     request=valid_request,
                     user_id="test-user-123",
-                    user_config={"tool:artifact:load": True},
+                    _={"tool:artifact:load": True},
                 )
             
             assert exc_info.value.status_code == status.HTTP_500_INTERNAL_SERVER_ERROR
@@ -331,7 +331,7 @@ class TestConvertToPdfEndpoint:
                 result = await convert_to_pdf(
                     request=request,
                     user_id="test-user-123",
-                    user_config={"tool:artifact:load": True},
+                    _={"tool:artifact:load": True},
                 )
                 
                 assert result.success is True, f"Failed for format: {filename}"
@@ -347,7 +347,7 @@ class TestConvertToPdfEndpoint:
                 await convert_to_pdf(
                     request=valid_request,
                     user_id="test-user-123",
-                    user_config={"tool:artifact:load": True},
+                    _={"tool:artifact:load": True},
                 )
                 
                 # Verify logging was called with user information
@@ -381,7 +381,7 @@ class TestConvertToPdfEndpointSecurity:
             result = await convert_to_pdf(
                 request=request,
                 user_id="valid-user",
-                user_config={"tool:artifact:load": True},
+                _={"tool:artifact:load": True},
             )
             assert result.success is True
 
@@ -398,7 +398,7 @@ class TestConvertToPdfEndpointSecurity:
             result = await convert_to_pdf(
                 request=request,
                 user_id="test-user",
-                user_config={"tool:artifact:load": True, "other:permission": False},
+                _={"tool:artifact:load": True, "other:permission": False},
             )
             assert result is not None
 
@@ -456,7 +456,7 @@ class TestConversionRequestEdgeCases:
             result = await convert_to_pdf(
                 request=request,
                 user_id="test-user",
-                user_config={"tool:artifact:load": True},
+                _={"tool:artifact:load": True},
             )
             assert result is not None
 
@@ -472,7 +472,7 @@ class TestConversionRequestEdgeCases:
                 await convert_to_pdf(
                     request=request,
                     user_id="test-user",
-                    user_config={"tool:artifact:load": True},
+                    _={"tool:artifact:load": True},
                 )
             
             # Router enforces 5MB limit with HTTP 413
@@ -490,7 +490,7 @@ class TestConversionRequestEdgeCases:
             result = await convert_to_pdf(
                 request=request,
                 user_id="test-user",
-                user_config={"tool:artifact:load": True},
+                _={"tool:artifact:load": True},
             )
             # Should work - service handles path extraction
             assert result is not None
@@ -508,7 +508,7 @@ class TestConversionRequestEdgeCases:
             result = await convert_to_pdf(
                 request=request,
                 user_id="test-user",
-                user_config={"tool:artifact:load": True},
+                _={"tool:artifact:load": True},
             )
             assert result is not None
 
@@ -525,6 +525,6 @@ class TestConversionRequestEdgeCases:
             result = await convert_to_pdf(
                 request=request,
                 user_id="test-user",
-                user_config={"tool:artifact:load": True},
+                _={"tool:artifact:load": True},
             )
             assert result is not None
--- a/tests/unit/gateway/http_sse/services/test_document_conversion_service.py
+++ b/tests/unit/gateway/http_sse/services/test_document_conversion_service.py
@@ -183,47 +183,48 @@ class TestDocumentConversionServiceSizeValidation:
         """Test that files exceeding max size are rejected."""
         # Create data larger than 1MB limit
         large_data = b"x" * (2 * 1024 * 1024)  # 2MB
-        
-        with pytest.raises(RuntimeError) as exc_info:
-            await service.convert_to_pdf(large_data, "large_document.docx")
-        
-        assert "File too large" in str(exc_info.value)
-        assert "1MB" in str(exc_info.value)  # Should mention the limit
+
+        pdf_bytes, error = await service.convert_binary_to_pdf(large_data, "large_document.docx")
+
+        assert pdf_bytes is None
+        assert error is not None
+        assert "File too large" in error or "too large" in error.lower()
 
     @pytest.mark.asyncio
     async def test_file_at_exact_limit_allowed(self, service):
         """Test that files at exactly the limit are allowed (but may fail conversion)."""
         # Create data exactly at limit
         exact_data = b"x" * (1024 * 1024)  # Exactly 1MB
-        
-        # This should not raise ValueError for size, but may fail on conversion
-        # since we don't have actual LibreOffice. We're testing size validation only.
-        with patch.object(service, 'convert_to_pdf', wraps=service.convert_to_pdf) as wrapped:
-            try:
-                await service.convert_to_pdf(exact_data, "exact.docx")
-            except (ValueError, RuntimeError) as e:
-                # Size validation should pass - if it's a ValueError, it shouldn't be about size
-                if isinstance(e, ValueError):
-                    assert "File too large" not in str(e)
+
+        # Mock subprocess so we don't need real LibreOffice
+        with patch('asyncio.create_subprocess_exec') as mock_subprocess:
+            mock_process = AsyncMock()
+            mock_process.communicate = AsyncMock(return_value=(b"", b""))
+            mock_process.returncode = 0
+            mock_subprocess.return_value = mock_process
+
+            pdf_bytes, error = await service.convert_binary_to_pdf(exact_data, "exact.docx")
+            # Size validation passes; conversion may fail without real LibreOffice output
+            # but the error should NOT be about file size
+            if error:
+                assert "File too large" not in error
 
     @pytest.mark.asyncio
     async def test_file_under_limit_allowed(self, service):
         """Test that files under the limit proceed to conversion."""
         small_data = b"x" * (512 * 1024)  # 512KB - under 1MB limit
-        
+
         # Mock the subprocess to test that we get past size validation
         with patch('asyncio.create_subprocess_exec') as mock_subprocess:
             mock_process = AsyncMock()
             mock_process.communicate = AsyncMock(return_value=(b"", b""))
             mock_process.returncode = 0
             mock_subprocess.return_value = mock_process
-            
-            # Should not raise ValueError about size
-            try:
-                await service.convert_to_pdf(small_data, "small.docx")
-            except RuntimeError:
-                # Expected - conversion will fail without actual output file
-                pass
+
+            # Should not return a size-related error
+            pdf_bytes, error = await service.convert_binary_to_pdf(small_data, "small.docx")
+            if error:
+                assert "File too large" not in error
 
 
 class TestDocumentConversionServiceConversion:
@@ -239,30 +240,31 @@ class TestDocumentConversionServiceConversion:
 
     @pytest.mark.asyncio
     async def test_convert_unavailable_service(self):
-        """Test conversion fails when service is not available."""
+        """Test conversion returns error when service is not available."""
         with patch.object(DocumentConversionService, '_find_libreoffice', return_value=None):
             service = DocumentConversionService()
             assert service.is_available is False
-            
-            with pytest.raises(ValueError) as exc_info:
-                await service.convert_to_pdf(b"content", "document.docx")
-            
-            assert "not available" in str(exc_info.value)
+
+            pdf_bytes, error = await service.convert_binary_to_pdf(b"content", "document.docx")
+
+            assert pdf_bytes is None
+            assert error is not None
+            assert "not available" in error
 
     @pytest.mark.asyncio
     async def test_convert_unsupported_format(self, service):
-        """Test conversion fails for unsupported formats."""
-        with pytest.raises(RuntimeError) as exc_info:
-            await service.convert_to_pdf(b"content", "image.png")
-        
-        assert "Unsupported format" in str(exc_info.value)
-        assert "png" in str(exc_info.value)
+        """Test conversion returns error for unsupported formats."""
+        pdf_bytes, error = await service.convert_binary_to_pdf(b"content", "image.png")
+
+        assert pdf_bytes is None
+        assert error is not None
+        assert "Unsupported format" in error or "unsupported" in error.lower()
 
     @pytest.mark.asyncio
     async def test_convert_timeout_handling(self, service):
         """Test that conversion timeout is properly handled."""
         test_data = b"test document content"
-        
+
         with patch('asyncio.create_subprocess_exec') as mock_subprocess:
             mock_process = AsyncMock()
             # Simulate timeout by making communicate raise TimeoutError
@@ -270,11 +272,12 @@ class TestDocumentConversionServiceConversion:
             mock_process.kill = MagicMock()
             mock_process.wait = AsyncMock()
             mock_subprocess.return_value = mock_process
-            
-            with pytest.raises(RuntimeError) as exc_info:
-                await service.convert_to_pdf(test_data, "document.docx")
-            
-            assert "timed out" in str(exc_info.value)
+
+            pdf_bytes, error = await service.convert_binary_to_pdf(test_data, "document.docx")
+
+            assert pdf_bytes is None
+            assert error is not None
+            assert "timed out" in error.lower() or "timeout" in error.lower()
             mock_process.kill.assert_called_once()
             mock_process.wait.assert_awaited_once()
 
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -137,16 +149,30 @@ COPY --from=node-binaries /usr/local/lib/node_modules /usr/local/lib/node_module
...
+        echo "See THIRD_PARTY_LICENSES/LIBREOFFICE.md for full attribution" && \
+        echo "============================================================" && \
+        apt-get install -y --no-install-recommends \
+        libreoffice-writer-nogui \
+        libreoffice-impress-nogui \
+        libreoffice-calc-nogui; \
+    else \
+        echo "Skipping LibreOffice installation (set INSTALL_LIBREOFFICE=true to enable binary artifact preview)"; \
+    fi && \
+    curl -sL https://deb.nodesource.com/setup_20.x | bash - && \
```
efunneko: There is a comment above that says 'no curl - due to vulnerabilities', but we are using curl here. I am not sure exactly what the issue with curl in this context is, but something needs to be fixed (usage or comment)

src/solace_agent_mesh/gateway/http_sse/routers/document_conversion.py
```diff
@@ -0,0 +1,155 @@
...
+            error=None,
+        )
+
+    except HTTPException:
+        raise
+    except Exception as e:
+        log.exception("%s Unexpected error during conversion: %s", log_prefix, e)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail=f"Conversion failed: {str(e)}",
```
efunneko: Claude had an issue with this line possibly leaking internal information

client/webui/frontend/src/lib/components/chat/preview/Renderers/PdfRenderer.tsx
```diff
@@ -0,0 +1,438 @@
...
+// Custom event for snip-to-chat functionality
+export const SNIP_TO_CHAT_EVENT = "snip-to-chat";
+
+export interface SnipToChatEventDetail {
+    file: File;
+    filename: string;
+}
+
+// Configure PDF.js worker
+pdfjs.GlobalWorkerOptions.workerSrc = `https://cdnjs.cloudflare.com/ajax/libs/pdf.js/${pdfjs.version}/pdf.worker.min.mjs`;
```
efunneko: I don't think we typically just reach out to CDNs for this type of thing. Can it not be a normal npm dependency?

src/solace_agent_mesh/gateway/http_sse/routers/document_conversion.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+    service = get_document_conversion_service()
+    return ConversionStatusResponse(
+        available=service.is_available,
+        supported_formats=service.get_supported_extensions(),
+    )
+
+
+@router.post(
+    "/to-pdf",
```
efunneko, line 116: Claude was concerned with the load that this will put on the backend and that a single user shouldn't be able to spawn more than one of these at a time. Do we know what the RAM and CPU costs of a conversion are?

client/webui/frontend/src/lib/components/chat/preview/previewUtils.ts
```diff
@@ -292,11 +355,12 @@ export function decodeBase64Content(content: string): string {
     }
 }
 
-const RENDER_TYPES = ["csv", "html", "json", "mermaid", "image", "markdown", "audio", "text", "yaml"];
-const RENDER_TYPES_WITH_RAW_CONTENT = ["image", "audio"];
+const RENDER_TYPES = ["csv", "html", "json", "mermaid", "image", "markdown", "audio", "text", "yaml", "docx", "pptx", "pdf"];
+const RENDER_TYPES_WITH_RAW_CONTENT = ["image", "audio", "docx", "pptx"];
```
efunneko, line 406: Do we have size limits for preview? If I understand how this works, we will be sending the content to the BE to be converted to a PDF and then previewed as a PDF in the FE. However, I have seen PPTs and DOCs with massive sizes and I fear that both sending it back down to the BE and running the conversion will put huge load on the webui gateway.

client/webui/frontend/src/lib/components/chat/preview/Renderers/OfficeDocumentRenderer.tsx
```diff
@@ -0,0 +1,308 @@
...
+    size(): number {
+        return this.cache.size;
+    }
+}
+
+const pdfConversionCache = new LRUCache(PDF_CACHE_MAX_ENTRIES);
+
+// Improved hash function for cache key using djb2 algorithm
+// Uses more content and includes a proper hash to reduce collision risk
+const hashContent = (content: string, filename: string): string => {
```
efunneko, line 97: Claude had this to say about the cache:

1. LRU cache uses weak DJB2 hash — [OfficeDocumentRenderer.tsx:87-102](vscode-webview://11ah1abm7r0g7u9kr1r6q0f45i894e4siq8m5lpvf3u8a1jq8tm7/client/webui/frontend/src/lib/components/chat/preview/Renderers/OfficeDocumentRenderer.tsx#L87-L102)

The cache key is filename:length:djb2(first_1000_chars). DJB2 is a fast but collision-prone hash. Two different files with the same name, similar length, and shared prefix would collide. Should use crypto.subtle.digest('SHA-256', ...) or at minimum hash more of the content.

2. LRU cache has no TTL and stores data URLs — [OfficeDocumentRenderer.tsx:27-83](vscode-webview://11ah1abm7r0g7u9kr1r6q0f45i894e4siq8m5lpvf3u8a1jq8tm7/client/webui/frontend/src/lib/components/chat/preview/Renderers/OfficeDocumentRenderer.tsx#L27-L83)

10 entries of base64 PDF data URLs (each potentially 5-10MB+) persist in a module-level global with no expiration. That's up to ~100MB permanently in browser memory with no cleanup mechanism.


I am less concerned with #1, but #2 does sound like a potential serious problem. As I mentioned elsewhere, I think we need a fairly low threshold on previewable doc sizes

src/solace_agent_mesh/gateway/http_sse/routers/document_conversion.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+    Converts a document to PDF format.
+
+    The input document should be base64-encoded. The response will contain
+    the converted PDF as a base64-encoded string.
+
+    Supported formats: PPTX, PPT, DOCX, DOC, XLSX, XLS, ODT, ODP, ODS
+    """
+    log_prefix = f"[DocumentConversion] User={user_id} -"
+    log.info("%s Conversion request for: %s", log_prefix, request.filename)
```
efunneko, line 145: Claude had concerns that request.filename has no size limit and could cause problems if left unbounded

src/solace_agent_mesh/gateway/http_sse/routers/document_conversion.py
```diff
@@ -0,0 +1,155 @@
...
+import logging
+from typing import TYPE_CHECKING
+
+from fastapi import APIRouter, Depends, HTTPException, status
+from pydantic import BaseModel, Field
+
+from ..dependencies import get_user_id, ValidatedUserConfig
+from ..services.document_conversion_service import get_document_conversion_service
+
+if TYPE_CHECKING:
```
efunneko: Should remove this

client/webui/frontend/src/lib/components/chat/preview/Renderers/PdfRenderer.tsx
```diff
@@ -0,0 +1,438 @@
...
+                <Document
+                    options={pdfOptions}
+                    file={url}
+                    onLoadSuccess={onDocumentLoadSuccess}
+                    onLoadError={onDocumentLoadError}
+                    loading={<div className="p-4 text-center">Loading PDF...</div>}
+                    error={<div className="p-4 text-center text-red-500">Failed to load PDF.</div>}
+                >
+                    <div ref={documentContainerRef} style={{ transform: `translate(${pan.x}px, ${pan.y}px)` }}>
+                        {numPages &&
```
efunneko: Should we render the entire doc at once? What if the doc was really big - are there any issues with putting it all in the preview?

src/solace_agent_mesh/gateway/http_sse/services/document_conversion_service.py
```diff
@@ -0,0 +1,331 @@
...
+                        "LibreOffice conversion failed (exit code %d): %s",
+                        process.returncode,
+                        error_msg,
+                    )
+                    raise RuntimeError(f"Conversion failed: {error_msg}")
+
+                # Find the output PDF file
+                # LibreOffice may take a moment to write the file, so we retry a few times
+                output_path = None
+                for attempt in range(5):
```
efunneko: Is 5 times with 0.2 seconds delay enough? What tests have we done to ensure this? What if the file is really big or the BE is slow at the time?

src/solace_agent_mesh/gateway/http_sse/routers/document_conversion.py
```diff
@@ -0,0 +1,155 @@
...
+            base64.b64decode(request.content)
+        except Exception as e:
+            log.warning("%s Invalid base64 content: %s", log_prefix, e)
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail="Invalid base64-encoded content",
+            )
+
+        # Perform conversion
+        pdf_base64, error = await service.convert_base64_to_pdf_base64(
```
efunneko: We just did a full decode a few lines above and discarded it. This feels very wasteful

client/webui/frontend/src/lib/components/chat/preview/Renderers/OfficeDocumentRenderer.tsx
```diff
@@ -0,0 +1,308 @@
...
+        };
+
+        if (content) {
+            initializeRenderer();
+        }
+
+        return () => {
+            isMounted = false;
+        };
+    }, [content, filename, checkConversionService, convertToPdf, binaryArtifactPreviewEnabled]);
```
efunneko, line 432: According to Claude, this we will have rerenders happening here which will cause unnecessary BE load. We should verify that that we don't try to do the conversion twice. Even just rerendering a large complicated PDF in the FE would be bad

client/webui/frontend/src/lib/components/chat/preview/Renderers/OfficeDocumentRenderer.tsx
```diff
@@ -0,0 +1,308 @@
...
+            return `data:application/pdf;base64,${data.pdfContent}`;
+        } catch (err) {
+            console.error("Failed to convert document to PDF:", err);
+            throw err;
+        }
+    }, [content, filename]);
+
+    // Main effect to check service and convert
+    useEffect(() => {
+        let isMounted = true;
```
efunneko: From Claude:

The isMounted ref pattern is deprecated in React. Should use AbortController for fetch cancellation — this also handles the missing request timeout issue (no timeout on the fetch calls).

client/webui/frontend/src/lib/components/chat/preview/previewUtils.ts
efunneko: I see that we have a default 5MB limit for preview. How expensive is this to do on the BE? It seems that this would cause quite a workload to upload to BE, convert and download again even for 5MB, but perhaps this isn't too hard?

client/webui/frontend/src/lib/components/chat/preview/previewUtils.ts
```diff
@@ -327,8 +401,10 @@ export const getFileContent = (file: FileAttachment | null) => {
 };
 
 // Configuration constants
-const MAX_ARTIFACT_SIZE = 5 * 1024 * 1024; // configurable limit
+const MAX_ARTIFACT_SIZE = 5 * 1024 * 1024; // configurable limit for content-based renderers
+const MAX_ARTIFACT_SIZE_URL_BASED = 100 * 1024 * 1024; // 100 MB limit for URL-based renderers (PDF)
```
efunneko: I don't really understand what a URL_BASED renderer is. 100MB seems huge to me, though.

Should we consider making these configurable?

THIRD_PARTY_LICENSES/LIBREOFFICE.md
```diff
@@ -0,0 +1,49 @@
+# LibreOffice Third-Party License Notice
```
efunneko, line 1: Do we need to make this visible in the UI? (under About?)

client/webui/frontend/src/lib/components/chat/preview/previewUtils.ts
amir-ghasemi: URL-based is for things like PDF and images that can be streamed efficiently from a URL. Lowered to 50MB.

client/webui/frontend/src/lib/components/chat/preview/Renderers/OfficeDocumentRenderer.tsx
amir-ghasemi: Fixed.

client/webui/frontend/src/lib/components/chat/preview/Renderers/OfficeDocumentRenderer.tsx
amir-ghasemi: Fixed. 3 layers of protection: 

- A ref that tracks the cache key of content being converted
- LRU cache
- Added conversionState ('idle' | 'checking' | 'converting' | 'success' | 'error') to track progress and prevent overlapping conversions.

src/solace_agent_mesh/gateway/http_sse/routers/document_conversion.py
amir-ghasemi: Fixed. Now we decode base64 once and pass binary data directly to the service

src/solace_agent_mesh/gateway/http_sse/services/document_conversion_service.py
amir-ghasemi: improved retry logic with  exponential backoff. The retry logic now has more headroom within that 30s request timeout to find the output file.

client/webui/frontend/src/lib/components/chat/preview/Renderers/PdfRenderer.tsx
amir-ghasemi: Fixed to use the bundled local worker.

client/webui/frontend/src/lib/components/chat/preview/previewUtils.ts
amir-ghasemi: added protection via rate limiting:
- Global: 5 concurrent conversions max (returns 503 if exceeded)
- Per-user: 1 concurrent conversion per session (returns 429)
We can change the limits if we find 5MB is too generous but seems reasonable (~5s for 4MB pptx)

Dockerfile
amir-ghasemi: we use curl during build, the comment was for runtime. Removed to avoid confusion.

src/solace_agent_mesh/gateway/http_sse/routers/document_conversion.py
amir-ghasemi: Fixed. Changed the error message to a generic one that doesn't expose internal details

src/solace_agent_mesh/gateway/http_sse/routers/document_conversion.py
amir-ghasemi: Fixed. Added two-tier rate limiting:

- Per-user limit: 1 concurrent conversion per session (uses session_id from cookie). Returns HTTP 429 if user tries to queue multiple conversions.
- Global limit: 5 concurrent conversions max across all users. Returns HTTP 503 if system is at capacity.

THIRD_PARTY_LICENSES/LIBREOFFICE.md
amir-ghasemi: we have provided attribution and link to source code as well as notice during installation. That seems to be checking off all the boxes but, if we still want more assurance, UI visibility can be added. Would need a consult with UX perhaps?

lgh-solace:
I've been experimenting with using the sam_developer tool alongside the FE best practices doc to review larger PRs. It's not perfect, but interesting to share for iterative feedback/improvement. I feel it does show that some "high" issues in the FE best practices should probably be lower.

---

## Code Review                                                                                                                                 
                  
  Great feature overall — the backend safety story is solid (rate limiting, size limits, auth gating, generic error messages to clients). The  main concerns are frontend best-practice violations and a few backend issues.                                                                  

  ---

  ### 🔴 Very High

  **1. Native `<button>` elements instead of the standard `Button` component**
  `PdfRenderer.tsx` lines 331, 339, 347, 356, 368 — all toolbar controls use `<button>` directly. These must be replaced with `Button` from
  `@SolaceLabs/solace-agent-mesh-ui` (e.g. `variant="ghost" size="icon"`).

  **2. `React.FC<Props>` pattern used in all three new/modified components**
  - `PdfRenderer.tsx:35`
  - `OfficeDocumentRenderer.tsx:146`
  - `ContentRenderer.tsx:17`

  All three should use direct prop typing:
  ```typescript
  // ✅
  export function PdfRenderer({ url, filename }: PdfRendererProps) { ... }
  ```

  **3. Default React import in `ContentRenderer.tsx:1`**
  ```typescript
  import React from "react"; // ❌ — namespace import
  ```
  React 19 doesn't require this for JSX. Remove it or switch to named imports.

  ---

  ### 🟠 High

  **4. Hardcoded Tailwind color utilities throughout `PdfRenderer.tsx`**
  Numerous violations: `bg-gray-100`, `dark:bg-gray-800`, `bg-white/80`, `hover:bg-gray-200`, `bg-blue-100`, `text-blue-600`, `bg-green-100`,
  `text-red-500`, etc. — these break theme support.

  Also in `getSelectionStyle()`, inline styles use hardcoded `#3b82f6` and `rgba(59, 130, 246, 0.1)`. All should use CSS variable palette classes
   (`text-(--color-info-wMain)`, `bg-(--color-info-w10)`, etc.).

  **5. Unversioned package dependency**
  `package.json`: `"react-pdf": "^9.2.1"` — exact versions are required: `"react-pdf": "9.2.1"` (no caret).

  ---

  ### 🟡 Medium

  **6. `setRenderError` called during the render phase (`ContentRenderer.tsx:43`)**
  ```typescript
  setRenderError("URL and filename are required for PDF preview.");
  return null;
  ```
  Calling a state setter directly during render is a React anti-pattern. Move this into a `useEffect` or handle it differently (e.g. render a
  fallback element without mutating parent state).

  **7. Timeout detection is dead code (`OfficeDocumentRenderer.tsx:323`)**
  ```typescript
  if (convError instanceof Error && convError.message.includes("timeout")) {
  ```
  `fetchWithTimeout` cancels via `AbortController`, which throws an `AbortError` (`name === "AbortError"`). That is already caught and returned
  early at lines 316–317, so the `"timeout"` branch here is never reached. Timeout errors silently fall through to the generic message.

  **8. Excessive `console.*` calls left from development**

  `PdfRenderer.tsx`:
  - `console.info("[PdfRenderer] sendToChat called..."` (lines 235–236, 246) — debug noise that should be removed before merge

  `OfficeDocumentRenderer.tsx`:
  - Multiple `console.log(...)` calls (lines 69, 178, 179, 253, 259, 268, 307, 308) — verbose debug output in production

  **9. `documentContainerRef` declared but `.current` is never read (`PdfRenderer.tsx:48`)**
  The ref is assigned to a `<div>` but never accessed. Dead code.

  **10. `getSelectionStyle()` called twice on one render (`PdfRenderer.tsx:403`)**
  ```typescript
  {selection && getSelectionStyle() && <div style={getSelectionStyle()!} />}
  ```
  Compute once into a local variable.

  **11. No Storybook stories for new components**
  `PdfRenderer` and `OfficeDocumentRenderer` are new components with multiple distinct states (loading, error, feature-disabled, success).
  Stories should cover at least: default/loading, conversion error, feature disabled, and successful preview.

  ---

  ### Backend (SAM Standards)

  **12. `_user_conversion_locks` grows unboundedly (`document_conversion.py:31`)**
  ```python
  _user_conversion_locks: dict[str, asyncio.Lock] = defaultdict(asyncio.Lock)
  ```
  Every unique `user_id` adds a permanent entry. In a long-running server with many users this is a memory leak. Consider an LRU dict with a size
   cap or periodic cleanup.

  **13. Dead code in service (`document_conversion_service.py:351–410`)**
  `convert_to_pdf()` and `convert_base64_to_pdf_base64()` are never called — the router uses `convert_binary_to_pdf` directly. Remove them.

  **14. Empty `TYPE_CHECKING` block (`document_conversion.py:19–20`)**
  ```python
  if TYPE_CHECKING:
      pass  # remove this
  ```

  **15. Unused injected dependency (`document_conversion.py:93`)**
  ```python
  user_config: dict = Depends(ValidatedUserConfig(["tool:artifact:load"]))
  ```
  The variable is only used for its auth side-effect. Rename to `_` to signal intentionally unused:
  ```python
  _: dict = Depends(ValidatedUserConfig(["tool:artifact:load"]))
  ```

  **16. Feature flag key naming inconsistency**
  `config.py` uses `binaryArtifactPreview` (camelCase) while all other feature flags (`projects`, `background_tasks`, `auto_title_generation`)
  use snake_case. Align with the existing convention or document the exception.

  ---

  ### Testing Philosophy

  The tests rely heavily on mocking — which is the anti-pattern the SAM testing philosophy explicitly flags. Per SAM standards, a good test
  creates real artifacts, verifies them, and tears them down. A bad test mocks 90% of the environment and asserts `mock.assert_called_once()`.

  Several tests fall into the bad category:

  **`test_convert_to_pdf_logs_user_activity`** — Directly asserts internal logging behaviour with `mock_log.info.assert_called()`. This is the
  textbook example of testing implementation details rather than behaviour. What matters is that the conversion succeeds and returns the correct
  PDF bytes, not how many times a logger is called.

  **`TestConvertToPdfEndpointSecurity`** — Neither security test actually enforces security. They pass valid `user_id` and `user_config` directly
   as function arguments, bypassing FastAPI's dependency injection entirely. These tests don't verify that unauthenticated requests are rejected.

  **`test_file_at_exact_limit_allowed`** — Uses a bare `try/except` to silently swallow expected errors, making it impossible to tell if the test
   passed or the assertion was simply never reached.

  The one test closest to the SAM ideal is `test_convert_successful`, which uses a real temp directory and verifies actual output bytes — more of
   this pattern is needed.

  ---

  ### Exception Handling

  **`document_conversion.py:204–211`** — Exception context is lost when raising `HTTPException`. Per SAM standards, use `raise ... from e` to
  preserve the chain:
  ```python
  # ❌ Current
  raise HTTPException(status_code=500, detail="...")

  # ✅ SAM standard
  raise HTTPException(status_code=500, detail="...") from e
  ```

  **`document_conversion_service.py:408–409`** — `convert_base64_to_pdf_base64` catches a bare `Exception` and returns `str(e)` directly as the
  error message, leaking internal error details to the caller. This contradicts the security hardening done in the router. (Though since this
  method is dead code — see point 13 above — removal resolves both issues.)

  ---

  ### Code Comments

  Several comments describe *what* the code does rather than *why*:
  ```python
  # Run LibreOffice conversion          ← states the obvious
  # --headless: Run without GUI         ← restates the flag name
  # Run conversion asynchronously       ← restates the code
  ```
  The one genuinely good "why" comment is line 241: `# (This is separate from the overall method timeout)` — which explains a non-obvious design
  decision. More comments should follow this pattern.

  ---

  ### Nits

  - `Optional[str]` style in `document_conversion_service.py` — prefer `str | None` (Python 3.10+ style used elsewhere in the codebase)
  - The djb2 hash in `hashContent()` only samples the first 1000 characters — low risk of cache collisions on large files with identical prefixes
   but different bodies
  - The `.locked()` pre-checks before semaphore/lock acquisition (lines 128–137) are inherently racy, but this is fine as a fast-reject
  optimisation — enforcement is correctly done by the `async with` blocks
  - PR title typo: `"Add ability to preview binary artifacts (PDF. PPTX, DOCX)"` — the `.` after PDF should be a `,`

PR #855, opened 2026-01-21
fix(DATAGO-122175): Fix inconsistent character limit behavior when writing to project description

shanechen-solace:

Fix inconsistent character limit behavior when writing to the project description. 

The new behavior is based on that of the existing validation in `EditInstructionsDialog.tsx`.

Added frontend validation errors for text over 1000 characters long in any dialog that can write to the project description. 

Previously the field would cap out at 1000 characters and prevent further typing. This PR changes the behavior to allow typing one extra character past the limit but displays an error and disables the submit button.

### Samples

Screenshot 2026-01-21 at 4 42 17 PM
Screenshot 2026-01-21 at 4 45 04 PM
Screenshot 2026-01-21 at 4 42 10 PM
Screenshot 2026-01-21 at 4 40 56 PM

- [X] Manual testing: Manually tested the new validation errors
- [x] Storybook tests: Updated tests with character limit cases

This is a frontend only bug. Input for these fields should already be validated in the backend.

### DIFF
--- a/client/webui/frontend/src/lib/components/ui/fieldFooter.tsx
+++ b/client/webui/frontend/src/lib/components/ui/fieldFooter.tsx
@@ -13,7 +13,7 @@ const FieldFooter: React.FC<FieldFooterProps> = ({ hasError = false, message, er
     const defaultAlign = hasError ? "left" : "right";
     const textAlign = align || defaultAlign;
 
-    return <div className={cn("text-xs", hasError ? "text-destructive" : "text-muted-foreground", textAlign === "right" && "text-right", className)}>{hasError ? error : message}</div>;
+    return <div className={cn("text-xs", hasError ? "text-(--color-error-wMain)" : "text-muted-foreground", textAlign === "right" && "text-right", className)}>{hasError ? error : message}</div>;
 };
 
 FieldFooter.displayName = "FieldFooter";


client/webui/frontend/src/lib/components/projects/CreateProjectDialog.tsx
```diff
@@ -13,11 +14,15 @@ interface CreateProjectDialogProps {
 }
 
 export const CreateProjectDialog: React.FC<CreateProjectDialogProps> = ({ isOpen, onClose, onSubmit, isSubmitting = false }) => {
+    const { validationLimits } = useConfigContext();
     const [name, setName] = useState("");
     const [description, setDescription] = useState("");
     const [error, setError] = useState<string | null>(null);
     const [nameError, setNameError] = useState<string | null>(null);
 
+    const MAX_DESCRIPTION_LENGTH = validationLimits?.projectDescriptionMax ?? 1000;
```
JKaram: `1000` is a magic number in this case. Can you move it out of the component as a constant like `const DEFAULT_MAX_DESCRIPTION_LENTGH=1000`

client/webui/frontend/src/lib/components/projects/CreateProjectDialog.tsx
```diff
@@ -87,16 +92,27 @@ export const CreateProjectDialog: React.FC<CreateProjectDialogProps> = ({ isOpen
...
                             </label>
-                            <Textarea id="project-description" value={description} onChange={e => setDescription(e.target.value)} disabled={isSubmitting} rows={3} maxLength={1000} />
-                            <div className="text-secondary-foreground text-right text-xs">{description.length}/1000</div>
+                            <Textarea
+                                id="project-description"
+                                value={description}
+                                onChange={e => setDescription(e.target.value)}
+                                disabled={isSubmitting}
+                                rows={3}
+                                maxLength={MAX_DESCRIPTION_LENGTH + 1}
```
JKaram, line 103: I would probably add the `+ 1` to the variable itself. Just keep it as the source of truth.

client/webui/frontend/src/lib/components/projects/ProjectDetailView.tsx
```diff
@@ -31,6 +33,9 @@ export const ProjectDetailView: React.FC<ProjectDetailViewProps> = ({ project, o
     const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
     const [isDeleting, setIsDeleting] = useState(false);
 
+    const MAX_DESCRIPTION_LENGTH = validationLimits?.projectDescriptionMax ?? 1000;
```
JKaram: This would be the reason why why use constants like I mentioned above. You can export  `const DEFAULT_MAX_DESCRIPTION_LENTGH=1000` from a shared util file and use it as the source of truth for both fields.

client/webui/frontend/src/lib/components/projects/ProjectDetailView.tsx
```diff
@@ -189,16 +194,27 @@ export const ProjectDetailView: React.FC<ProjectDetailViewProps> = ({ project, o
...
+                                placeholder="Project description"
+                                rows={4}
+                                disabled={isSaving}
+                                maxLength={MAX_DESCRIPTION_LENGTH + 1}
+                                className={`resize-none text-sm ${isDescriptionOverLimit ? "border-destructive" : ""}`}
+                            />
+                            <div className={`text-xs ${isDescriptionOverLimit ? "text-destructive" : "text-muted-foreground text-right"}`}>
+                                {isDescriptionOverLimit && `Description must be less than ${MAX_DESCRIPTION_LENGTH} characters`}
+                                {!isDescriptionOverLimit && `${editedDescription.length} / ${MAX_DESCRIPTION_LENGTH}`}
+                            </div>
```
JKaram: I would say since we are using the same logic in both components its worth creating a `<DescriptionTextArea>` component that can be reused in both spots.

shanechen-solace: ~~I've extracted these two components into a new component <ValidatedTextareaWithFooter>. This is a general component that can be used for any textareas that will require validation.~~

I've extracted the footer into its own component.

client/webui/frontend/src/lib/components/ui/validated-textarea-with-footer.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
```
lgh-solace: If we go with this, could you rename this file with camelCase? I realize we have a mix of naming conventions at the moment, but I think we want to consolidate on camelCase.

I'm not 100% sure that we want to go with this pattern...  just because we have many different kinds of input that may show a potential hint and error message and need to be consistent.

Maybe we could add the below tiny common components to community? And then could have a TextArea component that combined them, but we could use them separately as well? Thoughts?
```
export const InputHint = ({ children }: { children: React.ReactNode }) => {
    return <span className="text-xs text-muted-foreground ">{children}</span>;
};

export const InputError = ({ children }: { children: React.ReactNode }) => {
    return <span className="text-xs text-(--color-error-wMain)">{children}</span>;
};

export const InputCharacterCount = ({ currentLength, maxLength }: { currentLength?: number; maxLength: number }) => {
    return <div className="text-xs text-muted-foreground text-right">{`${currentLength || 0}/${maxLength}`}</div>;
};
```

shanechen-solace: @JKaram and I had a discussion about the organization of input fields and their related components. One solution (similar to your proposal) is to have components that compose the main component (an example of this in the codebase is with dialogs, at `client/webui/frontend/src/lib/components/ui/dialog.tsx`.

While searching the codebase, I discovered an unused file `client/webui/frontend/src/lib/components/ui/form.tsx` that seems to have the behavior we want. I can build off of this in a future iteration, but for now this change seems too large for the scope of the PR. 

For now, I've reverted the use of the combined pattern and I'd be happy to refactor the field-related dialogs in a later PR.

client/webui/frontend/src/lib/components/ui/fieldFooter.tsx
```diff
@@ -0,0 +1,22 @@
...
+    error: string;
+    align?: "left" | "right";
+    className?: string;
+}
+
+const FieldFooter: React.FC<FieldFooterProps> = ({ hasError = false, message, error, align, className }) => {
+    const defaultAlign = hasError ? "left" : "right";
+    const textAlign = align || defaultAlign;
+
+    return <div className={cn("text-xs", hasError ? "text-destructive" : "text-muted-foreground", textAlign === "right" && "text-right", className)}>{hasError ? error : message}</div>;
```
lgh-solace: Unfortunately, text-destructive is not correct for dark mode. I am hoping to fix the theme palette soon, but for now, can you change this to `text-(--color-error-wMain)` which should look ok in both light and dark modes. Thanks!

And, I can see this will need some refactoring to more generally useful for description field hints, but I agree we an do it later.

PR #859, opened 2026-01-22
chore(DATAGO-120293): Update a2a proxy agent authentication scheme extraction logic

ziyanwan:
This pull request improves how authentication is handled for A2A agent proxies, especially in relation to agent cards that define security schemes. The changes ensure that authentication credentials are stored and applied in a way that is consistent with the agent card's expectations, supporting both compliant (with security_schemes) and legacy/non-compliant agents (without security_schemes). The update also clarifies documentation and improves fallback logic for security scheme names.

**Authentication and Credential Handling Improvements:**

* Introduced the `_extract_security_scheme_name` and `_get_default_scheme_name` methods in `component.py` to reliably determine the correct security scheme name from the agent card for storing and retrieving credentials, with sensible fallbacks for backward compatibility.
* Updated logic in `_get_or_create_a2a_client` to only store credentials in the credential store if the agent card defines security_schemes; otherwise, authentication headers are applied directly via the httpx client. This applies to all authentication types (static_bearer, static_apikey, oauth2 flows). [[1]](diffhunk://#diff-36b1f700b0376360da3be36bde478f193fa65a59d64d1bcb7aacfc2c3fd4d06bL835-R979) [[2]](diffhunk://#diff-36b1f700b0376360da3be36bde478f193fa65a59d64d1bcb7aacfc2c3fd4d06bR1042-R1050) [[3]](diffhunk://#diff-36b1f700b0376360da3be36bde478f193fa65a59d64d1bcb7aacfc2c3fd4d06bR1059-R1084) [[4]](diffhunk://#diff-36b1f700b0376360da3be36bde478f193fa65a59d64d1bcb7aacfc2c3fd4d06bL938-R1096) [[5]](diffhunk://#diff-36b1f700b0376360da3be36bde478f193fa65a59d64d1bcb7aacfc2c3fd4d06bL957-R1129)
* Modified client creation to add the AuthInterceptor only if security_schemes are present in the agent card, ensuring correct middleware behavior.

**Documentation and Code Comments:**

* Clarified and expanded documentation in `_build_headers` and the `A2AProxiedAgentConfig` model to explain how custom headers interact with authentication and the AuthInterceptor, including the distinction between httpx client-level and middleware-level headers. [[1]](diffhunk://#diff-36b1f700b0376360da3be36bde478f193fa65a59d64d1bcb7aacfc2c3fd4d06bR265-R280) [[2]](diffhunk://#diff-fec6a5965863d1cacae0651bea19400ba98f68f3c9a2a5ba2de09166b160f9a8L206-R210)

### DIFF
--- a/client/webui/frontend/package-lock.json
+++ b/client/webui/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@SolaceLabs/solace-agent-mesh-ui",
-    "version": "1.37.0",
+    "version": "1.38.1",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@SolaceLabs/solace-agent-mesh-ui",
-            "version": "1.37.0",
+            "version": "1.38.1",
             "license": "Apache-2.0",
             "dependencies": {
                 "@hookform/resolvers": "^5.2.2",
--- a/client/webui/frontend/package.json
+++ b/client/webui/frontend/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@SolaceLabs/solace-agent-mesh-ui",
-    "version": "1.37.0",
+    "version": "1.38.1",
     "description": "Solace Agent Mesh UI components - React library for building agent communication interfaces",
     "author": "SolaceLabs <solacelabs@solace.com>",
     "license": "Apache-2.0",
--- a/client/webui/frontend/src/auth/authCallback.tsx
+++ b/client/webui/frontend/src/auth/authCallback.tsx
@@ -5,8 +5,12 @@ function AuthCallback() {
         const hash = window.location.hash.substring(1);
         const params = new URLSearchParams(hash);
         const accessToken = params.get("access_token");
+        const samAccessToken = params.get("sam_access_token");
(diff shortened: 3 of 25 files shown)


src/solace_agent_mesh/agent/proxies/a2a/component.py
```diff
@@ -902,29 +1043,50 @@ async def _get_or_create_a2a_client(
...
+                    # Only store credentials in credential store if agent card has security_schemes
+                    # (AuthInterceptor requires security_schemes to work)
+                    # If no security_schemes, auth is applied directly via httpx headers
+                    if has_security_schemes:
+                        scheme_name = self._extract_security_scheme_name(
+                            agent_card, auth_type, agent_name
+                        )
+                        await self._credential_store.set_credentials(
+                            session_id, scheme_name, access_token
+                        )
```
enavitan, line 1108: this bit is used 4 times, i'd extract it in a dedicated method:

```
_ensure_credentials(...): 
...
```

the usage then would turn into:
```
if has_security_scheme:
   await _ensure_credentials(agent_card, auth_config, session_id, token)
```

PR #862, opened 2026-01-22
feat(DATAGO-121398): Add custom health check support for agents and gateways

gregmeldrum:

Enables users to define custom Python functions for health checks, allowing application-specific health validation logic. This supports scenarios like checking external service connectivity, verifying model availability, or custom business logic checks.

- Added `custom_startup_check` and `custom_ready_check` configuration options in the `health_check` YAML section
- Implemented dynamic module loading via `importlib.import_module()` for custom check functions
- Custom functions receive the component instance for context access
- Callable is cached after first load to avoid reimporting on every health check
- Format: `module.path:function_name` (e.g., `my_agent.health:check_ready`)

### Key Design Decisions

- Custom functions must accept a single `component` parameter and return a boolean
- Non-boolean return values are coerced to boolean with a warning
- Exceptions in custom checks are logged and treated as unhealthy (returns False)
- Loading failures are cached to avoid repeated import attempts

- [x] Unit tests: 12 new tests in `test_app_base.py` covering custom health check scenarios
- [x] Tests cover: successful checks, failed checks, exceptions, invalid paths, non-callable attributes, caching behavior
- [x] Manual testing with custom health check module

This is part 3 of a stacked PR series for health checks:
- Part 1: Broker connection health checks
- Part 2: Database connectivity checks
- **Part 3 (this PR)**: Custom health check support

Example configuration:
```yaml
health_check:
  custom_startup_check: my_agent.health:check_startup
  custom_ready_check: my_agent.health:check_ready
```

### DIFF
--- a/src/solace_agent_mesh/common/app_base.py
+++ b/src/solace_agent_mesh/common/app_base.py
@@ -233,32 +233,6 @@ class SamAppBase(App):
 
         return True
 
-    def _get_primary_component(self):
-        """
-        Get the primary component for passing to custom health checks.
-
-        Returns the first component found that has a get_config method,
-        which is typically the main agent, gateway, or platform component.
-
-        Returns:
-            Component instance or None if no suitable component found.
-        """
-        if not hasattr(self, "flows") or not self.flows:
-            return None
-
-        for flow in self.flows:
-            if not hasattr(flow, "component_groups") or not flow.component_groups:
-                continue
-
-            for group in flow.component_groups:
-                for wrapper in group:
-                    component = getattr(wrapper, "component", wrapper)
-                    # Look for components with get_config (main SAM components)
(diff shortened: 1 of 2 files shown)


src/solace_agent_mesh/common/app_base.py
```diff
@@ -225,23 +230,159 @@ def _is_database_connected(
...
+            component = self._get_primary_component()
+            result = func(component)
+
+            if not isinstance(result, bool):
+                log.warning(
+                    "Custom health check '%s' returned non-boolean value: %s",
+                    check_path,
+                    result,
+                )
+                result = bool(result)
```
alimosaed: if the results tis non-boolean, this code can result wrong value! Suppose the result is None or a string other than True or False.

gregmeldrum: Set to be False

PR #873, opened 2026-01-24
feat(DATAGO-122903): Add support for Bedrock model in init GUI

TamimiGitHub:
## Summary
Add AWS Bedrock support for the SAM init GUI. The goal is to have a user friendly UI way to configure AWS Bedrock

Changes include

### Frontend
- Added AWS Bedrock to drop down menu
- Added prefix mapping for bedrock
- Added state variables for AWS Bedrock fields
- Bypass bedrock endpoint testing if bedrock is chosen
- Updated conditional rendering to hide standard fields when AWS Bedrock is selected

### Backend
- Added AWS Bedrock environment variables to `.env`
- Created a new `shared_config_bedrock.yaml` that handles this change

The changes allow for configuring bedrock from the GUI init interface

Screenshot 2026-01-23 at 4 38 45 PM
Screenshot 2026-01-23 at 4 38 37 PM

### DIFF
--- a/cli/commands/init_cmd/web_init_step.py
+++ b/cli/commands/init_cmd/web_init_step.py
@@ -81,12 +81,10 @@ def perform_web_init(current_cli_params: dict) -> dict:
 
             # Check if AWS Bedrock provider is being used
             llm_provider = config_from_portal.get("llm_provider", "")
-            
-            if llm_provider == "aws_bedrock":
-                # For AWS Bedrock, we don't need planning and general model names
-                # The model info is already in llm_model_name and aws_model_id
-                pass
-            else:
+
+            # For AWS Bedrock, we don't need planning and general model names
+            # The model info is already in llm_model_name and aws_model_id
+            if llm_provider != "aws_bedrock":
                 # Handle planning and general model names with fallback to single model name
                 config_from_portal["llm_service_planning_model_name"] = (
                     config_from_portal.get("llm_service_planning_model_name")


templates/shared_config_bedrock.yaml
```diff
@@ -0,0 +1,104 @@
+shared_config:
```
RohanSreelesh: Could you change this approach so that we don't have to add a duplicate `shared_config` template. I would recommend following the pattern which exists for artifact service where the injected lines in shared config are dependent on the chosen provider.

TamimiGitHub: Sounds good! I have a poc for this im just testing it out

cli/commands/init_cmd/web_init_step.py
```diff
@@ -79,22 +79,30 @@ def perform_web_init(current_cli_params: dict) -> dict:
...
-                or config_from_portal.get("llm_model_name")
-            )
+            # Check if AWS Bedrock provider is being used
+            llm_provider = config_from_portal.get("llm_provider", "")
+            
+            if llm_provider == "aws_bedrock":
+                # For AWS Bedrock, we don't need planning and general model names
+                # The model info is already in llm_model_name and aws_model_id
+                pass
+            else:
```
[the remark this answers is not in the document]
RohanSreelesh: @TamimiGitHub this suggestion makes sense

RohanSreelesh (approved):
LG can you just look at the copilot comment, I think you can accept that one. Also would be nice to have validation for the bedrock creds but we can address that in a follow up.

PR #903, opened 2026-01-28
chore(DATAGO-122069): Rename workflow config field from 'agent_name' to 'name'

solace-pjones:

Rename the `agent_name` field to `name` in `WorkflowAppConfig` for better clarity in the workflow context. The `agent_name` field was confusing since workflows are distinct from agents—this change makes the API clearer by using `name` as the primary identifier for workflow instances. The parent class's `agent_name` is automatically populated from `name` via a Pydantic validator for internal compatibility.

  - Added `name: str` field as the primary workflow identifier
  - Made `agent_name` optional and excluded from serialization
  - Added `@model_validator` to auto-populate `agent_name` from `name` for backward compatibility with parent `SamAgentAppConfig`

1. **Auto-populate `agent_name` from `name`**: Rather than breaking the parent class's interface, we use a Pydantic `@model_validator` to automatically populate `agent_name` from `name`. This maintains internal compatibility with `SamAgentAppConfig` while providing a cleaner external API.

2. **Exclude `agent_name` from serialization**: Since `agent_name` is derived and only exists for parent class compatibility, we set `exclude=True` in the Field definition to keep the public API clean.

3. **Required `name` field**: We made `name` required (no default value) to ensure workflows always have a clear identifier. This is stricter than `agent_name` but aligns with the intended use case.

4. **No backward compatibility concerns**: This is a new feature (workflows are new), so there's no legacy code to worry about breaking. The auto-population of `agent_name` is purely for internal parent class compatibility, not external API compatibility.

- [x] Manual testing: Verified all changes compile and configuration loads correctly
- [x] Unit tests: Created 8 new tests in `test_workflow_app_config.py` covering:
  - `name` field is required and accepted
  - `agent_name` is auto-populated from `name`
  - Workflow definition validation works correctly
  - Timeout configurations (custom and defaults) work as expected
  - Dict-based model creation via `model_validate()`
- [x] Regression testing: All 347 existing workflow unit tests pass without modification
- [x] Integration tests: Blocked by missing optional dependencies (`fastmcp`, `testcontainers`); would run if available but not critical for this change
  - The change is backward compatible at the implementation level (agent_name auto-populates) but represents a breaking change in the public API (users must now use `name` instead of `agent_name`)

1. **Breaking API change**: Users migrating from `agent_name` to `name` will need to update their workflow configuration YAML/dicts. The test fixtures in `conftest.py` have all been updated as an example.

### DIFF
--- a/src/solace_agent_mesh/workflow/app.py
+++ b/src/solace_agent_mesh/workflow/app.py
@@ -537,6 +537,21 @@ class WorkflowDefinition(BaseModel):
 class WorkflowAppConfig(SamAgentAppConfig):
     """Workflow app configuration extends agent config."""
 
+
+    # Rename agent_name to name for clarity in workflow context
+    name: str = Field(..., description="Unique name for this workflow instance.")
+
+    # Make parent's agent_name optional and populate from name via validator
+    agent_name: Optional[str] = Field(default=None, exclude=True)
+
+    @model_validator(mode='after')
+    def populate_agent_name_from_name(self):
+        """Populate agent_name from name for compatibility with parent class."""
+        if not self.agent_name:
+            self.agent_name = self.name
+        return self
+
+
     # Workflow definition
     workflow: WorkflowDefinition = Field(..., description="The workflow DAG definition")
 
@@ -594,7 +609,7 @@ class WorkflowApp(App):
 
         # Extract workflow-specific settings
         namespace = app_config.namespace
-        workflow_name = app_config.agent_name
+        workflow_name = app_config.name
 
         # Auto-populate agent card with workflow schemas in skills
         # Note: AgentCardConfig doesn't have input_schema/output_schema directly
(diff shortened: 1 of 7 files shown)


src/solace_agent_mesh/workflow/app.py
```diff
@@ -537,6 +537,21 @@ def _validate_branch_dependency(
...
 
+
+    # Rename agent_name to name for clarity in workflow context
+    name: str = Field(..., description="Unique name for this workflow instance.")
+
+    # Make parent's agent_name optional and populate from name via validator
+    agent_name: Optional[str] = Field(default=None, exclude=True)
+
+    @model_validator(mode='after')
+    def populate_agent_name_from_name(self):
```
efunneko, line 548: I don't see a reason to do this. We should just make sure that we have fixed all places that used to access agent_name

src/solace_agent_mesh/workflow/component.py
efunneko: I think this should access 'name' instead of 'agent_name' here

efunneko (changes requested):
I think there are a couple of small fixes needed

PR #906, opened 2026-01-28
fix(DATAGO-123235): Address issues with typing "/" in chat

amir-ghasemi:
This pull request improves the behavior of the prompt command ("/" trigger) in the chat input area, making the "/" command popover more intuitive and easier to dismiss. The most important changes are grouped below:

**Prompt Command Triggering and Dismissal Logic:**

* The prompt popover now only appears when "/" is typed as the very first character in the input, rather than after any space or newline. This makes the trigger behavior more predictable and less likely to appear unintentionally.
* The prompt popover is now dismissed if the input no longer starts with "/", ensuring the popover only stays open when appropriate.

**Popover Dismissal on Backspace:**

* Added a new `onBackspaceClose` prop to the `PromptsCommand` component, which is called when the user presses backspace with an empty prompt search (i.e., after typing "/"). This handler removes the "/" from the input and closes the popover, streamlining the UX for dismissing the prompt command. [[1]](diffhunk://#diff-bc94554d620a00e74c8fd7859ccbc086441ac7772c4a3dae72779aa5cc5de79dR39-R42) [[2]](diffhunk://#diff-bc94554d620a00e74c8fd7859ccbc086441ac7772c4a3dae72779aa5cc5de79dR207-R217) [[3]](diffhunk://#diff-8d133525d515e1506931bca68a08adc5460f6fadd75d45a64045faecb5f31936R905-R912)

### DIFF
--- a/client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
+++ b/client/webui/frontend/src/lib/components/chat/ChatInputArea.tsx
@@ -527,14 +527,13 @@ export const ChatInputArea: React.FC<{ agents: AgentCardInfo[]; scrollToBottom?:
 
         const cursorPosition = getCursorPosition();
         const textBeforeCursor = value.substring(0, cursorPosition);
-        const lastChar = textBeforeCursor[textBeforeCursor.length - 1];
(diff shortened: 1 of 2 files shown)


DevSol-UQ (approved):
LGTM

PR #908, opened 2026-01-28
chore (DATAGO-122351): Error in calling a workflow directly (remove from chat dropdown)

solace-pjones:
What is the purpose of this change?
Workflows are now registered as agents in the system (to enable their execution as sub-workflows), but they should not appear in the chat interface's agent selection dropdown. This change filters out workflow agents from the dropdown while keeping them available for other parts of the application that need them (like activity visualization).

How was this change implemented?
Files Modified:

Added a .filter() in the agent dropdown's render logic that checks for the workflow agent type extension (ext.params?.type === "workflow") and excludes those agents from the displayed options.

Key Design Decisions
UI-level filtering: Chose to filter at the UI layer (ChatInputArea component) rather than the backend API. This is a safer, less risky change that only affects the chat interface without modifying the agent discovery endpoint, which is used by other parts of the application.

Verifies OrchestratorAgent is visible in the dropdown
Confirms MockWorkflow agent does NOT appear in the HTML (proving filtering works)
All 103 Storybook tests pass (31 test files)
 Manual testing: Verified the filtering logic by adding mock workflow data and observing that it's correctly excluded from the dropdown

Known limitations:

### DIFF
--- a/client/webui/frontend/src/lib/hooks/useAgentCards.ts
+++ b/client/webui/frontend/src/lib/hooks/useAgentCards.ts
@@ -6,6 +6,7 @@ import type { AgentCard, AgentExtension, AgentCardInfo, AgentSkill } from "@/lib
 const DISPLAY_NAME_EXTENSION_URI = "https://solace.com/a2a/extensions/display-name";
 const PEER_AGENT_TOPOLOGY_EXTENSION_URI = "https://solace.com/a2a/extensions/peer-agent-topology";
 const TOOL_EXTENSION_URI = "https://solace.com/a2a/extensions/sam/tools";
+const AGENT_TYPE_EXTENSION_URI = "https://solace.com/a2a/extensions/agent-type";
 
 /**
  * Transforms a raw A2A AgentCard into a UI-friendly AgentCardInfo object,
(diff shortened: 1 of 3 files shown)


client/webui/frontend/src/lib/hooks/useAgentCards.ts
```diff
@@ -31,6 +33,11 @@ export const transformAgentCard = (card: AgentCard): AgentCardInfo => {
         if (toolsExtension?.params?.tools) {
             tools = toolsExtension.params.tools as AgentSkill[];
         }
+
+        const agentTypeExtension = card.capabilities.extensions.find((ext: AgentExtension) => ext.uri === AGENT_TYPE_EXTENSION_URI);
+        if (agentTypeExtension?.params?.type === "workflow") {
```
lgh-solace, line 38: Thanks for making the change!

I was having second thoughts, thinking perhaps having  a `type` instead of `isWorkflow` might be more flexible. But, we can wait until there's another type.

We should also use this to replace `isWorkflowAgent` that is working to figure out what goes in each tab in the agent mesh view. But, that can happen in another PR.

PR #910, opened 2026-01-28
fix(DATAGO-122883): Add total file upload limit for projects

shanechen-solace:

Add a limit to the total size of files that can be uploaded to a project to prevent unbounded storage usage across users.

Added `DEFAULT_MAX_TOTAL_UPLOAD_SIZE_BYTES` constant and validation logic in project_service.py that checks the cumulative size of user-uploaded files before accepting new uploads, rejecting requests with a 400 error when the total would exceed the limit. 

#### For now, the limit per-file has been set to *50mb* and the total upload limit is *100mb*.

The total project size limit is the primary constraint rather than enforcing strict per-file limits, providing flexibility for users to manage their file sizes as needed within the overall quota. LLM-generated artifacts are intentionally excluded from the limit (filtered by `source="project"`) to ensure that AI-generated content doesn't consume a user's upload quota, only their manually uploaded files count.

- [X] Manual testing: Manually added files of different sizes to projects to test the upload limit behavior.
- [X] Integration tests: Added a comprehensive test suite covering limit enforcement, valid uploads, ZIP imports, edge cases, and verification that only user-uploaded files count toward the limit.
- [X] Known limitations: Some test scenarios involving file deletion and replacement cannot be fully validated with the mock artifact service used in tests, but work correctly in production with a real artifact service.

### Demos
SLACK: https://solacedotcom.slack.com/archives/C09TE7FNFAR/p1770050818570359?thread_ts=1769705717.347159&cid=C09TE7FNFAR

https://github.com/user-attachments/assets/d02b7a95-4b3d-49d5-aefd-79b29a52c7d8

The test suite is about half of this PR.

### DIFF
--- a/client/webui/frontend/src/lib/components/projects/KnowledgeSection.tsx
+++ b/client/webui/frontend/src/lib/components/projects/KnowledgeSection.tsx
@@ -56,19 +56,16 @@ export const KnowledgeSection: React.FC<KnowledgeSectionProps> = ({ project }) =
 
     const handleValidateFileSizes = useCallback(
         (files: FileList) => {
-            // Validate per-file sizes
             const fileSizeResult = validateFileSizes(files, { maxSizeBytes: maxPerFileUploadSizeBytes });
             if (!fileSizeResult.valid) {
                 return fileSizeResult;
             }
 
-            // Validate batch upload size
             const batchSizeResult = validateBatchUploadSize(files, maxBatchUploadSizeBytes);
             if (!batchSizeResult.valid) {
                 return batchSizeResult;
             }
 
-            // Validate project total size
             const projectSizeLimitResult = validateProjectSizeLimit(currentProjectArtifactSizeBytes, files, maxProjectSizeBytes);
             if (!projectSizeLimitResult.valid) {
                 return { valid: false, error: projectSizeLimitResult.error };
--- a/src/solace_agent_mesh/gateway/http_sse/services/project_service.py
+++ b/src/solace_agent_mesh/gateway/http_sse/services/project_service.py
@@ -71,9 +71,11 @@ class ProjectService:
             service_class = MiddlewareRegistry.get_resource_sharing_service()
             self._resource_sharing_service = service_class()
 
+        # Get config values, with fallback to default (ensure values are integers for proper formatting)
         max_per_file_upload_config = (
             component.get_config("gateway_max_upload_size_bytes", DEFAULT_MAX_PER_FILE_UPLOAD_SIZE_BYTES)
             if component else DEFAULT_MAX_PER_FILE_UPLOAD_SIZE_BYTES
+        )
         self.max_per_file_upload_size_bytes = int(max_per_file_upload_config) if isinstance(max_per_file_upload_config, (int, float)) else DEFAULT_MAX_PER_FILE_UPLOAD_SIZE_BYTES
 
         max_batch_upload_config = (


client/webui/frontend/src/lib/utils/file-validation.ts
```diff
@@ -56,25 +64,25 @@ export function validateFileSizes(files: FileList | File[], options: FileSizeVal
     }
 
     // Build error message
-    const maxSizeMB = (maxSizeBytes / (1024 * 1024)).toFixed(0);
+    const maxSizeWithUnit = formatFileSize(maxSizeBytes, 0);
```
lgh-solace: Unrelated to your change, but would you mind swapping out the `formatFileSize` utility in this file with `formatBytes` from `format.ts` and deleting `formatFileSize`... just to get rid of some existing duplication?

client/webui/frontend/src/lib/components/projects/KnowledgeSection.tsx
```diff
@@ -47,13 +48,30 @@ export const KnowledgeSection: React.FC<KnowledgeSectionProps> = ({ project }) =
         });
     }, [artifacts]);
 
-    // Validate file sizes before showing upload dialog
-    // if maxUploadSizeBytes is not configured, validation is skipped and backend handles it
+    // Calculate current project size from artifacts
+    const currentProjectSize = React.useMemo(() => {
```
lgh-solace: Maybe `currentProjectArtifactSize`?

src/solace_agent_mesh/gateway/http_sse/utils/helpers.py
```diff
@@ -0,0 +1,35 @@
...
+    Args:
+        size_bytes: Size in bytes
+
+    Returns:
+        Size in megabytes (MB)
+    """
+    return size_bytes / (1024 * 1024)
+
+
+def sanitize_log_input(value: str) -> str:
```
lgh-solace: I feel this function is currently unnecessary? We are logging user_id and project_id, but those aren't user controlled, they're just IDs and UUIDs.

src/solace_agent_mesh/gateway/http_sse/utils/helpers.py
```diff
@@ -0,0 +1,35 @@
+"""
+Common helper utilities for the HTTP SSE Gateway.
+"""
+
+
+def bytes_to_mb(size_bytes: int) -> float:
```
lgh-solace: This one could be useful - but not sure it's useful enough to add to the shared helper since its only one line... 🤔 

Edit: Maybe we could keep it in `project_service.py`?

lgh-solace (commented):
Looks clean and has some nice additions, but perhaps goes too far on the BE? Thoughts?

client/webui/frontend/src/lib/components/projects/KnowledgeSection.tsx
JKaram: Agree - this could also be a good util or hook.

client/webui/frontend/src/lib/components/projects/KnowledgeSection.tsx
```diff
@@ -47,13 +48,30 @@ export const KnowledgeSection: React.FC<KnowledgeSectionProps> = ({ project }) =
...
+                return fileSizeResult;
+            }
+
+            // 2. Validate total upload limit
+            const totalUploadSizeResult = validateTotalUploadSize(currentProjectSize, files, maxTotalUploadSizeBytes);
+            if (!totalUploadSizeResult.valid) {
+                return { valid: false, error: totalUploadSizeResult.error };
+            }
+
+            return { valid: true };
```
JKaram: Can you remove the unnecessary comments. ` // 1. Validate individual file sizes ...`

client/webui/frontend/src/lib/utils/file-validation.ts
```diff
@@ -123,7 +131,43 @@ export function isFileSizeValid(file: File, maxSizeBytes?: number): boolean {
...
+ * This enforces a project-level storage limit, not a per-request limit.
+ *
+ * @param currentProjectSizeBytes - Current total size of project artifacts in bytes
+ * @param newFiles - FileList or array of Files to be uploaded
+ * @param maxTotalUploadSizeBytes - Maximum total upload size limit per project
+ * @returns Validation result with error message if limit would be exceeded
+ */
+export function validateTotalUploadSize(currentProjectSizeBytes: number, newFiles: FileList | File[], maxTotalUploadSizeBytes?: number): TotalUploadSizeValidationResult {
+    const fileArray = Array.from(newFiles);
+    const newSize = fileArray.reduce((sum, file) => sum + file.size, 0);
```
JKaram: This would be a good place for the util I mentioned above. We might even have one already.

PR #917, opened 2026-01-29
chore(DATAGO-115145): Project sharing dialog

JKaram:
This pull request introduces a new user typeahead component for searching and selecting users by email, and adds a complete API layer for project sharing and people search functionality. It also integrates these new APIs with React Query for efficient data fetching and caching. The changes are grouped into two main themes: project sharing APIs/hooks and user search/typeahead.

***This PR is only the Dialog implementation. It is not available in the UI yet***

**Project Sharing APIs and Hooks**

* Added new API endpoints in `service.ts` for project sharing: fetching, creating, deleting, and updating project shares. These endpoints handle requests for managing which users have access to a project.
* Introduced corresponding React Query hooks in `hooks.ts` for project sharing operations, enabling UI components to easily fetch and update project share data.
* Extended `projectKeys` in `keys.ts` to include a `shares` key for cache management of project share queries.

**User Search and Typeahead**

* Added a new `UserTypeahead` component that provides an inline, popover-based user search and selection UI, including keyboard navigation, debounced search, and exclusion of selected emails.
* Implemented the people search API and React Query hooks (`service.ts`, `hooks.ts`, `keys.ts`, `index.ts`) to support user lookup by email, with caching and debouncing for efficient searching. [[1]](diffhunk://#diff-629f1d630d8a00b21f50ffaa10ae67b1c2fd72e53bcef877b1ebdbbc58d8b867R1-R7) [[2]](diffhunk://#diff-e4e8f8902e7dbe9e6ccb7a827ffc99bd8310cc64b9ab8827d4efc2399b9f6beeR1-R12) [[3]](diffhunk://#diff-2c9ae44db9c35ad8eb89a91fc51f2dbfc10ec716b3aa81bd36d9601e6642339dR1-R8) [[4]](diffhunk://#diff-0c47bbd96b33f23d78275ee9f1ba9937ae7eccae8b39e00a0b390f294e6f9ebaR1-R2)
* Exported the new `UserTypeahead` from the common components index and added utility styles for the project share UI. [[1]](diffhunk://#diff-9206da0d5f31d446f461e003292dc17e8117d20dfa98998884bdc87a62f7bae3R14) [[2]](diffhunk://#diff-5e2035ced5ee43b3c22917d14209085fc662fd947bbaecbcbd1fa888d93d1feeR1-R15)

Additionally, the main API index was updated to export the new people API.

**With Identity Service - Uses typeahead component**

https://github.com/user-attachments/assets/3677d93e-443f-4214-a2b6-9bc6d98f97c7

**Without Identity Service - Uses manual email input**

https://github.com/user-attachments/assets/af87fd04-06e7-47f0-a247-f7b3e201cb80


client/webui/frontend/src/lib/contexts/ConfigContext.ts
```diff
@@ -65,6 +65,12 @@ export interface ConfigContextValue {
      * Requires persistence to be enabled.
      */
     autoTitleGenerationEnabled?: boolean;
+
+    /**
+     * Whether Identity Service is configured.
+     * When null, Identity Service is not configured.
+     */
+    identityServiceType: string | null;
```
JKaram, line 73: `/config` api now sends the identity_service_type https://github.com/SolaceLabs/solace-agent-mesh/pull/911

client/webui/frontend/src/lib/api/projects/service.ts
```diff
@@ -79,3 +79,38 @@ export const importProject = async (file: File, options: { preserveName: boolean
     const result = await api.webui.post("/api/v1/projects/import", formData);
     return result;
 };
+
+// Project Sharing APIs
+
+export const getProjectShares = async (projectId: string) => {
+    const response = await api.webui.get<ProjectSharesResponse>(`/api/v1/projects/${projectId}/shares`);
+    return response;
```
lgh-solace: Minor, but it feels unnecessary to assign  the `response` const in all these calls? We could just directly return the call?

client/webui/frontend/src/lib/api/projects/service.ts
```diff
@@ -79,3 +79,38 @@ export const importProject = async (file: File, options: { preserveName: boolean
...
+    const response = await fetch(api.webui.getFullUrl(`/api/v1/projects/${projectId}/shares`), {
+        method: "DELETE",
+        headers: {
+            "Content-Type": "application/json",
+            Authorization: `Bearer ${localStorage.getItem("access_token")}`,
+        },
+        body: JSON.stringify(data),
+    });
+
+    if (!response.ok) {
```
lgh-solace, line 97: There might be an error message in this response? I think there's a utility function `getErrorFromResponse` that might be helpful?

client/webui/frontend/src/lib/api/projects/service.ts
```diff
@@ -79,3 +79,38 @@ export const importProject = async (file: File, options: { preserveName: boolean
...
+    return response;
+};
+
+export const deleteProjectShares = async (projectId: string, data: BatchDeleteRequest) => {
+    // Using a custom fetch approach since api.webui.delete doesn't support body parameter
+    const response = await fetch(api.webui.getFullUrl(`/api/v1/projects/${projectId}/shares`), {
+        method: "DELETE",
+        headers: {
+            "Content-Type": "application/json",
+            Authorization: `Bearer ${localStorage.getItem("access_token")}`,
```
lgh-solace: Having a delete with a body is a bit controversial? I'm not sure what the body is, but we use something like this?

```
DELETE /api/v1/projects/${projectId}/shares?id-1,id-2,id-3...
```

Further, I don't think this authorization will always work as there can be different access tokens (there's a `sam_access_token` too now). So, I feel it would be best if we could go through some kind of `api.webui` call even if it have to open up the delete to take in optional body.

client/webui/frontend/src/lib/components/common/projectShareVariants.ts
```diff
@@ -0,0 +1,15 @@
+import { cva } from "class-variance-authority";
+
+// Icon button for remove/close actions
+export const classForIconButton = cva(["h-8", "w-8", "p-0", "text-[var(--muted-foreground)]", "hover:text-[var(--foreground)]"]);
```
lgh-solace: Minor: we can use shorter expressions like: `text-muted-foreground`

client/webui/frontend/src/lib/components/common/UserTypeahead.tsx
```diff
@@ -0,0 +1,193 @@
...
+                            ref={inputRef}
+                            type="text"
+                            placeholder="Search by email..."
+                            value={selectedEmail || searchQuery}
+                            onChange={handleInputChange}
+                            onKeyDown={handleKeyDown}
+                            onFocus={() => setIsOpen(true)}
+                            className={classForTypeaheadInput({ error })}
+                        />
+                        {isLoading && <Loader2 className={classForInputSpinner()} />}
```
lgh-solace: We have a `Spinner` component, maybe see if that works for you?

client/webui/frontend/src/lib/components/projects/ShareProjectDialog.tsx
```diff
@@ -0,0 +1,316 @@
...
+                {error && (
+                    <div className="flex-shrink-0 py-2">
+                        <MessageBanner variant="error" message={error} />
+                    </div>
+                )}
+
+                <div className="min-h-0 flex-1 overflow-y-auto">
+                    {isLoadingShares ? (
+                        <div className="flex items-center justify-center py-8">
+                            <Loader2 className="h-6 w-6 animate-spin text-[var(--muted-foreground)]" />
```
lgh-solace: Perhaps look at Spinner again... I think it takes a classname you can pass in for the colour.

client/webui/frontend/src/stories/projects/ShareProjectDialog.stories.tsx
```diff
@@ -0,0 +1,240 @@
...
+import { expect, screen, userEvent, within } from "storybook/test";
+import { http, HttpResponse } from "msw";
+import { ShareProjectDialog } from "@/lib";
+import type { Project, ProjectSharesResponse } from "@/lib/types/projects";
+
+// ============================================================================
+// Mock Data
+// ============================================================================
+
+const mockProject: Project = {
```
lgh-solace: We've started a practice of putting mocked data in their own files, can you check out `client/webui/frontend/src/stories/data` and see if that data is suitable or if you can files there? Thoughts?

lgh-solace (commented):
Very nice and clean! 

Only real concern is the api call for the batch delete, but could do a little tidying too: using condensed colour references, etc.

client/webui/frontend/src/lib/api/projects/service.ts
JKaram: OK I will take a look at the token part.

The id's are quite large, and I was worried we would hit the max header. Which is why we went with the body approach.

client/webui/frontend/src/lib/components/common/UserTypeahead.tsx
```diff
@@ -0,0 +1,192 @@
...
+interface UserTypeaheadProps {
+    id: string;
+    onSelect: (email: string, id: string) => void;
+    onRemove: (id: string) => void;
+    excludeEmails: string[];
+    selectedEmail?: string | null;
+    error?: boolean;
+}
+
+export const UserTypeahead: React.FC<UserTypeaheadProps> = ({ id, onSelect, onRemove, excludeEmails, selectedEmail, error }) => {
```
lgh-solace, line 28: Is this something that could be used in the Mentions UI?

client/webui/frontend/src/lib/components/common/UserTypeahead.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+    return (
+        <>
+            <Popover open={isOpen && showResults} onOpenChange={setIsOpen}>
+                <PopoverAnchor asChild>
+                    <div className="relative">
+                        <Input
+                            ref={inputRef}
+                            type="text"
+                            placeholder="Search by email..."
```
lgh-solace, line 118: Is it always emails?

client/webui/frontend/src/lib/schemas/shareProject.schema.ts
```diff
@@ -0,0 +1,35 @@
+import { z } from "zod";
+
+const EMAIL_REGEX = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
+
+const viewerEntrySchema = z.object({
+    id: z.string(),
+    email: z.string().nullable(),
+});
+
+export const createShareProjectFormSchema = (identityServiceType: string | null) =>
```
lgh-solace, line 10: Could we use camelCase for the file name?

lgh-solace (approved):

client/webui/frontend/src/lib/components/common/UserTypeahead.tsx
JKaram: Yes

client/webui/frontend/src/lib/components/common/UserTypeahead.tsx
JKaram: I tried, the requirements are a little too different for it to make sense to use it here.

PR #929, opened 2026-01-30
fix(DATAGO-123608): register trust cards in platform service

macstewart:
Signed identity and sam access token validation requires access to the gateway trust cards. The gateways and agents have access to these cards but the platform service did not, causing validation failures

This subscribes the platform service to trust manager if enabled so it can verify jwts from the webui gateway.

Note:
Platform service will currently also publish trust cards. This is a minor inefficiency that can be updated in an enterprise PR to allow receive-only trust manager configuration

### DIFF
--- a/client/webui/frontend/package-lock.json
+++ b/client/webui/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@SolaceLabs/solace-agent-mesh-ui",
-    "version": "1.38.11",
+    "version": "1.39.0",
     "lockfileVersion": 3,
     "requires": true,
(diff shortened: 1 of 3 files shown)


mo-radwan1 (approved):
Reviewed the changes - this PR safely adds trust card support to Platform Service.

**Key points verified:**
- Migrations remain in `__init__()` (no regression of PR #741 race condition fix)
- `super()._late_init()` only initializes Trust Manager which is purely in-memory (crypto keys, trust registry)
- No database operations in the late_init path
- Follows the same pattern used by agents and gateways

PR #941, opened 2026-02-01
fix(DATAGO-123803): Fix parallel workflow visualization showing empty container

efunneko:

Fixes a bug where parallel workflow executions were displayed incorrectly in the Activity visualization. The parallel container was shown empty, with the workflows appearing sequentially after/outside it instead of being displayed side-by-side within the container.

Modified `layoutEngine.ts` and `types.ts` to properly link workflow tool invocations to their corresponding workflow execution starts.

**Root cause:** `handleWorkflowStart` was using `step.functionCallId` to find the parallel block, but this was `undefined` for workflow executions. The workflow tool invocation's `functionCallId` wasn't being linked to the workflow execution start event.

**Fix:**
1. Added a new map `workflowParentTaskFunctionCallIds` to track workflow `functionCallIds` by their `parentTaskId` (the orchestrator's task ID)
2. In `handleToolInvocation` for workflow tools: store the `functionCallId` keyed by the tool invocation's `owningTaskId`
3. In `handleWorkflowStart`: look up `functionCallIds` using the workflow's `parentTaskId`, then find and add the workflow to the correct parallel block

The key insight is that `tool_invocation.owningTaskId == workflow_start.parentTaskId` (both are the orchestrator's task ID).

- [x] Manual testing: Verified parallel workflows now appear side-by-side within the parallel container
- [ ] Unit tests: No new tests - existing visualization tests cover the general flow
- [ ] Integration tests: N/A
- [ ] Known limitations: Only tested with two parallel workflows

The fix adds a second approach to finding the parallel block. The existing approach (via `subTaskId`) is preserved as a fallback. The new approach uses `parentTaskId` which is always available on workflow execution start events.

### DIFF
--- a/client/webui/frontend/src/lib/components/activities/FlowChart/utils/layoutEngine.ts
+++ b/client/webui/frontend/src/lib/components/activities/FlowChart/utils/layoutEngine.ts
@@ -50,6 +50,8 @@ export function processSteps(steps: VisualizerStep[], agentNameMap: Record<strin
         parallelPeerGroupMap: new Map(),
         parallelBlockMap: new Map(),
         subWorkflowParentMap: new Map(),
+        workflowFunctionCallIdMap: new Map(),
+        workflowParentTaskFunctionCallIds: new Map(),
     };
 
     // Process all steps to build tree structure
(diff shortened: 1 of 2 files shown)


client/webui/frontend/src/lib/components/activities/FlowChart/utils/layoutEngine.ts
```diff
@@ -569,9 +588,34 @@ function handleWorkflowStart(step: VisualizerStep, context: BuildContext): void
...
+    // 1. Direct lookup by owningTaskId or executionId (when subTaskId was available)
+    // 2. Lookup by parentTaskId (for workflows where subTaskId wasn't available)
+    // 3. Fall back to step.functionCallId
+    let functionCallId: string | undefined = context.workflowFunctionCallIdMap.get(step.owningTaskId || "") ||
+                                             context.workflowFunctionCallIdMap.get(executionId || "") ||
+                                             step.functionCallId;
     let addedToParallelBlock = false;
 
+    // If not found directly, try to find via parentTaskId
+    if (!functionCallId && step.parentTaskId) {
```
efunneko, line 601: Address the problem shown here:

Image

This now properly puts the parallel activities within the container box rather in series following it. Here is the fixed version:

Image

lgh-solace (approved):

PR #951, opened 2026-02-02
chore(DATAGO-123922): Add artifact tags for internal and user-uploaded files 

efunneko:

Add a tagging system to artifact metadata that allows labeling artifacts with arbitrary tags. The primary use case is reserving the `__internal` tag to mark artifacts that should be hidden from users by default (e.g., intermediate files from workflows or structured invocations), with a UI toggle to show them when needed.

**Backend changes:**
- Added `tags` field to `ArtifactInfo` type and data models (`types.py`, `data_parts.py`)
- Modified `save_artifact_with_metadata()` to accept and store tags in metadata JSON
- Modified `get_artifact_info_list()` to extract and return tags from metadata
- Auto-tag user uploads with `__user_uploaded` via `process_artifact_upload()`
- Auto-tag artifacts from structured invocations and workflows with `__internal` via callbacks, proxies, and workflow components
- Added tag parsing from fenced block params in `callbacks.py`

**Frontend changes:**
- Added `tags` field to `ArtifactInfo` interface
- Added `useArtifacts` hook enhancements: `showInternalArtifacts` toggle with localStorage persistence, filtering logic, `internalArtifactCount`
- Added "Show/Hide Internal Files" toggle in `ArtifactMorePopover`
- Updated `ArtifactPanel` to show count of hidden internal files when toggle is off

### Key Design Decisions

- System tags use `__` prefix (e.g., `__internal`, `__user_uploaded`) to distinguish from user tags
- Tags stored in `.metadata.json` files alongside artifacts for backward compatibility
- Internal artifacts hidden by default but accessible via toggle (not deleted)
- Case-insensitive tag matching for robustness

- [x] Manual testing: Verified internal files are hidden by default, toggle shows/hides them, user uploads get `__user_uploaded` tag
- [x] Integration tests: Added `TestArtifactTagging` class in `test_artifact_services.py` with 5 tests covering tag storage, retrieval, and backward compatibility
- [x] Known limitations: Frontend filtering tests not included (would require E2E testing infrastructure)

- The `__internal` tag is auto-applied in multiple locations (callbacks, proxies, workflow components) - verify all paths are covered
- Backward compatibility: Old artifacts without tags should display normally (tested in `test_artifact_without_tags`)
- The toggle state is persisted in localStorage (`sam_show_internal_artifacts`)


client/webui/frontend/src/lib/components/chat/artifact/ArtifactMorePopover.tsx
```diff
@@ -1,29 +1,53 @@
...
 
-    const menuActions: MenuAction[] = [
-        {
-            id: "refreshAll",
-            label: "Refresh",
-            onClick: () => {
-                artifactsRefetch();
-            },
-            icon: <RefreshCcw />,
+    const menuActions: MenuAction[] = [];
```
efunneko, line 26: Added hidden concept within artifact file list:

Image

and

Image

src/solace_agent_mesh/agent/adk/callbacks.py
```diff
@@ -306,13 +311,17 @@ async def process_artifact_blocks_callback(
                                 host_component=host_component,
                                 log_identifier=f"{log_identifier}[ResolveChunk]",
                             )
+                            # Extract tags from params (comma-separated string to list)
+                            tags_str = params.get("tags", "")
```
amir-ghasemi: the logic to parse tags from the parameters string is duplicated. Perhaps extract it to a helper function?

src/solace_agent_mesh/agent/adk/callbacks.py
```diff
@@ -395,6 +404,21 @@ async def process_artifact_blocks_callback(
                                     log_identifier,
                                     params["schema_max_keys"],
                                 )
+                        # Extract tags from params (comma-separated string to list)
```
amir-ghasemi, line 420: Same as above.

PR #957, opened 2026-02-04
fix(DATAGO-124109): silently drop redundant LLM-generated _notify_artifact_save calls 

efunneko:
## Summary
- Fixes duplicate artifact save notifications appearing in the UI visualization
- Modifies `sanitize_tool_names_callback` to detect and silently drop LLM-generated `_notify_artifact_save` calls (those without the `host-notify-` prefix)
- Preserves legitimate system-generated calls while filtering out redundant LLM-generated duplicates

## Root Cause
When artifacts are saved via fenced blocks, the system injects a legitimate `_notify_artifact_save` call with a `host-notify-` prefix. However, the LLM sometimes also generates its own redundant call with a different ID, causing duplicate notifications in the UI.

## Test plan
- [x] Added unit tests for unauthorized internal tool detection
- [x] All 36 existing tests pass
- [ ] Manual verification with reproduction scenario from stim file

### DIFF
--- a/src/solace_agent_mesh/agent/adk/callbacks.py
+++ b/src/solace_agent_mesh/agent/adk/callbacks.py
@@ -2632,20 +2632,31 @@ def sanitize_tool_names_callback(
         return None
     
     # Find all function calls and check for invalid tool names
-    invalid_calls = []
+    invalid_calls = []  # Calls that need error responses
+    silently_dropped_calls = []  # Calls to drop without error (e.g., redundant internal tool calls)
     valid_parts = []
-    
+
     for part in llm_response.content.parts:
         if part.function_call:
             tool_name = part.function_call.name
-            
+            function_call_id = part.function_call.id or ""
+
             # Check for placeholder patterns (tool names starting with $)
             # This catches $FUNCTION_NAME, $ARTIFACT_TOOL, $TOOL_NAME, etc.
             is_placeholder = tool_name.startswith('$')
-            
+
             # Check against Bedrock's validation pattern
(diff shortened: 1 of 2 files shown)


src/solace_agent_mesh/agent/adk/callbacks.py
```diff
@@ -2632,20 +2632,31 @@ def sanitize_tool_names_callback(
...
+
             # Check against Bedrock's validation pattern
             is_valid_format = VALID_TOOL_NAME_PATTERN.match(tool_name) is not None
-            
+
+            # Check for unauthorized _notify_artifact_save calls
+            # This tool should only be called by the system with host-notify- prefix
+            # LLM-generated calls (without this prefix) should be silently dropped
+            # since the artifact was already saved and a system call was injected
+            is_unauthorized_internal_tool = (
```
efunneko, line 2655: Occasionally, the LLM will call this internal tool that we auto-inject after an inline artifact creation. We now will filter it out to reduce the noise

src/solace_agent_mesh/agent/adk/callbacks.py
```diff
@@ -2632,20 +2632,31 @@ def sanitize_tool_names_callback(
...
+                tool_name == "_notify_artifact_save"
+                and not function_call_id.startswith("host-notify-")
```
amir-ghasemi, line 2657: The "host-notify-" prefix is used as a magic string. Consider defining it as a constant:

`SYSTEM_NOTIFY_PREFIX = "host-notify-"`

PR #959, opened 2026-02-04
fix(DATAGO-124054): prevent workflow arrow heads from being cut off by nodes

amir-ghasemi:
Fixes the issue where workflow arrow heads are being cut off by the nodes beneath them in the workflow diagram view.

Added a 4-pixel offset to the edge target Y position in `WorkflowDiagram.tsx`. This ensures the arrowheads are rendered slightly above the target nodes, making them fully visible instead of being obscured by the node elements.

The change is in the `calculatedEdges` computation where edges are positioned based on layout node positions.

### Key Design Decisions
- Used a small 4-pixel offset which is enough to make the arrowhead visible without creating a noticeable gap between the arrow and the node
- The arrowhead marker is 12px tall with refX=9, so the tip extends 3px past the path endpoint - the 4px offset ensures the full arrowhead is visible

- [x] Manual testing: Verified workflow diagrams display arrows with fully visible arrowheads
- [x] Build verification: npm run build passes without errors

The offset value (4px) was chosen to be minimal while still ensuring visibility. If the arrowheads still appear slightly cut off in some edge cases, this value could be increased.

AFTER:
image

### DIFF
--- a/client/webui/frontend/src/lib/components/workflowVisualization/WorkflowDiagram.tsx
+++ b/client/webui/frontend/src/lib/components/workflowVisualization/WorkflowDiagram.tsx
@@ -152,6 +152,10 @@ const WorkflowDiagram: React.FC<WorkflowDiagramProps> = ({
         // Build flat map of all node positions from layout tree
         const nodePositions = buildNodePositionMap(layout.nodes);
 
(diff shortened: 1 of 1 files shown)


cyrus2281 (approved):
LGTM

PR #961, opened 2026-02-04
fix(DATAGO-124129):  Workflows: “delay” setting not visible for Loop node in Properties panel

solace-pjones:

Add support for displaying the delay property on loop nodes in the workflow visualization UI. This allows users to see and configure the delay between loop iterations when viewing workflow details.

Below image shows Details Panel with the delay value set.
image

Backend:

Modified src/solace_agent_mesh/workflow/component.py to serialize the delay property for loop nodes in the workflow config JSON (only includes delay when set, maintaining backward compatibility)

Frontend:

Added delay?: string to the LayoutNode data interface in utils/types.ts
Updated WorkflowNodeDetailPanel.tsx to display the delay property in the details view for loop nodes
Created comprehensive Storybook stories at stories/workflowVisualization/WorkflowNodeDetailPanel.stories.tsx with tests verifying delay rendering behavior

Optional delay property: The delay is only included in serialized config when explicitly set, maintaining backward compatibility with existing workflows
UI display strategy: For loop nodes, the component displays node ID in the title (not the label), consistent with how loop nodes are referenced elsewhere in the system

### DIFF
--- a/client/webui/frontend/src/stories/workflowVisualization/WorkflowNodeDetailPanel.stories.tsx
+++ b/client/webui/frontend/src/stories/workflowVisualization/WorkflowNodeDetailPanel.stories.tsx
@@ -1,0 +1,71 @@
+import type { Meta, StoryObj } from "@storybook/react-vite";
+import { expect, screen } from "storybook/test";
+import WorkflowNodeDetailPanel from "@/lib/components/workflowVisualization/WorkflowNodeDetailPanel";
+import type { LayoutNode } from "@/lib/components/workflowVisualization/utils/types";
+
+const meta: Meta<typeof WorkflowNodeDetailPanel> = {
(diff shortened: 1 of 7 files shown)


cyrus2281 (approved):
LGTM

client/webui/frontend/src/stories/workflowVisualization/WorkflowNodeDetailPanel.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+        Story => (
+            <div className="h-screen bg-white dark:bg-gray-900">
+                <Story />
+            </div>
+        ),
+    ],
+};
+
+export default meta;
+type Story = StoryObj<typeof meta>;
```
lgh-solace, line 20: Nice!

PR #967, opened 2026-02-04
fix(DATAGO-123604): add descriptions for workflow logic nodes

RohanSreelesh:

Added descriptions for workflow logic nodes. The descriptions are similar to the ones in the docs

Added field picture:
image

### DIFF
--- a/client/webui/frontend/src/lib/components/workflowVisualization/WorkflowNodeDetailPanel.tsx
+++ b/client/webui/frontend/src/lib/components/workflowVisualization/WorkflowNodeDetailPanel.tsx
@@ -275,7 +275,7 @@ const WorkflowNodeDetailPanel: React.FC<WorkflowNodeDetailPanelProps> = ({ node,
                         )}
 
                         {/* Description (for logic nodes) */}
-                        {(node.type === "map" || node.type === "switch" || node.type === "loop") && LOGIC_NODE_DESCRIPTIONS[node.type] && (
+                        {(node.type === "map" || node.type === "switch" || node.type === "loop") && (
                             <div className="mb-4">
                                 <label className="mb-1 block text-sm font-medium text-(--color-secondary-text-wMain)">Description</label>
                                 <div className="text-sm">{LOGIC_NODE_DESCRIPTIONS[node.type]}</div>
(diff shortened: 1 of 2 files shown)


client/webui/frontend/src/lib/components/workflowVisualization/WorkflowNodeDetailPanel.tsx
```diff
@@ -268,6 +274,14 @@ const WorkflowNodeDetailPanel: React.FC<WorkflowNodeDetailPanelProps> = ({ node,
                             </div>
                         )}
 
+                        {/* Description (for logic nodes) */}
+                        {(node.type === "map" || node.type === "switch" || node.type === "loop") && LOGIC_NODE_DESCRIPTIONS[node.type] && (
```
lgh-solace: Not a big deal, but could reduce the number of checks, removing `LOGIC_NODE_DESCRIPTIONS[node.type]` or something from the condition since we know it's there?

client/webui/frontend/src/stories/workflowVisualization/WorkflowNodeDetailPanel.stories.tsx
```diff
@@ -64,8 +95,50 @@ export const LoopNodeWithoutDelay: Story = {
...
     },
 };
+
+export const MapNode: Story = {
+    args: {
+        node: createMapNode(),
+        workflowConfig: null,
+        agents: [],
+    },
+    play: async () => {
```
lgh-solace: The pattern pre-exists in this file, but I think it would be slightly beneficial to use the canvas, like this:

```
    play: async ({ canvas }) => {
        // Verify node ID appears twice: once in title header, once in Node ID section
        const pollingLoopElements = canvas.getAllByText("polling_loop");
```

It just makes the tests slightly more efficient since they are pre-scoped to the story container (we do need screen for things like dialogs or toasts since they are outside the canvas).

RohanSreelesh: done

client/webui/frontend/src/lib/components/workflowVisualization/WorkflowNodeDetailPanel.tsx
RohanSreelesh: removed

PR #973, opened 2026-02-04
fix(DATAGO-124046): fix nested agent visualization in activity flow

efunneko:
Fixes the activity flow visualization where nested agent nodes (like SupportTicketAgent) were not appearing correctly. The agent would be missing from the diagram and its tools would incorrectly appear directly under the parent Orchestrator.

**taskVisualizerProcessor.ts:**
- Search all tool calls for matching sub-tasks (not just those with `peer_`/`workflow_` prefix) to properly detect delegation relationships
- Trust function_call_id match when timestamp is unavailable (request events don't have `result.status.timestamp`)

**layoutEngine.ts:**
- `handleUserRequest`: Insert User node at beginning with `unshift()` to ensure it always appears first; check for existing agent before creating duplicates
- `handleLLMCall`: Create parent agent on-demand if not found when processing LLM call steps
- `handleToolInvocation`: Create parent agent on-demand for peer invocations if not found

These changes handle cases where steps arrive out of order (e.g., tool invocation before USER_REQUEST) which can happen when loading historical task events.

- [x] Manual testing: Loaded stim files with nested agent calls (SupportTicketAgent under Orchestrator) and verified correct visualization
- [ ] Unit tests: N/A - visualization logic is complex to unit test
- [x] Known limitations: Visual regression testing would be beneficial

The on-demand agent creation in `handleLLMCall` and `handleToolInvocation` - ensure this doesn't cause issues with normal real-time event processing where steps typically arrive in order.

### DIFF
--- a/client/webui/frontend/src/lib/components/activities/FlowChart/utils/layoutEngine.ts
+++ b/client/webui/frontend/src/lib/components/activities/FlowChart/utils/layoutEngine.ts
@@ -76,26 +76,6 @@ export function processSteps(steps: VisualizerStep[], agentNameMap: Record<strin
         processStep(step, context);
     }
 
-    // DEBUG: Print tree structure
-    console.log("=== LAYOUT ENGINE DEBUG ===");
-    console.log("Steps processed:", steps.length);
-    console.log("Root nodes:", context.rootNodes.length);
-    console.log("taskToNodeMap keys:", Array.from(context.taskToNodeMap.keys()));
-    console.log("subWorkflowParentMap keys:", Array.from(context.subWorkflowParentMap.keys()));
-
-    const printTree = (node: LayoutNode, indent: string = "") => {
-        console.log(`${indent}[${node.type}] ${node.data.label} (id=${node.id}, owningTaskId=${node.owningTaskId || "none"})`);
-        for (const child of node.children) {
-            printTree(child, indent + "  ");
-        }
-    };
-
-    console.log("Tree structure:");
-    for (const node of context.rootNodes) {
-        printTree(node);
-    }
-    console.log("=== END DEBUG ===");
-
     // Calculate layout (positions and dimensions)
     const nodes = calculateLayout(context.rootNodes);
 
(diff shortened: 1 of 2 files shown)


client/webui/frontend/src/lib/components/activities/FlowChart/utils/layoutEngine.ts
```diff
@@ -76,26 +76,6 @@ export function processSteps(steps: VisualizerStep[], agentNameMap: Record<strin
         processStep(step, context);
     }
 
-    // DEBUG: Print tree structure
```
efunneko, line 79: Past debug logs that should have been removed before

client/webui/frontend/src/lib/components/activities/FlowChart/utils/layoutEngine.ts
```diff
@@ -211,7 +197,25 @@ function handleUserRequest(step: VisualizerStep, context: BuildContext): void {
  * Handle AGENT_LLM_CALL - adds LLM child to current agent
  */
 function handleLLMCall(step: VisualizerStep, context: BuildContext): void {
-    const agentNode = findAgentForStep(step, context);
+    let agentNode = findAgentForStep(step, context);
```
efunneko, line 200: There was some annoying out of order events that caused rendering problems. Was:

Image

and now is:

Image

Note that this image uses the exact .stim file from the bug (I implemented stim upload to test -  separate branch)

client/webui/frontend/src/lib/components/activities/FlowChart/utils/layoutEngine.ts
```diff
@@ -341,8 +345,29 @@ function handleToolInvocation(step: VisualizerStep, context: BuildContext): void
...
+            "agent",
+            {
+                label: displayName,
+                visualizerStepId: step.id,
+            },
+            step.owningTaskId
+        );
+        context.rootNodes.push(agentNode);
+        context.currentAgentNode = agentNode;
+        context.taskToNodeMap.set(step.owningTaskId, agentNode);
```
lgh-solace, line 365: There seem to be repeating patterns in the layout handling - perhaps we can make helper functions to reduce duplication and make things more readable?

lgh-solace (approved):
Thanks!

PR #975, opened 2026-02-04
fix(DATAGO-123680): Use node name for showing switch-case details

alimosaed:
The detail side panel of switch-case node uses the node id to refer to nodes; however, node id is not meaningful for users.
image

Added a handler to UI to retrieve the node name and use it in the panel.
Screenshot 2026-02-04 at 4 33 30 PM

- [x] Manual testing: Built the project and loaded a switch-case example and checked the panel.
- [ ] Unit tests
- [ ] Integration tests
- [ ] Known limitations

Nothing


solace-pjones (approved):
looks good.

PR #976, opened 2026-02-04
fix(DATAGO-124223): Investigate and resolve prompt import error

amir-ghasemi:
This pull request updates the uniqueness constraint on the `command` field in the `prompt_groups` table to be per user rather than global. This change allows different users to create prompt groups with the same command name, addressing a limitation in the previous schema. The migration is implemented both at the database level and in the SQLAlchemy model, and relevant comments and code are updated for clarity.

**Database Schema Migration:**

* Added Alembic migration `20260204_fix_prompt_command_unique_constraint.py` to drop the global unique index on `command` and replace it with a composite unique index on (`command`, `user_id`), allowing commands to be unique per user. Also adds a non-unique index on `command` for performance.

**SQLAlchemy Model Update:**

* Updated `PromptGroupModel` in `prompt_model.py` to remove the global unique constraint from the `command` column and add a composite unique index on (`command`, `user_id`). The `command` column is now only indexed, not unique. [[1]](diffhunk://#diff-10baca1e9c1f47b4f44d237ec981fb277b9fc0d6b242bb3daafe8db4da7af20fL6-R6) [[2]](diffhunk://#diff-10baca1e9c1f47b4f44d237ec981fb277b9fc0d6b242bb3daafe8db4da7af20fR25-R38)

**Code and Comment Updates:**

* Updated comments and logic in `prompts.py` to clarify that command conflicts are now handled per user instead of globally.

### DIFF
--- a/src/solace_agent_mesh/gateway/http_sse/alembic/versions/20260204_fix_prompt_command_unique_constraint.py
+++ b/src/solace_agent_mesh/gateway/http_sse/alembic/versions/20260204_fix_prompt_command_unique_constraint.py
@@ -48,7 +48,13 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    """Revert to global unique constraint on command."""
+    """Revert to global unique constraint on command.
+    
+    Note: This is a best-effort downgrade. If there are duplicate commands
+    across users, the unique index creation will fail, but we allow the
+    migration to continue with a non-unique index instead.
+    """
+    from sqlalchemy import inspect
     
     # Drop the composite unique index
     op.drop_index('ix_prompt_groups_command_user_id', table_name='prompt_groups')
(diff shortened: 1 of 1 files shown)


src/solace_agent_mesh/gateway/http_sse/alembic/versions/20260204_fix_prompt_command_unique_constraint.py
```diff
@@ -0,0 +1,66 @@
...
+    
+    # Drop the composite unique index
+    op.drop_index('ix_prompt_groups_command_user_id', table_name='prompt_groups')
+    
+    # Drop the non-unique command index
+    op.drop_index('ix_prompt_groups_command', table_name='prompt_groups')
+    
+    # Recreate the original global unique index
+    # Note: This may fail if there are duplicate commands across users
+    op.create_index(
```
RohanSreelesh: Should we wrap this in a try catch so that its best effort? There will be a slight drift in the code and the actual table but I would think this is better than blocking the downgrade/rollback completely? In our case if migrations fail then the app doesn't startup which is what worries me

PR #983, opened 2026-02-05
fix(DATAGO-123843): show complete agent response in User Node modal

efunneko:
Fixes incomplete agent response display in the User Node modal on the Activities page. The modal was only showing the last text fragment instead of the complete response.

Modified `findUserNodeDetails()` in `nodeDetailsHelper.ts` to aggregate ALL `AGENT_RESPONSE_TEXT` steps at nesting level 0, instead of only using the last one. When multiple response steps exist (due to streaming interruptions from LLM calls or tool invocations), they are combined into a synthetic step with the full text.

### Key Design Decisions
- Used `filter()` to collect all top-level response steps rather than `find()` which only returns one
- Created a synthetic step when multiple responses exist to preserve the step structure expected by the rendering code
- Joined text without separators since the original streaming would have been continuous

- [x] Manual testing: Verified complete response now shows in User Node modal
- [x] Build verification: Frontend builds successfully

The aggregation assumes all top-level `AGENT_RESPONSE_TEXT` steps should be combined. This matches the existing post-processing logic in `taskVisualizerProcessor.ts` that aggregates text for `TASK_COMPLETED` steps.

### DIFF
--- a/client/webui/frontend/src/lib/components/activities/FlowChart/utils/nodeDetailsHelper.ts
+++ b/client/webui/frontend/src/lib/components/activities/FlowChart/utils/nodeDetailsHelper.ts
@@ -90,8 +90,34 @@ function findUserNodeDetails(node: LayoutNode, primaryStep: VisualizerStep, allS
 
     // Bottom user node: show final response
     if (node.data.isBottomNode) {
-        // Find the last AGENT_RESPONSE_TEXT at nesting level 0
-        const finalResponse = [...allSteps].reverse().find(s => s.type === "AGENT_RESPONSE_TEXT" && s.nestingLevel === 0);
+        // Find ALL AGENT_RESPONSE_TEXT steps at nesting level 0 (not just consecutive ones)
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/components/activities/FlowChart/utils/nodeDetailsHelper.ts
```diff
@@ -90,8 +90,34 @@ function findUserNodeDetails(node: LayoutNode, primaryStep: VisualizerStep, allS
 
     // Bottom user node: show final response
     if (node.data.isBottomNode) {
-        // Find the last AGENT_RESPONSE_TEXT at nesting level 0
-        const finalResponse = [...allSteps].reverse().find(s => s.type === "AGENT_RESPONSE_TEXT" && s.nestingLevel === 0);
+        // Find ALL AGENT_RESPONSE_TEXT steps at nesting level 0 (not just consecutive ones)
```
efunneko, line 93: When clicking on the final User node in the visualization, you now see the full task output. It goes from this:

Image

to this:

Image

PR #997, opened 2026-02-07
fix(DATAGO-124731): fixing repeated status calls

lgh-solace:

Status calls were not waiting 5s to repeat due to state changes that caused a callback to change too often.

Store backgroundTasks in a ref and read from it inside the callback. This makes checkAllBackgroundTasks stable, and using hasBackgroundTasks (a boolean) as the effect dependency ensures the interval only restarts when transitioning between having tasks and not having tasks—not on every task list mutation.

- [x] Manual testing: [See videos below.]

// BEFORE (267 requests)

https://github.com/user-attachments/assets/a74ac24e-3b6f-4d81-91a7-9317c2723049

// AFTER (45 requests, same prompt)

https://github.com/user-attachments/assets/12f5bd02-d8fc-48f7-b9f1-e2b08add12f2

### DIFF
--- a/client/webui/frontend/src/lib/hooks/useBackgroundTaskMonitor.ts
+++ b/client/webui/frontend/src/lib/hooks/useBackgroundTaskMonitor.ts
@@ -1,9 +1,4 @@
-/**
- * Hook for monitoring and reconnecting to background tasks.
- * Stores active background tasks in localStorage and automatically reconnects on session load.
- */
-
-import { useState, useEffect, useCallback } from "react";
+import { useState, useEffect, useCallback, useRef } from "react";
 import { api } from "@/lib/api";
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/hooks/useBackgroundTaskMonitor.ts
```diff
@@ -195,12 +201,13 @@ export function useBackgroundTaskMonitor({ userId, onTaskCompleted, onTaskFailed
 
     // Periodic checking to detect background task completion when not connected to SSE
     // This handles the case where a task completes while the user is on a different session
+    const hasBackgroundTasks = backgroundTasks.length > 0;
```
lgh-solace, line 204: We only care if there's more than 1.

solace-mdupls (approved):

PR #1002, opened 2026-02-09
fix(DATAGO-124829): Implement row locking to prevent race conditions in event buffer reads

amir-ghasemi:
This pull request updates the way events are queried from the `SSEEventBufferModel` in the `get_buffered_events` method to improve concurrency safety. Now, when events are being marked as consumed, the query uses row-level locking to prevent race conditions during concurrent read and update operations.

**Concurrency improvements:**

* Added the use of `with_for_update()` to the SQLAlchemy query in `get_buffered_events` when `mark_consumed` is `True`, ensuring row-level locks are acquired to prevent concurrent read-then-update race conditions.

### DIFF
--- a/src/solace_agent_mesh/gateway/http_sse/repository/sse_event_buffer_repository.py
+++ b/src/solace_agent_mesh/gateway/http_sse/repository/sse_event_buffer_repository.py
@@ -148,7 +148,7 @@ class SSEEventBufferRepository:
         Args:
             db: Database session
             task_id: The task ID to get events for
-            mark_consumed: Whether to mark events as consumed
+            mark_consumed: Whether to mark events as consumed 
             
         Returns:
             List of event dictionaries with type, data, and sequence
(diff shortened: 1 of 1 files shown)


src/solace_agent_mesh/gateway/http_sse/repository/sse_event_buffer_repository.py
```diff
@@ -153,10 +153,14 @@ def get_buffered_events(
         Returns:
             List of event dictionaries with type, data, and sequence
         """
-        events = db.query(SSEEventBufferModel)\
-            .filter(SSEEventBufferModel.task_id == task_id)\
-            .order_by(SSEEventBufferModel.event_sequence)\
-            .all()
+        query = db.query(SSEEventBufferModel)\
+            .filter(SSEEventBufferModel.task_id == task_id)
```
rudraneel-chakraborty, line 157: I Think to make this 100 percent safe, just add 
```
.filter(SSEEventBufferModel.consumed == False)
```
without this the duplicate read problem will remain 

this current change is great for preventing `concurrent MODIFICATION of same rows`

amir-ghasemi: Thanks, added.

PR #1006, opened 2026-02-10
fix(DATAGO-122883): Add limits for batch upload and project size

shanechen-solace:

1. Add a limit to the total size of files that can be uploaded in one batch to a project. This is to prevent file processing from taking a long time.
2. Add a limit to the total size of files that can be uploaded to a project to prevent unbounded storage usage across users.

Designed a three-tier validation system in project_service.py that enforces limits at each stage of the upload process:

Per-file limit (default 50 MB) — each individual file is validated during streaming upload
Batch upload limit (default 100 MB) — the total size of all files in a single upload request
Total project size limit (default 100 MB) — cumulative size of all user-uploaded files in a project

### Key Design Decisions

- The total project size limit only counts user-uploaded artifacts (filtered by source="project"). LLM-generated artifacts are excluded so AI-generated content doesn't consume a user's upload quota.
- The existing `gateway_max_upload_size_bytes` config variable was kept as-is for the per-file limit rather than renaming it to something more specific (e.g., `gateway_max_per_file_upload_size_bytes`) to maintain backwards compatibility with existing gateway configurations. New config variables (`gateway_max_batch_upload_size_bytes`, `gateway_max_project_size_bytes`) were added for the additional limits.
- Note that ZIP imports have their own existing size limit (100 MB) and gracefully skip oversized individual files with warnings rather than failing the entire import. This limit was largely untouched during the implementation of these other limits.

- [x] Manual testing: Manually added files of different sizes to projects to test the upload limit behavior.

- [x] Integration tests: Added a comprehensive test suite covering limit enforcement, valid uploads, ZIP imports, edge cases, and verification that only user-uploaded files count toward the limit.

https://github.com/user-attachments/assets/78bb2bf5-e008-4b5a-a63a-2f701f9d4d7f

Contact @zhenyu369 for specifications on these file limits.

### DIFF
--- a/client/webui/frontend/src/lib/utils/fileValidation.ts
+++ b/client/webui/frontend/src/lib/utils/fileValidation.ts
@@ -72,14 +72,14 @@ export function validateFileSizes(files: FileList | File[], options: FileSizeVal
     if (oversizedFiles.length === 1) {
         const file = oversizedFiles[0];
         if (includeFileSizes) {
-            const fileSizeWithUnit = formatBytes(file.size, 2);
+            const fileSizeWithUnit = formatBytes(file.size);
             errorMsg = `File "${file.name}" (${fileSizeWithUnit}) exceeds the maximum size of ${maxSizeWithUnit}.`;
         } else {
             errorMsg = `File "${file.name}" exceeds the maximum size of ${maxSizeWithUnit}.`;
         }
     } else {
         const fileList = oversizedFiles.slice(0, maxFilesToList);
-        const fileNames = includeFileSizes ? fileList.map(f => `${f.name} (${formatBytes(f.size, 2)})`) : fileList.map(f => f.name);
+        const fileNames = includeFileSizes ? fileList.map(f => `${f.name} (${formatBytes(f.size)})`) : fileList.map(f => f.name);
 
         const remaining = oversizedFiles.length - maxFilesToList;
         const suffix = remaining > 0 ? ` and ${remaining} more` : "";
@@ -113,7 +113,7 @@ export function validateBatchUploadSize(files: FileList | File[], maxBatchUpload
         return { valid: true };
     }
 
-    const totalBatchWithUnit = formatBytes(totalBatchSize, 2);
+    const totalBatchWithUnit = formatBytes(totalBatchSize);
     const limitWithUnit = formatBytes(maxBatchUploadSizeBytes, 0);
 
     return {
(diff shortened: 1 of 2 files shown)


src/solace_agent_mesh/gateway/constants.py
```diff
@@ -0,0 +1,28 @@
...
+
+# Message size limits
+DEFAULT_GATEWAY_MAX_MESSAGE_SIZE_BYTES = 10_000_000  # 10MB - max message size for gateway publishing
+
+# ===== FILE UPLOAD SIZE LIMITS =====
+
+# Production defaults
+DEFAULT_MAX_PER_FILE_UPLOAD_SIZE_BYTES = 52428800  # 50MB - per-file upload limit
+DEFAULT_MAX_BATCH_UPLOAD_SIZE_BYTES = 104857600  # 100MB - batch upload limit (sum of files in one upload)
+DEFAULT_MAX_ZIP_UPLOAD_SIZE_BYTES = 104857600  # 100MB - ZIP import limit
```
lgh-solace, line 27: Interesting... does this mean a user could create a project with 100MB of files, export it and then not be able to import it because, with the small project files (for name, description, default agent, etc.) it would be a little bit more?

client/webui/frontend/src/lib/utils/file-validation.ts
```diff
@@ -56,25 +66,25 @@ export function validateFileSizes(files: FileList | File[], options: FileSizeVal
...
+            const fileSizeWithUnit = formatBytes(file.size, 2);
+            errorMsg = `File "${file.name}" (${fileSizeWithUnit}) exceeds the maximum size of ${maxSizeWithUnit}.`;
         } else {
-            errorMsg = `File "${file.name}" exceeds the maximum size of ${maxSizeMB} MB.`;
+            errorMsg = `File "${file.name}" exceeds the maximum size of ${maxSizeWithUnit}.`;
         }
     } else {
         const fileList = oversizedFiles.slice(0, maxFilesToList);
-        const fileNames = includeFileSizes ? fileList.map(f => `${f.name} (${(f.size / (1024 * 1024)).toFixed(2)} MB)`) : fileList.map(f => f.name);
+        const fileNames = includeFileSizes ? fileList.map(f => `${f.name} (${formatBytes(f.size, 2)})`) : fileList.map(f => f.name);
```
lgh-solace: I think "2" is the default so don't need to express it in most usages here.

src/solace_agent_mesh/gateway/constants.py
shanechen-solace: I believe the validation only cares about the user-uploaded artifacts. Other metadata does not count.

client/webui/frontend/src/lib/utils/file-validation.ts
shanechen-solace: Fixed. Also renamed file-validation.ts to fileValidation.ts to match the convention in the `utils/` folder.

lgh-solace (approved):
Thanks! Please wait for Zhenyu to have a chance to look at it before merging.

PR #1015, opened 2026-02-11
fix(DATAGO-125073): ensure projects are fetched fresh on projects page mount

JKaram:

   `/projects` api is only being called once on app render. This PR will refetch `/projects` on page mount. 
   
   This issue is especially prominent with project sharing.

    React Query `refetchOnMount` property. 

    No
    
    

https://github.com/user-attachments/assets/bf795bbd-4f49-42f7-b407-a8bc6b690c7f

### DIFF
--- a/client/webui/frontend/src/lib/api/projects/hooks.ts
+++ b/client/webui/frontend/src/lib/api/projects/hooks.ts
@@ -22,8 +22,12 @@ export function useProjects() {
     return useQuery({
         queryKey: projectKeys.lists(),
         queryFn: projectService.getProjects,
+        refetchOnMount: "always",
     });
 }
+
+/** Ensures projects are fetched fresh on component mount */
+export const useFetchProjectsOnMount = () => useProjects();
 
 export function useProjectArtifacts(projectId: string | null) {
     return useQuery({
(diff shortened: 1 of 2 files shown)


client/webui/frontend/src/lib/components/projects/ProjectsPage.tsx
```diff
@@ -13,9 +13,10 @@ import type { Project } from "@/lib/types/projects";
 import { Button, Header } from "@/lib/components";
 import { downloadBlob, getErrorMessage } from "@/lib/utils";
 import { useChatContext, useIsProjectSharingEnabled } from "@/lib/hooks";
-import { useExportProject, useImportProject } from "@/lib/api/projects/hooks";
+import { useExportProject, useImportProject, useFetchProjectsOnMount } from "@/lib/api/projects/hooks";
 
 export const ProjectsPage: React.FC = () => {
+    useFetchProjectsOnMount();
```
JKaram, line 19: project data comes from the ProjectsProvider. That Provider is wrapped around the App. Making it so it will not fetch projects until they are invalidated.

Using a useEffect is an anti pattern with React Query.

client/webui/frontend/src/lib/api/projects/hooks.ts
```diff
@@ -22,9 +22,13 @@ export function useProjects() {
     return useQuery({
         queryKey: projectKeys.lists(),
         queryFn: projectService.getProjects,
+        refetchOnMount: "always",
     });
 }
 
+/** Ensures projects are fetched fresh on component mount */
+export const useFetchProjectsOnMount = () => useProjects();
```
JKaram, line 30: I want the name to make it obvious why we call `useProjects()` on ProjectsPage.

lgh-solace (approved):

PR #1029, opened 2026-02-13
fix(DATAGO-123104): - Investigate and fix the issue with conversation history length in Solace Chat

enavitan:

Automatic Conversation History Summarization with Progressive Compression

```
 Implemented automatic conversation history summarization using Google ADK's `LlmEventSummarizer` with progressive compression. When context limits are hit, the system automatically:

  1. Summarizes the oldest conversation turns into a compact summary (uses preset % for compaction)
  2. Preserves recent context for continuity
  3. Retries the failed request each time applying same % compaction (on remaining context window)
  4. Notifies users that summarization occurred
```
```
  - Progressive Compression: Each new summary re-compresses (old summary + new turns), keeping total size bounded instead of growing infinitely
  - Parallel Task Coordination: Per-session locks prevent duplicate summarization when multiple tasks hit limits simultaneously
  - Automatic Ghost Event Filtering: Wrapper service automatically filters compacted events on all session loads
  - Audit Trail Preserved: Events stay in DB (append-only), only filtered in-memory
```

current task introduced 3 new env vars:
```
SAM_ENABLE_AUTO_SUMMARIZATION=true
SAM_COMPACTION_TRIGGER_TOKEN_LIMIT_THRESHOLD=5000 
SAM_COMPACTION_PERCENTAGE=0.50
```

 Ensure compaction functionality works as expected and doesn't break functionality

to test do:
```
run examples/agents/orchestrator_example.yaml examples/agents/a2a_agents_example.yaml examples/gateways/webui_gateway_example.yaml examples/services/platform_service_example.yaml examples/agents/a2a_multimodal_example.yaml
```

orchestrator_example.yaml - contains required cfg for autosummarisation pressent

```
    # default auto_summarization settings.
    auto_summarization: &default_auto_summarization
      # Enable/disable automatic conversation history summarization
      # Set to true to enable, false to disable (default: false)
      enabled: ${SAM_ENABLE_AUTO_SUMMARIZATION, false}

      # Percentage of conversation to compact when summarisation is triggered
      # Conservative default: 0.25 = 25%
      # Range: 0.0 - 1.0 (e.g., 0.25 = 25%, 0.5 = 50%, 0.8 = 80%)
      compaction_percentage: ${SAM_COMPACTION_PERCENTAGE, 0.25}
```

have .env - file:
```
SAM_ENABLE_AUTO_SUMMARIZATION=true
SAM_COMPACTION_PERCENTAGE=0.30

TEST_TOKEN_TRIGGER_THRESHOLD=1000 -- special env var used by unit/integration tests + mannual testing in order to force proactive comapction - to avoid burning thousands of tokens to trigger the overflow, this will never be set in prod.
```

### DIFF
--- a/src/solace_agent_mesh/agent/adk/runner.py
+++ b/src/solace_agent_mesh/agent/adk/runner.py
@@ -105,7 +105,9 @@ def _get_test_token_threshold() -> int:
     Returns:
         Token threshold value, or -1 if not set (test mode disabled)
     """
-    return int(os.getenv("TEST_TOKEN_TRIGGER_THRESHOLD", "-1"))
+    toke_trigger = int(os.getenv("TEST_TOKEN_TRIGGER_THRESHOLD", "-1"))
+    log.info(f"Proactive compaction enabled? {toke_trigger > 0} and value is: {toke_trigger} ")
+    return toke_trigger
 
 
 def _is_test_mode_trigger_enabled() -> bool:
--- a/examples/agents/a2a_agents_example.yaml
+++ b/examples/agents/a2a_agents_example.yaml
@@ -215,6 +215,7 @@ apps:
       artifact_handling_mode: "reference"
       enable_embed_resolution: true
       enable_artifact_content_instruction: true
+      auto_summarization: *default_auto_summarization
 
       agent_card:
         description: "An agent that fetches content from web URLs."


amir-ghasemi:
I have not looked at the code yet but I see the trigger is based on a character threshold
Given that different models have different context window sizes, a more robust approach would be defining the threshold based on a percentage of context window that is used e.g. summarize when crossing 80% of current model's context window. Would that be possible?

enavitan:
> I have not looked at the code yet but I see the trigger is based on a character threshold Given that different models have different context window sizes, a more robust approach would be defining the threshold based on a percentage of context window that is used e.g. summarize when crossing 80% of current model's context window. Would that be possible?

So  the character based trigger is added to proactively trigger compaction - for testing purposes (if value is -1 or 0, it is ignored). Now the percentage here plays the role of the compaction threshold, if its set to 0.3 - then
in case context window error is triggered (naturally by LLM), we will: 
 - count total nr of characters (this is our 100%) - in user session events.
 - apply 30% slicing (respecting user interaction boundaries, so we don't slice it in the middle of interaction) to oldest interactions.
 - compact above 30%/persist/reload (filtering out the 30% ghosts - events) retry last operation.
 
 Originally i took a more conservative approach, if we intercept BadRequest -  i'd summarise just first N user interactions) - ran this proposal thru the team and was suggested to apply % based approach where we compact certain % of the bloated window.

Hope it make sense.

amir-ghasemi:
> > I have not looked at the code yet but I see the trigger is based on a character threshold Given that different models have different context window sizes, a more robust approach would be defining the threshold based on a percentage of context window that is used e.g. summarize when crossing 80% of current model's context window. Would that be possible?
> 
> So the character based trigger is added to proactively trigger compaction - for testing purposes (if value is -1 or 0, it is ignored). Now the percentage here plays the role of the compaction threshold, if its set to 0.3 - then in case context window error is triggered (naturally by LLM), we will:
> 
> * count total nr of characters (this is our 100%) - in user session events.
> * apply 30% slicing (respecting user interaction boundaries, so we don't slice it in the middle of interaction) to oldest interactions.
> * compact above 30%/persist/reload (filtering out the 30% ghosts - events) retry last operation.
> 
> Originally i took a more conservative approach, if we intercept BadRequest - i'd summarise just first N user interactions) - ran this proposal thru the team and was suggested to apply % based approach where we compact certain % of the bloated window.
> 
> Hope it make sense.

One downside I see is that this seems to be a reactive approach I.e. we wait for the context window error before doing compaction. Given that at this point the context window could be fairly full, even with 30-50% compaction, the next LLM call could take a bit to process and will be blocking until we have the summary. If instead we trigger this proactively once we reach a threshold of the context window (instead of waiting for an error), we are more likely to be ready for the next user query. I believe this is how Claude and several other platforms behave as well.

Anyway, this is a thought. Your approach is still much better than existing behaviour :) happy to discuss further.

enavitan:
> > > I have not looked at the code yet but I see the trigger is based on a character threshold Given that different models have different context window sizes, a more robust approach would be defining the threshold based on a percentage of context window that is used e.g. summarize when crossing 80% of current model's context window. Would that be possible?
> > 
> > 
> > So the character based trigger is added to proactively trigger compaction - for testing purposes (if value is -1 or 0, it is ignored). Now the percentage here plays the role of the compaction threshold, if its set to 0.3 - then in case context window error is triggered (naturally by LLM), we will:
> > 
> > * count total nr of characters (this is our 100%) - in user session events.
> > * apply 30% slicing (respecting user interaction boundaries, so we don't slice it in the middle of interaction) to oldest interactions.
> > * compact above 30%/persist/reload (filtering out the 30% ghosts - events) retry last operation.
> > 
> > Originally i took a more conservative approach, if we intercept BadRequest - i'd summarise just first N user interactions) - ran this proposal thru the team and was suggested to apply % based approach where we compact certain % of the bloated window.
> > Hope it make sense.
> 
> One downside I see is that this seems to be a reactive approach I.e. we wait for the context window error before doing compaction. Given that at this point the context window could be fairly full, even with 30-50% compaction, the next LLM call could take a bit to process and will be blocking until we have the summary. If instead we trigger this proactively once we reach a threshold of the context window (instead of waiting for an error), we are more likely to be ready for the next user query. I believe this is how Claude and several other platforms behave as well.
> 
> Anyway, this is a thought. Your approach is still much better than existing behaviour :) happy to discuss further.

This would be fairly straightforward to achieve, all we would need to do is have a mapping of model -> [window size], then the proactive compaction would: "compare session size against current model" - if say ~90% mark is hit then compact oldest N %. 

But yes as you pointed current impl will be engaged only when we hit the critical window size so reactive rather than proactive, my main goal was to engage new compaction only when it is needed.

amir-ghasemi:
> > > > I have not looked at the code yet but I see the trigger is based on a character threshold Given that different models have different context window sizes, a more robust approach would be defining the threshold based on a percentage of context window that is used e.g. summarize when crossing 80% of current model's context window. Would that be possible?
> > > 
> > > 
> > > So the character based trigger is added to proactively trigger compaction - for testing purposes (if value is -1 or 0, it is ignored). Now the percentage here plays the role of the compaction threshold, if its set to 0.3 - then in case context window error is triggered (naturally by LLM), we will:
> > > 
> > > * count total nr of characters (this is our 100%) - in user session events.
> > > * apply 30% slicing (respecting user interaction boundaries, so we don't slice it in the middle of interaction) to oldest interactions.
> > > * compact above 30%/persist/reload (filtering out the 30% ghosts - events) retry last operation.
> > > 
> > > Originally i took a more conservative approach, if we intercept BadRequest - i'd summarise just first N user interactions) - ran this proposal thru the team and was suggested to apply % based approach where we compact certain % of the bloated window.
> > > Hope it make sense.
> > 
> > 
> > One downside I see is that this seems to be a reactive approach I.e. we wait for the context window error before doing compaction. Given that at this point the context window could be fairly full, even with 30-50% compaction, the next LLM call could take a bit to process and will be blocking until we have the summary. If instead we trigger this proactively once we reach a threshold of the context window (instead of waiting for an error), we are more likely to be ready for the next user query. I believe this is how Claude and several other platforms behave as well.
> > Anyway, this is a thought. Your approach is still much better than existing behaviour :) happy to discuss further.
> 
> This would be fairly straightforward to achieve, all we would need to do is have a mapping of model -> [window size], then the proactive compaction would: "compare session size against current model" - if say ~90% mark is hit then compact oldest N %.
> 
> But yes as you pointed current impl will be engaged only when we hit the critical window size so reactive rather than proactive, my main goal was to engage new compaction only when it is needed.

I have an older branch with some the mapping stuff (LLM -> context windows size) which might be of use (amir/token-tracking). I think a hybrid would make sense where we compact at say 80% proactively and then catch the remaining failures using the above reactive approach as even with proactive approach you could have situations where a single large message could push one above the threshold and result in error.

enavitan:
> > > > > I have not looked at the code yet but I see the trigger is based on a character threshold Given that different models have different context window sizes, a more robust approach would be defining the threshold based on a percentage of context window that is used e.g. summarize when crossing 80% of current model's context window. Would that be possible?
> > > > 
> > > > 
> > > > So the character based trigger is added to proactively trigger compaction - for testing purposes (if value is -1 or 0, it is ignored). Now the percentage here plays the role of the compaction threshold, if its set to 0.3 - then in case context window error is triggered (naturally by LLM), we will:
> > > > 
> > > > * count total nr of characters (this is our 100%) - in user session events.
> > > > * apply 30% slicing (respecting user interaction boundaries, so we don't slice it in the middle of interaction) to oldest interactions.
> > > > * compact above 30%/persist/reload (filtering out the 30% ghosts - events) retry last operation.
> > > > 
> > > > Originally i took a more conservative approach, if we intercept BadRequest - i'd summarise just first N user interactions) - ran this proposal thru the team and was suggested to apply % based approach where we compact certain % of the bloated window.
> > > > Hope it make sense.
> > > 
> > > 
> > > One downside I see is that this seems to be a reactive approach I.e. we wait for the context window error before doing compaction. Given that at this point the context window could be fairly full, even with 30-50% compaction, the next LLM call could take a bit to process and will be blocking until we have the summary. If instead we trigger this proactively once we reach a threshold of the context window (instead of waiting for an error), we are more likely to be ready for the next user query. I believe this is how Claude and several other platforms behave as well.
> > > Anyway, this is a thought. Your approach is still much better than existing behaviour :) happy to discuss further.
> > 
> > 
> > This would be fairly straightforward to achieve, all we would need to do is have a mapping of model -> [window size], then the proactive compaction would: "compare session size against current model" - if say ~90% mark is hit then compact oldest N %.
> > But yes as you pointed current impl will be engaged only when we hit the critical window size so reactive rather than proactive, my main goal was to engage new compaction only when it is needed.
> 
> I have an older branch with some the mapping stuff (LLM -> context windows size) which might be of use (amir/token-tracking). I think a hybrid would make sense where we compact at say 80% proactively and then catch the remaining failures using the above reactive approach as even with proactive approach you could have situations where a single large message could push one above the threshold and result in error.

think its doable, leme try adding this bit.

enavitan:
> > > > > > I have not looked at the code yet but I see the trigger is based on a character threshold Given that different models have different context window sizes, a more robust approach would be defining the threshold based on a percentage of context window that is used e.g. summarize when crossing 80% of current model's context window. Would that be possible?
> > > > > 
> > > > > 
> > > > > So the character based trigger is added to proactively trigger compaction - for testing purposes (if value is -1 or 0, it is ignored). Now the percentage here plays the role of the compaction threshold, if its set to 0.3 - then in case context window error is triggered (naturally by LLM), we will:
> > > > > 
> > > > > * count total nr of characters (this is our 100%) - in user session events.
> > > > > * apply 30% slicing (respecting user interaction boundaries, so we don't slice it in the middle of interaction) to oldest interactions.
> > > > > * compact above 30%/persist/reload (filtering out the 30% ghosts - events) retry last operation.
> > > > > 
> > > > > Originally i took a more conservative approach, if we intercept BadRequest - i'd summarise just first N user interactions) - ran this proposal thru the team and was suggested to apply % based approach where we compact certain % of the bloated window.
> > > > > Hope it make sense.
> > > > 
> > > > 
> > > > One downside I see is that this seems to be a reactive approach I.e. we wait for the context window error before doing compaction. Given that at this point the context window could be fairly full, even with 30-50% compaction, the next LLM call could take a bit to process and will be blocking until we have the summary. If instead we trigger this proactively once we reach a threshold of the context window (instead of waiting for an error), we are more likely to be ready for the next user query. I believe this is how Claude and several other platforms behave as well.
> > > > Anyway, this is a thought. Your approach is still much better than existing behaviour :) happy to discuss further.
> > > 
> > > 
> > > This would be fairly straightforward to achieve, all we would need to do is have a mapping of model -> [window size], then the proactive compaction would: "compare session size against current model" - if say ~90% mark is hit then compact oldest N %.
> > > But yes as you pointed current impl will be engaged only when we hit the critical window size so reactive rather than proactive, my main goal was to engage new compaction only when it is needed.
> > 
> > 
> > I have an older branch with some the mapping stuff (LLM -> context windows size) which might be of use (amir/token-tracking). I think a hybrid would make sense where we compact at say 80% proactively and then catch the remaining failures using the above reactive approach as even with proactive approach you could have situations where a single large message could push one above the threshold and result in error.
> 
> think its doable, leme try adding this bit.

@amir-ghasemi - I propose we defer this approach - the code modifications introduce a bit of  complexity - we now need to exactly match a model then fallback to matching - falling back to existing interceptor. Current fallback approach should protect us for most cases, we will compact existing events, for up to 3 times each time reducing history by same % of remaining window - so it should shrink aggressively - yes this would be slow - but so would be proactive compaction - the only difference we would do it only when we hit the limit - which is probably 1-5% of all interactions. WDYT?

amir-ghasemi:
> > > > > > > I have not looked at the code yet but I see the trigger is based on a character threshold Given that different models have different context window sizes, a more robust approach would be defining the threshold based on a percentage of context window that is used e.g. summarize when crossing 80% of current model's context window. Would that be possible?
> > > > > > 
> > > > > > 
> > > > > > So the character based trigger is added to proactively trigger compaction - for testing purposes (if value is -1 or 0, it is ignored). Now the percentage here plays the role of the compaction threshold, if its set to 0.3 - then in case context window error is triggered (naturally by LLM), we will:
> > > > > > 
> > > > > > * count total nr of characters (this is our 100%) - in user session events.
> > > > > > * apply 30% slicing (respecting user interaction boundaries, so we don't slice it in the middle of interaction) to oldest interactions.
> > > > > > * compact above 30%/persist/reload (filtering out the 30% ghosts - events) retry last operation.
> > > > > > 
> > > > > > Originally i took a more conservative approach, if we intercept BadRequest - i'd summarise just first N user interactions) - ran this proposal thru the team and was suggested to apply % based approach where we compact certain % of the bloated window.
> > > > > > Hope it make sense.
> > > > > 
> > > > > 
> > > > > One downside I see is that this seems to be a reactive approach I.e. we wait for the context window error before doing compaction. Given that at this point the context window could be fairly full, even with 30-50% compaction, the next LLM call could take a bit to process and will be blocking until we have the summary. If instead we trigger this proactively once we reach a threshold of the context window (instead of waiting for an error), we are more likely to be ready for the next user query. I believe this is how Claude and several other platforms behave as well.
> > > > > Anyway, this is a thought. Your approach is still much better than existing behaviour :) happy to discuss further.
> > > > 
> > > > 
> > > > This would be fairly straightforward to achieve, all we would need to do is have a mapping of model -> [window size], then the proactive compaction would: "compare session size against current model" - if say ~90% mark is hit then compact oldest N %.
> > > > But yes as you pointed current impl will be engaged only when we hit the critical window size so reactive rather than proactive, my main goal was to engage new compaction only when it is needed.
> > > 
> > > 
> > > I have an older branch with some the mapping stuff (LLM -> context windows size) which might be of use (amir/token-tracking). I think a hybrid would make sense where we compact at say 80% proactively and then catch the remaining failures using the above reactive approach as even with proactive approach you could have situations where a single large message could push one above the threshold and result in error.
> > 
> > 
> > think its doable, leme try adding this bit.
> 
> @amir-ghasemi - I propose we defer this approach - the code modifications introduce a bit of complexity - we now need to exactly match a model then fallback to matching - falling back to existing interceptor. Current fallback approach should protect us for most cases, we will compact existing events, for up to 3 times each time reducing history by same % of remaining window - so it should shrink aggressively - yes this would be slow - but so would be proactive compaction - the only difference we would do it only when we hit the limit - which is probably 1-5% of all interactions. WDYT?

Sounds good. We can always revisit later once we see this in production. The reactive approach is way better than the current situation. One thing we might consider is to show how much of the context has been filled and show a component in the UI so that user can choose to proactively compact themselves. I already have this bit implemented and approved UX-wise in my branch. Might be worth considering as an add-on on top of yours. But even that can wait until this is merged.

src/solace_agent_mesh/agent/adk/runner.py
```diff
@@ -21,17 +21,880 @@ class TaskCancelledError(Exception):
...
+    """
+    if not events:
+        return 0
+
+    total_chars = 0
+    for event in events:
+        if event.content and event.content.parts:
+            for part in event.content.parts:
+                if part.text:
+                    total_chars += len(part.text)
```
gregmeldrum: Are there other `part` types that would count towards the context. I'm thinking of inline images and function calls as examples.

enavitan: good point! I assumed artifacts are stored separately and sessions is kept free of these, but appears we do have binary data in sessions, leme fix it.

enavitan: @gregmeldrum - this has been updated, i've used exact logic we use when building content for LiteLLM, however token_counter won't work on video format - so it is approximated.

src/solace_agent_mesh/agent/adk/runner.py
```diff
@@ -128,14 +1058,125 @@ async def run_adk_async_task_thread_wrapper(
...
+                            "%s Proactive compaction triggered: total_tokens=%d exceeds threshold=%d",
+                            component.log_identifier,
+                            total_tokens,
+                            token_threshold
+                        )
+                        raise BadRequestError(
+                            message=f"Too many tokens: {total_tokens} tokens exceed token limit {token_threshold} (proactive compaction triggered)",
+                            model="test-model",
+                            llm_provider="test-provider"
+                        )
```
efunneko: Raising a fake `BadRequestError` with hardcoded `model="test-model"` and `llm_provider="test-provider"` is using exceptions for control flow and could confuse logging/debugging downstream.

Is there now way to do this differently?

src/solace_agent_mesh/agent/adk/runner.py
```diff
@@ -21,17 +22,940 @@ class TaskCancelledError(Exception):
...
+# Per-session locks for compaction to prevent parallel tasks from duplicate summarization
+# When multiple tasks hit context limit simultaneously, only one compacts per session
+_compaction_locks: TTLCache = TTLCache(maxsize=10000, ttl=3600)
+_compaction_locks_mutex = asyncio.Lock()
+
+# Per-session summaries for deferred notification (after successful response)
+# When compaction occurs during retries, we store the summary here instead of sending immediately
+# This ensures users see the actual response first, then a clean notification about summarization
+# Dict operations are atomic in asyncio (single event loop), so no mutex needed
+_session_summaries: dict[str, str] = {}  # session_id → latest summary text
```
efunneko: These module-level globals (`_compaction_locks: TTLCache`, `_session_summaries: dict`) are shared across all agent instances in the process:

- `_session_summaries` is a plain dict that grows unboundedly if sessions complete but summaries are never consumed (e.g., if finalization fails before the `pop` at line 1213).
- If multiple agent components share a process, session IDs from different agents could theoretically collide (though unlikely with UUIDs).

Consider making these per-component (e.g., instance attributes on `SamAgentComponent`) rather than module-level globals. At minimum, `_session_summaries` should have some eviction strategy (TTLCache like `_compaction_locks`, or periodic cleanup) to prevent unbounded growth.

src/solace_agent_mesh/agent/adk/runner.py
```diff
@@ -21,17 +22,940 @@ class TaskCancelledError(Exception):
...
+    Args:
+        session_id: The ADK session ID
+
+    Returns:
+        asyncio.Lock instance for this session
+    """
+    async with _compaction_locks_mutex:
+        if session_id not in _compaction_locks:
+            _compaction_locks[session_id] = asyncio.Lock()
+        return _compaction_locks[session_id]
```
efunneko: The `_compaction_locks_mutex` protects lock creation, but `TTLCache` eviction happens automatically when entries expire (TTL=3600s). If a lock is evicted while a compaction is still in progress (unlikely given compaction takes <60s, but possible under extreme conditions), a parallel task calling `_get_compaction_lock` would create a **new** lock for the same session ID, bypassing the coordination and potentially starting a duplicate compaction.

The 1-hour TTL is generous enough that this is unlikely in practice, but worth documenting as an assumption. Alternatively, consider using a plain dict with explicit cleanup (e.g., remove the lock after compaction completes) for stronger guarantees.

src/solace_agent_mesh/agent/adk/runner.py
```diff
@@ -21,17 +22,940 @@ class TaskCancelledError(Exception):
...
+
+    # OPTIMIZATION: Pre-calculate tokens for all events once (O(N) with N token_counter calls)
+    import time
+
+    start_time = time.time()
+    event_tokens = []
+    for event in events:
+        if event.content:
+            try:
+                tokens = _calculate_content_tokens(event.content, model=model)
```
efunneko, line 278: Is it possible to use the token counts we get back from previous requests to use for token count? Then we don't have to do this work and we can be sure of the counts

efunneko: Also, using a fixed model is not really accurate

src/solace_agent_mesh/agent/adk/runner.py
```diff
@@ -21,17 +22,940 @@ class TaskCancelledError(Exception):
...
+        raise  # Fail retry if we can't persist
+
+    # Log compression stats
+    log.info(
+        "%s Compacted %d events into summary (%d tokens → ~%d tokens, ~%dx compression)",
+        log_identifier,
+        len(events_to_compact),
+        sum(len(str(e.content)) for e in events_to_compact if e.content) // 4,
+        len(summary_text) // 4,
+        max(1, sum(len(str(e.content)) for e in events_to_compact if e.content) // max(1, len(summary_text)))
```
efunneko: We are doing an approximation for tokens here even though we did a calculatoin

src/solace_agent_mesh/agent/adk/runner.py
enavitan: This bit here is for testing purposes and aim to trigger same exact execution flow be it manual/proactive trigger or the natural one triggered by LLM. the above values were just simple mocks since BadRequestError expects these 3 arguments. I see 2 solutions here:
- remove this logic/drop token_treshold_limit_trigger
- access model/llm_provider from the actual model we use, but still keep exception throwing as is + add an LOG.warn("Proactive trigger") WDYT?

src/solace_agent_mesh/agent/adk/runner.py
enavitan: good point on converting _session_summaries to TTLCache - indeed this would protect us from cases when else block never ran.

my main concern in moving _session_summaries into the SamAgentComponent - is the scope, by having this struct now part of the component everyone has access to it and thus opens a question why is it there who else wants to access it  and use it, in current state the _session_summaries is encapsulated as part of the runner and the scope is limited to this logical block. I propose to keep here WDYT?

src/solace_agent_mesh/agent/adk/runner.py
enavitan: good point, my assumption was ttl will be reset on next access, but reading thru pydoc its not the case, how about instead we do:
```
    async with _compaction_locks_mutex:
        if session_id not in _compaction_locks:
            _compaction_locks[session_id] = asyncio.Lock()
        else:
            # Re-insert to reset TTL on access (idle timeout behavior)
            lock = _compaction_locks.pop(session_id)
            _compaction_locks[session_id] = lock
        return _compaction_locks[session_id]
```
i.e we force put same lock thus ensuring ttl is reset on each access.

src/solace_agent_mesh/agent/adk/runner.py
enavitan: > Is it possible to use the token counts we get back from previous requests to use for token count? Then we don't have to do this work and we can be sure of the counts

yup, this makes perfect sense!

enavitan: > Also, using a fixed model is not really accurate

right, main reason was i am ok with approximation since we just need a way to find total token count then certain % of it, as long as both these calculation are consistent we are good. this was one of the impls which i reverted to keep it simple, i will revert it to and use the underlying models model.

src/solace_agent_mesh/agent/adk/runner.py
enavitan: nice catch!

enavitan:
@efunneko - would you mind taking a second look ?

Thanks
Eugene

src/solace_agent_mesh/agent/adk/runner.py
efunneko: I am not following you. This code is for test only? I don't understand how it would be triggered in test and not in normal operation, but perhaps I am not looking at the bigger picture enough

src/solace_agent_mesh/agent/adk/runner.py
efunneko: There is a very good chance that we might move to creating many agents within the same process. This is actually done right now, but due to how we are dynamically importing modules, module level vars are separate from each other. But if we move to a more efficient way of doing this, these module level vars would be shared between agents. It is much cleaner to just keep them in the object and be sure that if we create two agents, they are isolated

src/solace_agent_mesh/agent/adk/runner.py
enavitan: correct, in order to be able to engage with above proactive trigger (rather than wait for the context window exhaustion) flow you will need:
```
SAM_ENABLE_AUTO_SUMMARIZATION=true
SAM_COMPACTION_TRIGGER_TOKEN_LIMIT_THRESHOLD=1000. # this one will ensure the above check will be engaged.
SAM_COMPACTION_PERCENTAGE=0.30
```

correction: 
SAM_COMPACTION_TRIGGER_TOKEN_LIMIT_THRESHOLD - is used to force smaller context window and trigger compaction in:
- unit tests
- integration tests
- manual testing to hit limit earlier 
all scenarios above will go thru the same handler, in prod we will NOT export SAM_COMPACTION_TRIGGER_TOKEN_LIMIT_THRESHOLD - and as such proactive compaction will not be engaged

src/solace_agent_mesh/agent/adk/runner.py
efunneko: Yes, this will work

src/solace_agent_mesh/agent/adk/runner.py
efunneko: As long as we can use the counts coming back from the LLM request, then we are good and don't have to worry about doing any token counting. Make sure to include both cached and non-cached tokens

src/solace_agent_mesh/agent/adk/runner.py
enavitan: ok, moving above structs into the component.

enavitan: grouped compaction related states into 1 aggregated component which is now part of SamAgentComponent.

src/solace_agent_mesh/agent/adk/runner.py
```diff
@@ -128,14 +1012,128 @@ async def run_adk_async_task_thread_wrapper(
...
+        retry_count = 0
+        is_paused = False
+
+        while retry_count <= max_retries:
+            try:
+                # Proactively trigger compaction when token count exceeds threshold
+                if compaction_enabled and token_threshold > 0 and adk_session.events and adk_content.role == 'user':
+                    total_tokens = _calculate_session_context_tokens(adk_session.events, model=str(component.adk_agent.model))
+                    log.info(
+                        "%s Proactive compaction check: total_tokens=%d, threshold=%d, exceeds=%s",
```
efunneko: I am not sure this info log is needed. It will be very busy since it will happen many times per task

src/solace_agent_mesh/agent/adk/runner.py
efunneko: I am okay with this, but I think you should put a clear comment that this code is raising the same error that the LLM would have raised if you hit the context limit. It looks like it is an error path, but it is really normal operation if you have the threshold set

src/solace_agent_mesh/agent/adk/runner.py
```diff
@@ -21,17 +22,894 @@ class TaskCancelledError(Exception):
...
+    # OPTIMIZATION: Pre-calculate tokens and build cumulative sum in single pass (O(N))
+    import time
+
+    start_time = time.time()
+    cumulative_tokens = [0]
+    for event in events:
+        if event.content:
+            try:
+                tokens = _calculate_content_tokens(event.content, model=model)
+            except Exception as e:
```
efunneko, line 279: I am not sure that you understood my comment here. I wanted to know if you could extract the exact token counts that the LLM returns to us. Then we wouldn't have do do this calculation

enavitan: as discussed, will alter code to explicitly outline - this is just for testing never to be used in prod.

enavitan:
@efunneko - as discussed:
- removed proactive  auto compaction from config, now its env var dependend
- improved readability around trigger to outline this is testing only. 

Thank you
Eugene

solace-mdupls:
By-passing the PR-size check as this is a complex change which evolved to a larger than anticipated change during development. Test coverage is sufficient.

PR #1032, opened 2026-02-13
fix(DATAGO-122505): Handle revoked MCP tool credentials by triggering re-auth

gregmeldrum:
## Summary
- Detect 401/403 auth errors in MCP tool responses after execution (the ADK only checks credentials pre-execution, missing revoked tokens)
- Clear stale credentials from both in-memory cache and credential service
- Trigger re-authentication via `request_credential()` so the user is seamlessly redirected to authenticate again without manual retry

## Test plan
- [ ] Configure an MCP tool with OAuth authentication
- [ ] Authenticate and make a successful tool call
- [ ] Revoke the token on the OAuth provider side
- [ ] Make another tool call — should trigger re-auth flow instead of returning a raw 401 error
- [ ] Complete re-auth — tool call should succeed with the fresh token

### DIFF
--- a/src/solace_agent_mesh/agent/adk/embed_resolving_mcp_toolset.py
+++ b/src/solace_agent_mesh/agent/adk/embed_resolving_mcp_toolset.py
@@ -386,9 +386,10 @@ class EmbedResolvingMCPTool(_BaseMcpToolClass):
 
         try:
             await tool_context.save_credential(auth_config)
-        except ValueError:
+        except Exception:
             log.debug(
-                "No credential service available to clear persisted credential.",
+                "Failed to clear persisted credential (best-effort).",
+                exc_info=True,
             )
 
     async def _execute_tool_with_audit_logs(self, tool_call, tool_context):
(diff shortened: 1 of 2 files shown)


src/solace_agent_mesh/agent/adk/embed_resolving_mcp_toolset.py
```diff
@@ -336,6 +335,62 @@ async def _resolve_embeds_recursively(
...
+        auth_config.exchanged_auth_credential = None
+
+        # Also clear the wrapped tool's credential manager cache
+        original_cm = getattr(self._original_mcp_tool, "_credentials_manager", None)
+        if original_cm and hasattr(original_cm, "_auth_config"):
+            original_cm._auth_config.exchanged_auth_credential = None
+
+        try:
+            await tool_context.save_credential(auth_config)
+        except ValueError:
```
enavitan: should we be more generic here and intercept any kind of errors ? ValueError - assumes we landed on the no credential service - but what if error comes from service itself?

gregmeldrum: Good catch — this was indeed too narrow. I've broadened it to `except Exception` since credential clearing is a best-effort operation. If `save_credential` fails for any reason (not just missing credential service), we should still proceed with the re-auth flow rather than letting the exception propagate and break it. Added a test for non-ValueError exceptions as well. See 6b11283e.

PR #1033, opened 2026-02-13
fix(DATAGO-122310): Character limit for project file descriptions

shanechen-solace:

Artifact file descriptions had no length enforcement, meaning any length string was accepted and stored. This PR adds backend validation to reject descriptions exceeding 1000 characters, matching the limit already exposed to the frontend via the config endpoint.

**Backend:** Added `_validate_file_descriptions()` to `ProjectService`, called in all three write paths: `create_project()`, `add_artifacts_to_project()`, and `update_artifact_metadata()`. Added`DEFAULT_MAX_PROJECT_FILE_DESCRIPTION_LENGTH` constant as single source of truth.

**Frontend:** Refactored project file-related dialogs (`AddProjectFilesDialog.tsx` and `EditFileDescriptionDialog.tsx`) using `form.tsx` and the existing `react-hook-form` dependency. Added Client-side enforcement character limits. Two new Storybook stories were added to cover the over-limit validation state for both dialogs.

### Key Design Decisions

Service-layer validation was chosen over Pydantic DTOs because the POST endpoints receive descriptions embedded in a `file_metadata` JSON string with dynamic keys (filenames), which doesn't map cleanly to static Pydantic fields. The PATCH endpoint could use a DTO, but since the service-layer check already covers it and the constant already exists as a single source of truth, a DTO would add indirection with no practical benefit.

- [X] Manual testing: Submit descriptions over/at 1000 chars via create project, add artifacts, and update artifact metadata endpoints
- [X] Integration tests: New `test_projects_description_limits.py` with boundary tests (1000 chars accepted, 1001 chars rejected)
- [X] Storybook tests: New `DescriptionOverLimit` stories for `AddProjectFilesDialog` and `EditFileDescriptionDialog` — paste 1001 chars, assert error message and disabled submit button

### Demo

https://github.com/user-attachments/assets/ac7eb776-b27e-4f60-ad12-6d5a2a2a9d7d

### DIFF
--- a/client/webui/frontend/src/lib/components/projects/KnowledgeSection.tsx
+++ b/client/webui/frontend/src/lib/components/projects/KnowledgeSection.tsx
@@ -5,7 +5,7 @@ import { useConfigContext, useDownload, useIsProjectOwner } from "@/lib/hooks";
 import { useProjectArtifacts } from "@/lib/api/projects/hooks";
 import { useProjectContext } from "@/lib/providers";
 import type { ArtifactInfo, Project } from "@/lib/types";
-import { formatRelativeTime, validateFileSizes, validateBatchUploadSize, validateProjectSizeLimit, calculateTotalFileSize } from "@/lib/utils";
+import { validateFileSizes, validateBatchUploadSize, validateProjectSizeLimit, calculateTotalFileSize } from "@/lib/utils";
 
 import { ArtifactBar } from "../chat/artifact";
 import { FileDetails } from "../chat/file";
@@ -200,7 +200,7 @@ export const KnowledgeSection: React.FC<KnowledgeSectionProps> = ({ project }) =
                                             <ArtifactBar
                                                 key={artifact.filename}
                                                 filename={artifact.filename}
-                                                description={artifact.description || formatRelativeTime(artifact.last_modified)}
+                                                description={artifact.description || "No description provided"}
                                                 mimeType={artifact.mime_type}
                                                 size={artifact.size}
                                                 status="completed"


client/webui/frontend/src/lib/components/projects/AddProjectFilesDialog.tsx
```diff
@@ -15,61 +20,95 @@ interface AddProjectFilesDialogProps {
...
+
+                            return (
+                                <FormField
+                                    key={fieldName}
+                                    control={form.control}
+                                    name={fieldName}
+                                    rules={{
+                                        maxLength: {
+                                            value: FILE_DESCRIPTION_MAX,
+                                            message: `Description must not exceed ${FILE_DESCRIPTION_MAX} characters`,
```
lgh-solace: Could you check around and verify if this is a consistent message? I think I've seen `Blah exceeds the maximum...`?

client/webui/frontend/src/lib/components/projects/AddProjectFilesDialog.tsx
```diff
@@ -15,61 +20,95 @@ interface AddProjectFilesDialogProps {
...
+                                        },
+                                    }}
+                                    render={({ field, fieldState }) => (
+                                        <FormItem>
+                                            <FileLabel fileName={file.name} fileSize={file.size} />
+                                            <FormControl>
+                                                <Textarea {...field} rows={2} disabled={isSubmitting} maxLength={FILE_DESCRIPTION_MAX + 1} className={cn("mt-2 resize-none text-sm", fieldState.error && "border-destructive")} />
```
lgh-solace: There's also some `invalid` support in the Textarea component - it may need fixing, can you take a look? If/when working you won't need to pass in the border this way but just set the field as invalid.

client/webui/frontend/src/lib/components/ui/form.tsx
```diff
@@ -80,19 +83,30 @@ function FormDescription({ className, ...props }: React.ComponentProps<"p">) {
...
     return (
-        <p data-slot="form-message" id={formMessageId} className={cn("text-destructive text-sm", className)} {...props}>
+        <p data-slot="form-error" id={formMessageId} className={cn("text-xs text-(--color-error-wMain)", className)} {...props}>
             {body}
         </p>
     );
 }
 
-export { useFormField, Form, FormItem, FormLabel, FormControl, FormDescription, FormMessage, FormField };
+function FormMessage({ className, ...props }: React.ComponentProps<"p">) {
```
lgh-solace: Could this be more useful if one of the props took in some kind of rightAlign prop? We could call it `FormInputLabel` and use it more generally?

client/webui/frontend/src/lib/components/projects/AddProjectFilesDialog.tsx
shanechen-solace: I want the phrase "{Form} exceeds the maximum of {LIMIT} characters" (ex. "Description exceeds the maximum of 1000 characters") to be the standard. Other forms do have different messages right now, but changing that is for another PR. So for a refactor for something like the edit project instructions dialog, we would 1. use the new `form.tsx` component and 2. update the error message to be standard. For writing to a project file description, however, this will be the current standard.

client/webui/frontend/src/lib/components/ui/form.tsx
shanechen-solace: Renamed to FormInputLabel and added a rightAlign prop.

lgh-solace:
Thanks for the video! Looking at it, there's something odd happening to the description in the page while editing it. I see 0s then 2s in the artifact list. Can you check this issue out? Is it something pre-existing, I don't think I've ever seen it before... 

image

lgh-solace (approved):
Thanks, please wait for Rudraneel's approval before merging.

PR #1037, opened 2026-02-15
chore(DATAGO-125491): add ability to scan task contexts by predicate

efunneko:
Adds a `scan_contexts` method to `TaskContextManager` that allows scanning stored contexts using a predicate function. This is a foundational utility needed for the deferred Solace event acknowledgment feature (DATAGO-125491), which requires the ability to look up task contexts by criteria other than task ID.

Added a new `scan_contexts(predicate)` method to `TaskContextManager` in `src/solace_agent_mesh/gateway/base/task_context.py`. The method takes a callable predicate that receives `(task_id, context_data)` and returns matching `(task_id, context_data)` tuples. The scan is performed under the existing `_lock` to ensure thread safety.

- [x] Manual testing: Verified the method works with various predicate functions
- [x] Unit tests in the Event Mesh Gateway test suite

This is a small, self-contained utility addition. The method is thread-safe (uses the existing lock) and returns a copy of matching entries, so callers can safely iterate without holding the lock.

### DIFF
--- a/src/solace_agent_mesh/gateway/base/task_context.py
+++ b/src/solace_agent_mesh/gateway/base/task_context.py
@@ -4,7 +4,7 @@ Manages context for tasks being processed by a gateway.
 
 import logging
 import threading
-from typing import Dict, Optional, Any
+from typing import Callable, Dict, List, Optional, Any, Tuple
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


efunneko:
Superseded by #1145

PR #1042, opened 2026-02-17
chore(DATAGO-125203): Pin pillow package version to 12.1.1

Hugo-Pare:

Addresses a security vulnerability in the Pillow package by upgrading from version 10.2.0 to 12.1.1.

Added `pillow==12.1.1` as a direct dependency in `pyproject.toml`. This overrides the vulnerable transitive dependency from `python-pptx`, ensuring the secure version is installed across all environments.

### Key Design Decisions

Used an exact pin (`==12.1.1`) rather than a minimum version constraint to ensure consistent, deterministic builds and explicit control over security updates. Added an inline comment explaining this is a security fix for a transitive dependency to provide context for future maintainers.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ dependencies = [
     "psycopg2-binary==2.9.10",
     "azure-cognitiveservices-speech==1.41.1",
     "authlib==1.6.5",  # For CVE-2025-59420: May be removed if next update to google-adk upgrades its transitive dependency on authlib
-    "pillow==12.1.1",  # Security fix: addresses vulnerabilities in older versions (transitive dependency from python-pptx)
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -72,6 +72,7 @@ dependencies = [
     "psycopg2-binary==2.9.10",
     "azure-cognitiveservices-speech==1.41.1",
     "authlib==1.6.5",  # For CVE-2025-59420: May be removed if next update to google-adk upgrades its transitive dependency on authlib
+    "pillow==12.1.1",  # Security fix: addresses vulnerabilities in older versions (transitive dependency from python-pptx)
```
enavitan: any chances u could list what exact CVE it fixes? in no time the above comment will be irrelevant - due to new version for same lib.

PR #1054, opened 2026-02-20
fix(DATAGO-123916): fixing no content parse for project file deletion

lgh-solace:

Removing effort to parse a project file deletion response - that was causing refresh of the artifact list to fail.

This pull request makes a small change to the `removeFileFromProject` function in `service.ts`. The change simplifies the return value by removing unnecessary data from the response and now only returns the `sseLocation` header.

None

### DIFF
--- a/client/webui/frontend/src/lib/api/projects/service.ts
+++ b/client/webui/frontend/src/lib/api/projects/service.ts
@@ -46,8 +46,7 @@ export const removeFileFromProject = async (projectId: string, filename: string)
         throw new Error(await getErrorFromResponse(response));
     }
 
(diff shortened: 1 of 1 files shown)


client/webui/frontend/src/lib/api/projects/service.ts
```diff
@@ -46,8 +46,7 @@ export const removeFileFromProject = async (projectId: string, filename: string)
         throw new Error(await getErrorFromResponse(response));
     }
 
-    const result = await response.json();
```
lgh-solace, line 49: It's no content response, so cannot parse it.

PR #1066, opened 2026-02-23
fix(DATAGO-126173): adding dynamic "index_search" injection for project usage

lgh-solace:

The goal: make  the new tool `index_search` available to any agent when the user is in a project chat, without requiring explicit YAML configuration.

## How it works:
When a user is in a project chat, the gateway passes project_id in the request metadata (gated on project_indexing.enabled). A new _inject_project_tools_callback in the agent's before_model chain detects this metadata and dynamically injects index_search into the LLM's available tools — no YAML configuration required on the agent side.

## Data flow:
* Gateway (tasks.py) adds project_id to additional_metadata when project_indexing.enabled is true
* Metadata flows through Solace messaging → a2a_context["original_message_metadata"]
* _inject_project_tools_callback reads project_id from metadata, looks up index_search in the tool registry, wraps it via FunctionTool(ADKToolWrapper(...)), and adds it to tools_dict + config.tools
* Tool instructions are stored in callback state for inject_dynamic_instructions_callback to pick up
* Callback chain ordering: peer tools → project tools (new) → capability filter → dynamic instructions

## Early-return guards
* No project_id in metadata → skip
* index_search already in tools_dict (explicit YAML config) → skip
* index_search not in tool registry (bm25s not installed) → skip

## Known limitations
* HTTP/SSE gateway only. The project_id metadata is added in the HTTP/SSE gateway router (tasks.py). Requests arriving through the generic gateway, direct Solace messaging, or other interfaces won't have project_id — the tool won't be injected for those paths. Acceptable since projects are currently a web UI feature.

* Peer delegation doesn't propagate project context. When the orchestrator delegates to a peer agent, the peer's a2a_context is freshly constructed without original_message_metadata. Only the agent directly handling the gateway request gets index_search injected — downstream peers do not.

* Callback ordering is implicit. The correct position (after peer tools, before capability filter) is enforced only by code placement in setup.py. No runtime ordering validation exists — a future refactor of the chain could silently break this.

* FunctionTool wrapping is necessary because tools_dict expects BaseTool-compatible objects. Other builtin tools go through load_adk_tools → AppLlmAgent which handles this wrapping. Since we inject after initialization, we wrap ADKToolWrapper in FunctionTool manually and set origin="builtin" for capability filtering.

## Alternatives considered

* Auto-include index_search in _load_internal_tools at startup. Rejected because project association is per-request (resolved from message metadata), not per-agent. _load_internal_tools runs once at init time with no project context — every agent would get index_search unconditionally, including non-project chats.

* Require explicit YAML config on each agent. This is the status quo and the problem we're solving. Users would need to add index_search to every agent's tools list, and it would always be present even outside project chats.

* Inject at the gateway/protocol layer before dispatching to the agent. Rejected because the gateway doesn't have access to the LLM tool chain — tool injection needs to happen within the agent's before_model callback where tools_dict and config.tools are available. The gateway can only pass metadata.

* Use BaseTool subclass directly instead of FunctionTool(ADKToolWrapper(...)) wrapping. Rejected because ADKToolWrapper already handles embed resolution, artifact pre-loading, config injection, and error boundaries. Writing a new BaseTool subclass would duplicate all of that. The FunctionTool wrapper adds the BaseTool interface (name, description, is_long_running, run_async) that tools_dict requires.

For manual testing with current limitations, see: https://miro.com/app/board/uXjVJgx7by0=/?moveToWidget=3458764660940483603&cot=14


examples/gateways/webui_gateway_example.yaml
```diff
@@ -112,6 +112,7 @@ apps:
       frontend_feature_enablement:
         background_tasks: ${BACKGROUND_TASKS_ENABLED, true} # NOTE: task_logging must also be enabled for background tasks to work
         auto_title_generation: ${AUTO_TITLE_GENERATION_ENABLED, true}  # Enable automatic chat title generation
+        projectIndexing: ${PROJECT_FILE_INDEXING, false} # Enable citations
```
efunneko, line 115: I wonder about this default. Is there a reason that if projects are turned on that you wouldn't want to be able to do an indexed search on them? Perhaps there are reasons you don't want it, which would justify the flag, but it feels like it should be on if projects are on

src/solace_agent_mesh/agent/sac/component.py
```diff
@@ -1114,6 +1117,97 @@ def _inject_peer_tools_callback(
                 )
         return None
 
+    def _inject_project_tools_callback(
+        self, callback_context: CallbackContext, llm_request: LlmRequest
+    ) -> Optional[LlmResponse]:
+        """Dynamically add project-specific tools (e.g., index_search) when in a project chat."""
```
efunneko, line 1123: This function feels very 'boilerplatey'. Are there other functions in this file that are doing the same thing? Should it be refactored? 

(note - I haven't looked myself)

efunneko (commented):
Just a couple of comments

I do think that the approach is fine for this. I agree that we don't want to force everyone to add the indexed search tool

examples/gateways/webui_gateway_example.yaml
lgh-solace: > I wonder about this default. Is there a reason that if projects are turned on that you wouldn't want to be able to do an indexed search on them? Perhaps there are reasons you don't want it, which would justify the flag, but it feels like it should be on if projects are on

This is just during initial testing, after feature release, the flags will be removed.

src/solace_agent_mesh/agent/sac/component.py
lgh-solace: > This function feels very 'boilerplatey'. Are there other functions in this file that are doing the same thing? Should it be refactored?
> 
> (note - I haven't looked myself)

I don't mind but Claude thought this:

> The boilerplate that really stands out now is the tool injection mechanics (lines 1157-1190 of component.py). That pattern — wrap callable, create declaration, null-check config.tools, append to both dictionaries — is the same thing _inject_peer_tools_callback does at lines 1093-1106. And it's the same thing any future dynamic tool injection would need.
> 
> But here's the thing: the two callbacks wrap fundamentally different tool types. Peer tools are BaseTool subclasses (PeerAgentTool, WorkflowAgentTool) that have their own _get_declaration() method. Project tools are ADKToolWrapper → FunctionTool with a FunctionDeclaration built from registry data. A shared helper would need to handle both paths, which means it would either be parameterized to the point of being harder to read than the duplication, or it would only extract the 5-line config.tools append block — not worth an abstraction for two callers.
> 
> I'd leave it as-is. The function reads top-to-bottom, each section has a clear purpose, and the "boilerplate" is really just the ADK's tool registration API surface. If a third dynamic tool injection callback appears, that's when extracting a shared helper makes sense.

PR #1074, opened 2026-02-24
fix(DATAGO-126294): Add async write queue to PersistentSSEEventBuffer

amir-ghasemi:
This pull request introduces asynchronous database write buffering to the `PersistentSSEEventBuffer` class, improving performance and preventing blocking during slow database writes. Events are now enqueued to a bounded queue and processed by a background worker thread. The change also adds queue statistics tracking and modifies the hybrid mode flush logic to use the async queue. Unit tests have been updated to reflect the new async behavior.

**Async Write Buffering and Worker Thread**

* Added a bounded async write queue (`queue.Queue`) and a background worker thread to handle database writes asynchronously, preventing blocking and unbounded memory growth. Includes logic for graceful startup, shutdown, and draining of the queue. (`src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py`) [[1]](diffhunk://#diff-2ea67531122d1b475af9b1b6cc36d0910183db56800012d38ed10028be943ab7R45-R54) [[2]](diffhunk://#diff-2ea67531122d1b475af9b1b6cc36d0910183db56800012d38ed10028be943ab7R81-R102) [[3]](diffhunk://#diff-2ea67531122d1b475af9b1b6cc36d0910183db56800012d38ed10028be943ab7R113-R329)
* Provided queue statistics and event drop tracking via `get_async_queue_stats`, and ensured dropped events are logged and counted. (`src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py`)

**Buffering and Flushing Logic**

* Changed `_buffer_event_to_db` to enqueue events for async writing, dropping events if the queue is full and logging a warning. (`src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py`) [[1]](diffhunk://#diff-2ea67531122d1b475af9b1b6cc36d0910183db56800012d38ed10028be943ab7L309-R551) [[2]](diffhunk://#diff-2ea67531122d1b475af9b1b6cc36d0910183db56800012d38ed10028be943ab7L319-R611)
* Modified `flush_task_buffer` to enqueue all RAM-buffered events to the async queue, re-adding failed events to RAM buffer for retry if the queue is full. (`src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py`)

**Unit Test Updates**

* Updated tests to verify that events are queued for async writing rather than written directly to the DB, and to check queue stats and proper shutdown. (`tests/unit/gateway/http_sse/test_persistent_sse_buffer_comprehensive.py`) [[1]](diffhunk://#diff-1db4cedd9951732899f50cd3a41cb2b8ad27ef5c02dc9e0cb2c982f364c6220bL220-R220) [[2]](diffhunk://#diff-1db4cedd9951732899f50cd3a41cb2b8ad27ef5c02dc9e0cb2c982f364c6220bL232-R232) [[3]](diffhunk://#diff-1db4cedd9951732899f50cd3a41cb2b8ad27ef5c02dc9e0cb2c982f364c6220bL244-R246) [[4]](diffhunk://#diff-1db4cedd9951732899f50cd3a41cb2b8ad27ef5c02dc9e0cb2c982f364c6220bL274-R277) [[5]](diffhunk://#diff-1db4cedd9951732899f50cd3a41cb2b8ad27ef5c02dc9e0cb2c982f364c6220bL285-R287)

These changes collectively improve scalability and robustness for event buffering and database persistence in the gateway.


efunneko (commented):
## Deep Review: Async Write Queue for PersistentSSEEventBuffer

Overall the approach is solid — moving DB writes off the caller thread via a bounded queue + worker is a good pattern. The shutdown/drain logic, atexit handler, and bounded queue are well thought out.

However, there are a few correctness issues that need attention before merge — primarily around read-after-enqueue race conditions introduced by the fact that `flush_task_buffer` no longer writes synchronously to DB.

See inline comments for details on each issue.

src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py
```diff
@@ -89,6 +131,236 @@ def is_hybrid_mode_enabled(self) -> bool:
...
+        log.debug("%s Async write worker loop exited", self.log_identifier)
+
+    def _drain_queue_on_shutdown(self) -> None:
+        """Drain and process remaining items in the queue during shutdown."""
+        drained_count = 0
+        while True:
+            try:
+                item = self._async_write_queue.get_nowait()
+                if item is None:
+                    continue
```
efunneko, line 265: **Bug: Missing `task_done()` for sentinel `None` items**

When a `None` sentinel is encountered here, `continue` skips calling `task_done()`. In contrast, the main worker loop (line 205) correctly calls `task_done()` for sentinels.

If `queue.join()` is ever called after shutdown draining, it could hang.

```suggestion
                if item is None:
                    self._async_write_queue.task_done()
                    continue
```

src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py
```diff
@@ -72,13 +96,31 @@ def __init__(
         # RAM buffer for hybrid mode: task_id -> list of (event_type, event_data, timestamp)
         self._ram_buffer: Dict[str, List[Tuple[str, Dict[str, Any], int]]] = {}
```
efunneko: **Nit: Stale type annotation**

This says 3-tuple `Tuple[str, Dict[str, Any], int]` but `_buffer_event_hybrid` (line 540) stores 5-tuples `(event_type, event_data, timestamp, session_id, user_id)`. Should be:
```python
self._ram_buffer: Dict[str, List[Tuple[str, Dict[str, Any], int, str, str]]] = {}
```

src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py
```diff
@@ -89,6 +131,236 @@ def is_hybrid_mode_enabled(self) -> bool:
...
+                    self.log_identifier,
+                    task_id,
+                    event_type,
+                )
+                return True
+            finally:
+                db.close()
+        except Exception as e:
+            log.error(
+                "%s [Async] Failed to write event to DB for task %s: %s",
```
efunneko, line 343: **Consider: Silent data loss on DB write failure**

If a DB write fails here (e.g., transient connection error), the event is dequeued and lost — the error is logged but there's no retry.

Consider:
1. Re-enqueuing on transient failures (with a retry limit to avoid infinite loops), or
2. At minimum, tracking failed-write counts separately from dropped-event counts in `get_async_queue_stats()` so operators can distinguish "queue full" drops from "DB error" losses.

src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py
```diff
@@ -316,119 +592,112 @@ def _buffer_event_to_db(
...
+            Number of events enqueued for flushing
         """
         if not self.is_hybrid_mode_enabled():
             return 0
         
-        # Extract events from RAM buffer under lock
-        events_to_flush = []
+        # Hold lock for entire operation to prevent race conditions where
+        # new events could be added between pop and re-add, causing out-of-order events.
         with self._lock:
```
efunneko, line 830: **Minor: Lock held during queue operations with nested lock acquisition**

The lock is held for the entire flush including `put_nowait` calls and nested `_dropped_events_lock` acquisitions (line 676). While safe today since `_dropped_events_lock` is never held when acquiring `_lock`, this nested locking pattern is fragile. Consider documenting the lock ordering contract (always acquire `_lock` before `_dropped_events_lock`, never the reverse).

efunneko (commented):
### Additional issues on unchanged lines (can't comment inline):

**Critical: `get_buffered_events` (line 766-768) — Read-after-enqueue race condition**

In hybrid mode, `get_buffered_events` calls `flush_task_buffer` which now enqueues events to the async queue instead of writing them directly to DB. Then it immediately queries the DB — but the async worker may not have processed those events yet, so the DB query will return incomplete results.

Before this PR, `flush_task_buffer` wrote synchronously to DB, so the subsequent read was guaranteed to see the flushed events. That guarantee is now broken. Clients reconnecting to retrieve buffered events may get incomplete data.

**Suggested fix**: Either:
1. Call `self._async_write_queue.join()` after flushing and before querying DB (ensures all pending writes complete), or
2. Have `flush_task_buffer` take an optional `synchronous=True` parameter that writes directly to DB for read-path callers, or
3. Return the in-flight queue events combined with DB results

---

**Critical: `has_unconsumed_events` (line 818-821) — Same race condition**

In hybrid mode, this checks the RAM buffer then falls through to the DB check. But events that were recently flushed from RAM to the async queue (but not yet written to DB by the worker) would be invisible to *both* checks — RAM is empty (they were popped by flush), and DB doesn't have them yet. Returns a false negative.

---

**Low: `SSEManager.close_all()` doesn't call `_persistent_buffer.shutdown()`**

The `close_all()` method flushes hybrid buffers but never calls `shutdown()` on the persistent buffer, so the async worker thread keeps running. The only shutdown path is via the `atexit` handler. Consider calling `self._persistent_buffer.shutdown()` from `close_all()` for explicit lifecycle management.

src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py
amir-ghasemi: fixed

src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py
amir-ghasemi: The _drain_queue_on_shutdown method now properly calls task_done() for sentinel None values to prevent queue.join() hangs.

src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py
amir-ghasemi: Updated get_async_queue_stats(): Now includes failed_writes count in the returned stats.

src/solace_agent_mesh/gateway/http_sse/persistent_sse_event_buffer.py
amir-ghasemi: Documented the required lock acquisition order to prevent deadlocks.

amir-ghasemi:
superseded.

PR #1307, opened 2026-03-31
fix(DATAGO-123668): Implement verification step in deep research process

amir-ghasemi:
This pull request introduces support for displaying and handling deep research plan verification in the chat UI. It adds a new `ResearchPlanVerification` component, updates message processing logic to recognize and render research plan data, and ensures proper handling of these messages throughout the chat provider and utility functions. The changes also include comprehensive tests for the new logic.

**Deep Research Plan Verification Feature:**

* Added a new `ResearchPlanVerification` component that displays research plans for user review, allows editing of steps, and provides Start/Cancel actions before deep research begins. The component is integrated into the chat message rendering flow. [[1]](diffhunk://#diff-c25650a7f000608c521a44e1a5be92d34f4fc16664f4f02d81538564a0fab997R1-R214) [[2]](diffhunk://#diff-f9b9a25a941b9debe58c474838e457efe96c638ee37362656c18e86024aca620R13) [[3]](diffhunk://#diff-3be55076cb2c38a259217b6f8e878c194770f0e688b83e9e6199cd51cdb3daf9R3) [[4]](diffhunk://#diff-f9b9a25a941b9debe58c474838e457efe96c638ee37362656c18e86024aca620R649-R655)

**Chat Message Processing and State Management:**

* Updated `ChatProvider` to recognize `deep_research_plan` data parts, log them, and manage their insertion or replacement in the chat message list, similar to how progress updates are handled. [[1]](diffhunk://#diff-d33c8017d7b1273101dabad6b9fb45a9a00f0312bada8b8ec387b325a21bc2f7R1267-R1276) [[2]](diffhunk://#diff-d33c8017d7b1273101dabad6b9fb45a9a00f0312bada8b8ec387b325a21bc2f7L1431-R1470)

**Message Utilities and Filtering:**

* Enhanced the message processing utilities to treat `deep_research_plan` data parts as renderable and visible content, ensuring they are not filtered out and are properly displayed in the chat. [[1]](diffhunk://#diff-1643df935f4ab2bbe59e90f59bec68108d0551ec38514113f372d5215f90f96fL6-R15) [[2]](diffhunk://#diff-1643df935f4ab2bbe59e90f59bec68108d0551ec38514113f372d5215f90f96fL28-R35)

**Testing:**

* Added and updated tests to verify that `deep_research_plan` data parts are kept by the filtering logic and recognized as containing visible content. [[1]](diffhunk://#diff-49c790bd8bb19c401478f2e6bf10f379e3237ca4ccd26b89d72de9e3315f817dR37-R41) [[2]](diffhunk://#diff-49c790bd8bb19c401478f2e6bf10f379e3237ca4ccd26b89d72de9e3315f817dR89-R92)

image

When editing:

image


lgh-solace (commented):
Consider these potential issues. Thanks!

## 🔴

**Infinite blocking poll with no timeout**
`_wait_for_plan_response` in `deep_research_tools.py` loops forever with `await asyncio.sleep(0.5)`. If the SSE connection drops or the user closes the tab before responding, the agent task hangs permanently — no timeout, no cancellation path. The auto-approve timeout was explicitly disabled. At minimum a hard timeout (e.g. 5–10 minutes) should be reinstated so the tool can fail gracefully.

**`_is_webui_gateway` always returns `True`**
The function is supposed to detect whether the current gateway supports interactive verification, but always returns `True` with a TODO comment. Non-interactive gateways (Slack, Teams, MCP) will trigger the verification step and then block forever. This needs real detection (e.g. reading a flag from `a2a_context`) before merging, or the verification step must be gated behind an explicit opt-in config flag.

---

## 🟠

**`ResearchPlanVerification` is re-submittable after page reload**
After the user responds, the component returns `null` but the `deep_research_plan` data part remains in the message. On reload the component remounts in its initial interactive state — clicking Start again writes a stale `plan_id` to the cache with no backend listener. The responded state needs to be persisted (e.g. in the message itself via ChatProvider) rather than held only in local component state.

**API call bypasses `lib/api` pattern**
`handleResponse` in `ResearchPlanVerification.tsx` calls `api.webui.post(...)` directly with manual `useState` for loading — violates the `useQuery`/`useMutation` pattern. Should be extracted to `lib/api/research/hooks.ts` with a `useSubmitPlanResponse` mutation.

**No `displayError` for user-facing errors**
The `catch` in `handleResponse` only calls `console.error`. If the POST fails the user sees nothing. Use `displayError` from `useChatContext()`.

**Native `<input>` in edit mode**
The step edit fields use a native `<input type="text">` — should use the standard `Input` component from `@SolaceLabs/solace-agent-mesh-ui`.

---

## 🟡

**`console.log` left in production code**
`ChatProvider.tsx` in the `deep_research_plan` case has a `console.log` that should be removed.

**`onResponded` prop is dead code**
`ResearchPlanVerification` defines an `onResponded` callback prop but it is never passed at the call site in `ChatMessage.tsx`. Either wire it up or remove the prop.

**`skip_verification` config key is undocumented**
Read via `config.get("skip_verification", False)` but never mentioned in `shared_config.yaml`, `agent_template.yaml`, or any docstring.

**Extra LLM call to reformat queries into plan steps**
`_generate_research_plan` makes a full LLM round-trip just to convert search queries into human-readable steps. This adds latency and cost to every invocation. A lightweight text transform (strip search operators, title-case) would likely be sufficient.

**`auto_approve_seconds` is exposed but inert**
The field is sent in `DeepResearchPlanData`, shown in the component props, and has a commented-out countdown UI — but both the backend timeout and frontend timer are disabled. Remove the field or implement it; the current state is misleading.

---

## ⚪ Low / Style

**`React.FC` pattern** — `ResearchPlanVerification` should use direct prop typing per frontend patterns.

**Missing explicit return type** on `ResearchPlanVerification`.

**No Storybook story or test** for `ResearchPlanVerification`.

**`DashedCircle` SVG declared before import statements** in `ResearchPlanVerification.tsx` — minor ordering issue.

**Cache key not namespaced by user/session** — `deep_research_plan_{plan_id}` uses a UUID so collision is unlikely, but a compromised or buggy client could write a response for another user's plan if the UUID is known.

lgh-solace (commented):
Considerations?

1. Wrong cache key in test — [tests/unit/gateway/http_sse/routers/test_research.py:85](vscode-webview://1nneiq5slm60qg8japnf8a3mbiiqg1s642jm1ndpvj6o2g8k39mv/tests/unit/gateway/http_sse/routers/test_research.py#L85) asserts "deep_research_plan_plan-abc" but the router produces "deep_research_plan:user-1:plan-abc". Test fails on first run.

2. auto_approve_seconds doesn't exist on the model — [tests/unit/common/test_deep_research_plan_data.py:27,39,43](vscode-webview://1nneiq5slm60qg8japnf8a3mbiiqg1s642jm1ndpvj6o2g8k39mv/tests/unit/common/test_deep_research_plan_data.py#L27) constructs and asserts a field that isn't on DeepResearchPlanData. The model docstring promises an "auto-approve countdown timer" though, so someone needs to decide: drop the assertion or add the field.

3. Cache TTL shorter than wait window — [routers/research.py:94](vscode-webview://1nneiq5slm60qg8japnf8a3mbiiqg1s642jm1ndpvj6o2g8k39mv/src/solace_agent_mesh/gateway/http_sse/routers/research.py#L94) stores with expiry=120, but [deep_research_tools.py:985](vscode-webview://1nneiq5slm60qg8japnf8a3mbiiqg1s642jm1ndpvj6o2g8k39mv/src/solace_agent_mesh/agent/tools/deep_research_tools.py#L985) waits up to 600. Users who read the plan for >2 min lose their response. Share the constant between the two.

Other stuff:

- DeepResearchPlanData missing from SignalData union ([data_parts.py:554](vscode-webview://1nneiq5slm60qg8japnf8a3mbiiqg1s642jm1ndpvj6o2g8k39mv/src/solace_agent_mesh/common/data_parts.py#L554))
- Unused FastAPIRequest param in the new router
- No feature-flag gating for the new UX
- FE: index-as-key during edit, no "cancel edit" affordance, missing aria-labels on icon buttons, research_question sent but never rendered
- Security: new POST endpoint not CSRF-checked, no null guard on user_id

lgh-solace (approved):
Please just review the file comments from the co-pilot review. Thanks!

client/webui/frontend/src/lib/components/research/ResearchPlanVerification.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+import { useSubmitPlanResponse } from "@/lib/api";
+import { Button } from "@/lib/components/ui/button";
+import { Input } from "@/lib/components/ui/input";
+import { useChatContext } from "@/lib/hooks";
+import type { DataPart } from "@/lib/types";
+
+import { getRespondedPlansSnapshot, markPlanResponded, subscribeRespondedPlans } from "./respondedPlansStore";
+
+const DashedCircle = ({ className }: { className?: string }) => (
```
lgh-solace: There is a lucide icon for this that may be cleaner - https://lucide.dev/icons/circle-dashed

PR #1362, opened 2026-04-09
fix(DATAGO-130050): GH1258 Workflow nodes silently fail with "mandatory result embed" error if artifact_management tool group is missing 

solace-pjones:

Artifact management tools are a baseline capability that most agents rely on directly or indirectly. Previously, agents whose YAML didn't declare the `artifact_management` group would silently start without these tools and fail later at runtime in confusing ways. This change makes the group a runtime guarantee so the class of misconfiguration is eliminated.

- `SamAgentApp.__init__` now invokes `_ensure_artifact_management_tools()` on the raw config dict before Pydantic validation. The function prepends `{tool_type: builtin-group, group_name: artifact_management}` to the `tools` list when the group isn't already present. Straight-line logic, no exception swallowing.
- New Pydantic field `auto_inject_artifact_tools: bool = True` on `SamAgentAppConfig` provides an explicit opt-out for the rare case where a custom Python tool collides with one of the 8 builtin artifact tool names. The field's description enumerates the tool names so authors hitting a collision know what to opt out of.
- Extracted two shared constants to prevent literal drift:
  - `BUILTIN_GROUP_TOOL_TYPE` in `tool_config_types.py`, derived via `get_args(...)` from the Pydantic Literal itself — a rename of the Literal propagates automatically.
  - `ARTIFACT_MANAGEMENT_CATEGORY` in `builtin_artifact_tools.py`, used by all 7 tool-decorator sites and the injector.
- Added `auto_inject_artifact_tools` section and opt-out documentation to `docs/docs/documentation/components/builtin-tools/artifact-management.md`.

- **No silent failure path.** The injector has no `try/except`; the body only performs infallible dict/list operations on a freshly-parsed YAML dict, so any genuine error surfaces loudly with a traceback rather than leaving the agent in the exact silent state this PR is meant to prevent.
- **Explicit opt-out over collision detection.** Python tools' names come from the function, not the YAML, so collision detection at injection time would require pre-loading every tool module. The explicit flag is simpler and gives authors full control.
- **Default-on, not feature-flagged.** The behavior is the new baseline; opt-out exists but shouldn't be needed in practice.


amir-ghasemi (commented):
Overall the fix addresses a real usability footgun, but the broad `except` contradicts the PR's stated goal and the "no opt-out" design carries real backwards-compat risk for agents with custom tools that collide with the injected names.

src/solace_agent_mesh/agent/sac/app.py
```diff
@@ -547,6 +547,36 @@ def _validate_model_requirement(self) -> "SamAgentAppConfig":
...
+        }
+        if "artifact_management" not in configured_groups:
+            app_config_dict["tools"] = [
+                {"tool_type": "builtin-group", "group_name": "artifact_management"}
+            ] + list(tools)
+            log.info(
+                "Agent '%s': auto-injected 'artifact_management' tool group (not found in YAML config).",
+                agent_name,
+            )
+    except Exception:
```
amir-ghasemi: **🟡 Silent failure mode recreates the bug this PR claims to fix**

The body of the `try` is essentially infallible (dict `.get`, set comprehension, list concat on values already coming from a parsed YAML dict). If anything *does* go wrong, the agent boots without artifact tools and later hits the exact original *"mandatory result embed"* failure — now with no stack trace back to this injection site.

The PR's motivation is to eliminate a silent failure; this bare `except Exception` re-introduces the same failure class one layer up.

**Suggested**: drop the try/except entirely (let Pydantic surface a much clearer error than the current runtime symptom), or narrow it to the specific exception you're actually worried about with a comment explaining the concrete scenario.

src/solace_agent_mesh/agent/sac/app.py
```diff
@@ -547,6 +547,36 @@ def _validate_model_requirement(self) -> "SamAgentAppConfig":
...
+    try:
+        tools = app_config_dict.get("tools") or []
+        configured_groups = {
+            t.get("group_name")
+            for t in tools
+            if isinstance(t, dict) and t.get("tool_type") == "builtin-group"
+        }
+        if "artifact_management" not in configured_groups:
+            app_config_dict["tools"] = [
+                {"tool_type": "builtin-group", "group_name": "artifact_management"}
```
amir-ghasemi: **🟡 "No opt-out" will break existing agents on tool-name collision**

In `agent/adk/setup.py` the tool loader raises `ValueError` on duplicate tool names. After this change, any existing agent YAML that defines a Python tool named `load_artifact`, `save_artifact`, `append_to_artifact`, `list_artifacts`, `delete_artifact`, `extract_content_from_artifact`, `artifact_search_and_replace_regex`, or `apply_embed_and_create_artifact` will **fail to start** where it previously worked — with no migration path.

Options:
1. Skip injection when a name collision would occur (and log it).
2. Add an explicit escape hatch (`auto_inject_artifact_tools: false`).
3. Document as a breaking change in release notes and audit internal agents first.

The PR body says this is by design, so at minimum please confirm internal agents have been audited for the 8 tool names above.

amir-ghasemi: **🟡 Hardcoded literals drift risk**

`"builtin-group"` and `"artifact_management"` are hardcoded here, duplicating:
- the `Literal["builtin-group"]` type in `agent/tools/tool_config_types.py`
- the `@tool_registry(category="artifact_management")` decorators in `agent/tools/builtin_artifact_tools.py`

If either is ever renamed, this injector silently stops matching (→ double-injection, hard `ValueError` at tool-load time) with no compile-time signal. Recommend extracting shared constants (or at minimum importing the existing `Literal` source so a rename fails loudly).

src/solace_agent_mesh/agent/sac/app.py
```diff
@@ -547,6 +547,36 @@ def _validate_model_requirement(self) -> "SamAgentAppConfig":
         return self
 
 
+def _ensure_artifact_management_tools(app_config_dict: Dict[str, Any], agent_name: str) -> None:
```
amir-ghasemi, line 562: **🟡 Missing unit test — PR checkbox says "Unit tests" but none were added**

`_ensure_artifact_management_tools` is a pure function over a dict — trivially unit-testable. Grep across `tests/` finds zero references. The natural home is `tests/unit/agent/sac/test_agent_sac_config.py`.

The integration fixture only covers one case (*different* builtin-group already configured). Uncovered:
- `artifact_management` already present → should be a no-op
- `tools` key missing entirely
- `tools: null`
- `tools: []`
- non-dict entries in the tools list (exception path)

A parametrized `pytest` unit test would cover all five in ~20 lines and protect against the literal-drift issue flagged separately.

tests/integration/scenarios_declarative/test_data/workflows/test_workflow_missing_artifact_tools.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - static_response:
+      choices:
+        - message:
+            role: "assistant"
+            content: |
+              «««save_artifact: filename="summary.json" mime_type="application/json" description="Summary output"
+              {"summary": "This is a summary of the input text."}
+              »»»
+    expected_request:
+      tools_present: ["load_artifact"]
```
amir-ghasemi: **🟡 Assertion is too narrow for the behavior under test**

The injected `artifact_management` group contributes 8 tools. Asserting only `load_artifact` means a partial-injection regression (e.g., group membership shrinks upstream, or the registry category is renamed) would still pass here.

Compare with `tests/integration/scenarios_declarative/test_data/dynamic_tools/test_empty_provider.yaml` which uses `assert_tools_exact` with the full set — mirror that:

```yaml
    expected_request:
      assert_tools_exact:
        - "_continue_generation"
        - "_notify_artifact_save"
        - "append_to_artifact"
        - "apply_embed_and_create_artifact"
        - "artifact_search_and_replace_regex"
        - "delete_artifact"
        - "extract_content_from_artifact"
        - "list_artifacts"
        - "load_artifact"
```

or at minimum widen `tools_present` to include `save_artifact`, `list_artifacts`, `delete_artifact`.

solace-pjones: fixed

src/solace_agent_mesh/agent/sac/app.py
solace-pjones: fixed

src/solace_agent_mesh/agent/sac/app.py
solace-pjones: fixed

solace-pjones: fixed

src/solace_agent_mesh/agent/sac/app.py
solace-pjones: fixed

amir-ghasemi (approved):
LGTM

PR #1368, opened 2026-04-10
chore(DATAGO-131398): Create general and planning model by default is seeding does not occur

solace-pjones:
image

Seed placeholder default models so UI always shows general & planning

- Seed "general" and "planning" model configs as placeholders at startup, even when no config/env vars are provided. The UI always shows these defaults with a "Not configured" state and warning icon until the user configures them.
- Placeholder values (`"undefined"`) are stored in the DB to satisfy column constraints, then stripped to `null` in API responses so the frontend receives clean data.
- Default models (general, planning) cannot be deleted — the delete endpoint returns a 422 error.
- Frontend handles nullable `provider` and `modelName` throughout: detail page, edit page, list view, and provider icon component all gracefully show "Not configured" for unconfigured defaults.

1. Sentinel value in DB vs. nullable columns
Chose to store "undefined" as a sentinel rather than making the DB columns nullable. This satisfies existing LENGTH >= 1 constraints without a migration, and gets stripped to null at the API boundary. Trade-off: an extra stripping step in _to_response() and _to_raw_litellm_config(), but avoids schema changes.

2. Always-seed vs. on-demand creation
_ensure_default_aliases() runs on every startup after config/env seeding, not just on first boot. This means if someone manually deletes rows from the DB (outside the API), they come back on restart. The API delete endpoint blocks it, but this is a safety net.

- [X] Manual testing: [describe scenarios]
- [X] Unit tests: [new/modified tests]


cyrus2281 (approved):
LGTM

PR #1375, opened 2026-04-10
fix(DATAGO-130051): update tool name and description handling for DynamicTool subclasses, closes #1255

cyrus2281:

This PR fixes how DynamicTool subclasses are handled when generating tool manifests and scope maps. Previously, these tools could show up with placeholder values ("dynamic_tool_placeholder") in the agent card tool manifest instead of their real names and descriptions.

High-level approach involves checking for tool_name/tool_description properties on DynamicTool subclasses before falling back to name/description attributes:

- fix: update tool name and description handling for DynamicTool subclasses
- Updated tool name retrieval in both the scopes map builder and manifest builder
- Added unit tests to verify both DynamicTool and regular tool handling

### Key Design Decisions

The implementation uses property accessors (tool_name, tool_description) to get the real values from DynamicTools rather than relying on the inherited name/description attributes which often contain placeholders until the tool is actually invoked.

- [x] Unit tests: Added two comprehensive test files covering both the scopes map builder and the manifest builder
- [x] Tests verify that DynamicTools are properly identified by their real names, not placeholders
- [x] Tests validate mixed scenarios with both DynamicTool and regular tool types
- [ ] Known limitations: This change does not impact runtime tool declaration - only affects agent card manifest and scope maps

Pay attention to the property accessor pattern used to retrieve real names from DynamicTool subclasses and verify this is compatible with all DynamicTool implementations.

Closes #1255


gregmeldrum (approved):
LG

PR #1377, opened 2026-04-10
feat(DATAGO-131548): Add per-request model override via A2A task metadata

bwiebe-solace:
Nice work on this one — solid design decisions (alias-only for MVP, ContextVar isolation, feature flag gating). Code is clean and test coverage is thorough. A few items worth discussing below.

### Highlights
- **ContextVar approach** for per-request isolation is the right pattern — clean, task-scoped, avoids plumbing override config through every call layer.
- **Feature flag gating** with `offline_evals` makes this safe to merge without impacting production traffic.
- **Request deduplication** in `resolve()` (concurrent callers for the same alias share one broker round-trip) is a nice optimization.
- **Alias-only for MVP** is the correct security posture — the exfiltration risk of accepting raw configs is well articulated in the PR description.

### Summary
| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🟡 Medium | `dynamic_model_provider.py` | Wildcard subscription broadness |
| 2 | 🟡 Medium | `dynamic_model_provider.py` | Topic parsing is fragile |
| 3 | 🟡 Medium | `event_handlers.py` | Silent error drop when no reply topic |
| 4 | 🔵 Low | `dynamic_model_provider.py` | Accessing private `_model_id` from receiver |
| 5 | 🔵 Low | `dynamic_model_provider.py` | `complete_pending_resolve` accesses private `_async_loop` |
| 6 | 🔵 Low | `lite_llm.py` | Shallow `.copy()` on override config |
| 7 | 🔵 Low | `event_handlers.py` | Feature flag evaluated per-request — confirm caching |
| 8 | ⚪ Nit | `test_dynamic_model_provider_resolve.py` | Deprecated `get_event_loop()` usage |

### DIFF
--- a/src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py
+++ b/src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py
@@ -91,10 +91,15 @@ class ModelConfigReceiverComponent(ComponentBase):
 
             model_config = payload.get("model_config")
 
-            # Extract model_id from topic: .../response/{model_id}/{component_id}
-            # or .../model/{model_id} for config updates
+            # Extract model_id from topic based on shape:
+            #   Bootstrap response: .../response/{model_id}/{component_id} → parts[-2]
+            #   Config update:      .../configuration/model/{model_id}    → parts[-1]
             parts = topic.split("/")
-            topic_model_id = parts[-2] if len(parts) >= 2 else None
+            is_bootstrap_response = len(parts) >= 3 and parts[-3] == "response"
+            if is_bootstrap_response:
+                topic_model_id = parts[-2]
+            else:
+                topic_model_id = parts[-1] if parts else None
 
             if topic_model_id == self.model_provider._model_id:
                 # Response for this provider's own model — existing behavior
--- a/src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py
+++ b/src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py
@@ -2,12 +2,14 @@
 Dynamic Model Provider for enterprise model configuration.
 """
 
-from typing import Any, Dict, Union
 import asyncio
+import logging
+import threading
+from typing import Any, Dict, List, Optional, Union
+
 from solace_agent_mesh.agent.adk.models.lite_llm import LiteLlm
 from solace_ai_connector.components.component_base import ComponentBase
 from solace_ai_connector.common.message import Message as SolaceMessage
-import logging
 
 from .dynamic_model_provider_topics import (
     get_bootstrap_request_topic,
@@ -19,6 +21,7 @@ log = logging.getLogger(__name__)
 
 _MAX_BOOTSTRAP_RETRIES = 3
 _BOOTSTRAP_RETRY_INTERVAL_SECONDS = 5
+_RESOLVE_TIMEOUT_SECONDS = 10.0
 
 
 # SAC Component Info for ModelConfigReceiverComponent
@@ -75,13 +78,6 @@ class ModelConfigReceiverComponent(ComponentBase):
         log.info("%s ModelConfigReceiverComponent initialized.", self.log_identifier)
 
     def invoke(self, message: SolaceMessage, data: Dict[str, Any]) -> None:
-        """
-        Processes the incoming model configuration message.
-
-        Args:
-            message: The SolaceMessage object from BrokerInput.
-            data: The data extracted by BrokerInput (payload, topic, user_properties).
-        """
         log_id_prefix = f"{self.log_identifier}[Invoke]"
         try:
             topic = data.get("topic", "")
@@ -93,26 +89,36 @@ class ModelConfigReceiverComponent(ComponentBase):
                 topic,
             )
 
-            # Check if model_config exists in payload
             model_config = payload.get("model_config")
 
-            if model_config:
-                log.info(
-                    "%s Model config found, updating LiteLlm: %s",
-                    log_id_prefix,
-                    model_config.get('model', 'N/A'),
-                )
-                self.model_provider.mark_initialized()
-                self.model_provider.update_litellm_model(model_config)
+            # Extract model_id from topic: .../response/{model_id}/{component_id}
+            # or .../model/{model_id} for config updates
+            parts = topic.split("/")
+            topic_model_id = parts[-2] if len(parts) >= 2 else None
+
+            if topic_model_id == self.model_provider._model_id:
+                # Response for this provider's own model — existing behavior
+                if model_config:
+                    log.info(
+                        "%s Model config found, updating LiteLlm: %s",
+                        log_id_prefix,
+                        model_config.get('model', 'N/A'),
+                    )
+                    self.model_provider.mark_initialized()
+                    self.model_provider.update_litellm_model(model_config)
+                else:
+                    log.info(
+                        "%s No model config or empty config, removing LiteLlm model",
+                        log_id_prefix,
+                    )
+                    self.model_provider.remove_litellm_model()
             else:
-                log.info(
-                    "%s No model config or empty config, removing LiteLlm model",
-                    log_id_prefix,
+                # Response for a different model_id — one-shot alias resolution
+                self.model_provider.complete_pending_resolve(
+                    topic_model_id, model_config
                 )
-                self.model_provider.remove_litellm_model()
 
             message.call_acknowledgements()
-            log.debug("%s Message acknowledged to BrokerInput.", log_id_prefix)
 
         except Exception as e:
             log.exception(
@@ -139,6 +145,11 @@ class DynamicModelProvider:
 
         # Initial model configuration
         self._initialized = False
+
+        # One-shot alias resolution state (thread-safe)
+        self._pending_resolves: Dict[str, List[asyncio.Future]] = {}
+        self._resolve_lock = threading.Lock()
+
         asyncio.create_task(self.initialize())
 
     async def initialize(self):
@@ -232,9 +243,10 @@ class DynamicModelProvider:
                 )
                 raise ValueError("Main app broker configuration is missing.")
 
-            # Subscribe to the model config update topic
             config_update_topic = get_model_config_update_topic(self._component.namespace, self._model_id)
-            config_bootstrap_topic = get_bootstrap_response_topic(self._component.namespace, self._model_id, self._component.get_component_id())
+            # Wildcard subscription catches both own-model bootstrap responses
+            # and one-shot alias resolution responses for any model_id
+            config_bootstrap_topic = get_bootstrap_response_topic(self._component.namespace, "*", self._component.get_component_id())
             component_id = self._component.get_component_id()
 
             broker_input_cfg = {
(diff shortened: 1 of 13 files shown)


src/solace_agent_mesh/agent/protocol/event_handlers.py
```diff
@@ -1121,6 +1123,54 @@ async def _handle_send_message_request(
...
+            if component._dynamic_model_provider:
+                resolved = await component._dynamic_model_provider.resolve(model_id)
+                if resolved:
+                    task_metadata["model_override"] = resolved
+                    log.info(
+                        "%s Resolved model override alias '%s' to model=%s",
+                        component.log_identifier,
+                        model_id,
+                        resolved.get("model"),
+                    )
```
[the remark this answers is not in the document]
bwiebe-solace: Acknowledged: the resolved config dict does contain credentials and flows through original_message_metadata. However, this is consistent with how DynamicModelProvider already operates: it stores the full raw config (including api_key) on the LiteLlm instance, which is accessible throughout the component lifecycle with the same logging exposure.

src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py
```diff
@@ -93,26 +97,40 @@ def invoke(self, message: SolaceMessage, data: Dict[str, Any]) -> None:
...
-                    "%s Model config found, updating LiteLlm: %s",
-                    log_id_prefix,
-                    model_config.get('model', 'N/A'),
-                )
-                self.model_provider.mark_initialized()
-                self.model_provider.update_litellm_model(model_config)
+            # Extract model_id from topic based on shape:
+            #   Bootstrap response: .../response/{model_id}/{component_id} → parts[-2]
+            #   Config update:      .../configuration/model/{model_id}    → parts[-1]
+            parts = topic.split("/")
```
josephabonasara: ### 🟡 Wildcard subscription is overly broad

Changing from a specific `model_id` subscription to `"*"` means this component now receives bootstrap responses for **every model config** in the namespace — not just the ones this agent cares about.

In a deployment with many model configs, this adds unnecessary broker traffic to every agent instance. Consider:
- A **separate, scoped subscription** added on-demand when `resolve()` is called (subscribe to the specific alias, unsubscribe after response)
- Or at minimum, document the expected traffic impact for deployments with many model configs

src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py
```diff
@@ -75,12 +78,13 @@ def __init__(self, **kwargs: Any):
         log.info("%s ModelConfigReceiverComponent initialized.", self.log_identifier)
 
     def invoke(self, message: SolaceMessage, data: Dict[str, Any]) -> None:
-        """
-        Processes the incoming model configuration message.
```
josephabonasara, line 79: ### 🟡 Topic parsing is fragile

Positional extraction of `model_id` from the topic string (`parts[-2]` vs `parts[-1]`) is brittle — it will silently extract wrong values if the topic format ever changes.

Consider using the existing topic-building functions in `dynamic_model_provider_topics.py` to create a shared parser, or at least extracting named constants for the expected positions. This keeps extraction in sync with construction.

```python
# e.g., a helper that mirrors the builders:
def extract_model_id_from_response_topic(topic: str) -> Optional[str]:
    ...
```

src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py
```diff
@@ -75,12 +78,13 @@ def __init__(self, **kwargs: Any):
...
+        """Process an incoming model configuration message.
 
-        Args:
-            message: The SolaceMessage object from BrokerInput.
-            data: The data extracted by BrokerInput (payload, topic, user_properties).
+        Routes by model_id extracted from the topic:
+        - If model_id matches this provider's configured model, updates
+          the LiteLlm instance (bootstrap response or config change).
```
josephabonasara, line 86: ### 🔵 Accessing private `_model_id`

`self.model_provider._model_id` reaches into a private attribute. Consider exposing a public property (e.g., `model_id`) on `DynamicModelProvider` to keep the receiver decoupled from internal state.

src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py
```diff
@@ -232,9 +255,10 @@ def _ensure_config_listener_flow_is_running(self) -> None:
                 )
                 raise ValueError("Main app broker configuration is missing.")
```
josephabonasara, line 258: ### 🔵 Accessing private `_async_loop`

```python
loop = self._component._async_loop
```

Same pattern — reaching into a private attribute of the component. If `_async_loop` isn't set yet (e.g., cleanup races setup), this raises `AttributeError`. A defensive check or a public accessor would be safer.

src/solace_agent_mesh/agent/adk/models/lite_llm.py
```diff
@@ -132,6 +132,33 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
 
+# Per-request model override via ContextVar.
+# Set by the model_override callback (before_model chain) and consumed by
+# generate_content_async(). Each asyncio.Task gets its own copy, so
+# concurrent requests with different overrides are isolated.
+_model_override_var: contextvars.ContextVar[Optional[Dict[str, Any]]] = contextvars.ContextVar(
+    "model_override_config", default=None
+)
+
+
```
josephabonasara, line 144: ### 🔵 Shallow `.copy()` on override config

```python
override_copy = override_config.copy()
```

If `override_config` ever contains nested dicts (e.g., provider-specific options, headers), this shallow copy shares mutable references with the original — mutations to `completion_args` could leak back.

Not a problem today with flat configs, but `copy.deepcopy` (or documenting the flat-config contract) would prevent subtle bugs as the schema evolves.

src/solace_agent_mesh/agent/protocol/event_handlers.py
```diff
@@ -1046,6 +1048,75 @@ def _send_error_response_and_nack(
```
josephabonasara, line 1045: ### 🔵 Feature flag evaluated per-request

This calls `openfeature_api.get_client().get_boolean_value(...)` on every incoming A2A message that carries `model_override`. Worth confirming that the OpenFeature client caches flag evaluations locally rather than making a remote call each time — otherwise this becomes a latency/availability concern on the hot path.

src/solace_agent_mesh/agent/protocol/event_handlers.py
```diff
@@ -1046,6 +1048,75 @@ def _send_error_response_and_nack(
...
+        )
+        task_metadata.pop("model_override", None)
+        return None
+
+    if not (
+        isinstance(model_override, dict)
+        and isinstance(model_override.get("model_id"), str)
+        and model_override["model_id"]
+    ):
+        log.warning(
```
josephabonasara, line 1085: ### 🟡 Silent error drop when no reply topic exists

```python
if target_topic:
    component.publish_a2a_message(...)
message.call_negative_acknowledgements()
return None
```

If `target_topic` is `None` (no `reply_topic_from_peer` and no `client_id`), the error response is silently dropped — the caller never learns why their request failed. The message is NACKed but the caller receives no error payload.

Consider adding a `log.warning` for this path, or asserting that eval requests always carry a reply topic.

tests/unit/agent/adk/models/test_dynamic_model_provider_resolve.py
```diff
@@ -0,0 +1,188 @@
+"""Tests for DynamicModelProvider.resolve() — one-shot model alias resolution."""
+
+import asyncio
+import pytest
+from unittest.mock import MagicMock, patch
+
+
+def _make_provider():
+    """Create a DynamicModelProvider with mocked dependencies."""
```
josephabonasara, line 9: ### ⚪ Nit: Deprecated `get_event_loop()`

```python
component._async_loop = asyncio.get_event_loop()
```

`asyncio.get_event_loop()` is deprecated in Python 3.10+ when no loop is running. In `@pytest.mark.asyncio` tests, `asyncio.get_running_loop()` is the correct call.

josephabonasara (commented):
Claude Code Review:
Nice work on this one — solid design decisions (alias-only for MVP, ContextVar isolation, feature flag gating). Code is clean and test coverage is thorough. A few items worth discussing below.

### Highlights
- **ContextVar approach** for per-request isolation is the right pattern — clean, task-scoped, avoids plumbing override config through every call layer.
- **Feature flag gating** with `offline_evals` makes this safe to merge without impacting production traffic.
- **Request deduplication** in `resolve()` (concurrent callers for the same alias share one broker round-trip) is a nice optimization.
- **Alias-only for MVP** is the correct security posture — the exfiltration risk of accepting raw configs is well articulated in the PR description.

### Summary
| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🟡 Medium | `dynamic_model_provider.py` | Wildcard subscription broadness |
| 2 | 🟡 Medium | `dynamic_model_provider.py` | Topic parsing is fragile |
| 3 | 🟡 Medium | `event_handlers.py` | Silent error drop when no reply topic |
| 4 | 🔵 Low | `dynamic_model_provider.py` | Accessing private `_model_id` from receiver |
| 5 | 🔵 Low | `dynamic_model_provider.py` | `complete_pending_resolve` accesses private `_async_loop` |
| 6 | 🔵 Low | `lite_llm.py` | Shallow `.copy()` on override config |
| 7 | 🔵 Low | `event_handlers.py` | Feature flag evaluated per-request — confirm caching |
| 8 | ⚪ Nit | `test_dynamic_model_provider_resolve.py` | Deprecated `get_event_loop()` usage |

src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py
bwiebe-solace: The wildcard is intentional for the one-shot alias resolution. However, the scope is narrower than it appears: the subscription pattern is `response/*/{component_id}`, so it only matches bootstrap responses addressed to this specific component. Other agents' bootstrap responses are directed to their own component IDs and won't match.

Config update notifications (`configuration/model/{model_id}`) use a separate, model-specific subscription and are unaffected by the wildcard.

An on-demand subscribe/unsubscribe per `resolve()` call was considered, but it would add latency (broker round-trip for subscription setup before we can send the request) and lifecycle complexity. Since the wildcard only captures responses meant for this agent, the additional traffic is limited to resolve requests this agent actually initiates.

src/solace_agent_mesh/agent/protocol/event_handlers.py
bwiebe-solace: The OpenFeature client evaluates flags locally from in-memory state (backed by `InMemoryProvider` loaded from `features.yaml` at startup), so there are no remote calls or latency concerns.

Additionally, this code path is only reached when model_override is present in task metadata, which only happens for eval requests. Normal agent traffic never enters this branch so for users using the system normally there won't be any significant change in performance.

josephabonasara (approved):

PR #1386, opened 2026-04-13
chore(DATAGO-131252): Make Model Configuration Optional During SAM Init

solace-pjones:
image

image

LLM provider setup is now optional
The AI Provider step in the config portal now defaults to "Setup Later", allowing users to initialize a project without configuring an LLM provider upfront.

Add Agent GUI no longer allows you to select from a static drop down. You must now type the model provider your agent will use. It will default to the general model.

When "Setup Later" is selected:
- shared_config.yaml is generated without a models section
- All other generated YAML files (orchestrator, webui, agents) omit the model field but retain model_provider references, so they're ready to use once a model is configured

When a provider is selected:
- Existing behavior is preserved — the models section is generated in shared_config.yaml with the appropriate configuration

On startup, SAM seeds the model_configuration table from whatever is defined in shared_config.yaml

Frontend (Config Portal)

- Add "Setup Later" as the default option in the LLM provider dropdown
- Remove the useEffect that auto-selected OpenAI as the default provider
- Skip validation and LLM connection test when no provider is selected
- Hide provider-specific fields (API key, endpoint, model name) when no provider is selected
- Fix CompletionStep review page to hide the AI Provider section when no provider is selected
- Preserve llm_provider in submitted form data (was being deleted by cleanDataBeforeSubmit)

Backend (CLI Init Steps)

- orchestrator_step.py: Strip models section from shared_config.yaml and model: line from orchestrator config when no provider
- webui_gateway_step.py: Strip model: line from webui config when no provider
- env_step.py: Skip LLM-related environment variables when no provider
- web_init_step.py: Skip model name mapping when no provider
- agent_cmd.py: When adding an agent, check if shared_config.yaml has a models section — if not, strip the model: anchor from the generated agent config


templates/agent_template.yaml
```diff
@@ -20,6 +20,8 @@ apps:
       agent_name: "__AGENT_NAME__"
       display_name: "__AGENT_SPACED_NAME__ Agent"
       model: __MODEL_ALIAS__
```
cyrus2281: Can we git rid of `model`?
Let's only add it with `__MODEL_ALIAS__`

solace-pjones: Fixed this

cyrus2281 (approved):
LGTM

PR #1388, opened 2026-04-13
chore(DATAGO-131834): Test Connection prior to saving

solace-pjones:
https://github.com/user-attachments/assets/2ec5120d-efd9-411a-80ae-659b948c1c53

Automatically test model connection before saving — if the test fails, a confirmation dialog warns the user and offers "Save Anyway" or "Go Back" to fix; if the test passes, the save proceeds transparently with no extra UI

The save flow in ModelEditPage.tsx was changed from a single-step save to a two-step test-then-save, all within the handleSave function:

Before: Click Save → call createModelConfig/updateModelConfig → navigate

After: Click Save → call testModelConnection silently → if test passes, call createModelConfig/updateModelConfig → navigate. If test fails, show a ConfirmationDialog with the error. User can "Go Back" to fix or "Save Anyway" to bypass.

The key changes in [ModelEditPage.tsx](vscode-webview://1097p16tosrfolka280h4s8o9f23uc5e4hidf77v9couulo0502n/client/webui/frontend/src/lib/components/models/ModelEditPage.tsx):

handleSave (line 81) — now builds the payload, runs testModelConnection() first, and only calls performSave() if the test passes. On failure, it stores the payload in a ref and opens the dialog.

performSave (line 55) — extracted from the old handleSave as a shared helper. Used both on the happy path (test passes) and when the user clicks "Save Anyway" in the dialog.

handleSaveAnyway (line 76) — closes the dialog and calls performSave with the stored payload from the ref.

ConfirmationDialog (line 153) — reuses the existing component with custom content (error banner + prompt text) and actionLabels ("Go Back" / "Save Anyway"). Title is context-aware: "Add Model Configuration Failed" vs "Save Model Configuration Failed".

No backend changes, no new components, no changes to ModelEdit.tsx or TestConnectionSection.tsx. The two API calls (validateOnly=true for test, then the actual create/update) are made sequentially from the frontend — the user only sees a dialog if something goes wrong.


client/webui/frontend/src/stories/models/ModelEditPage.stories.tsx
```diff
@@ -442,6 +444,108 @@ export const TestConnectionFailure: Story = {
...
+        // Verify dialog is closing (Radix keeps DOM element during animation)
+        expect(screen.queryByRole("dialog")).toHaveAttribute("data-state", "closed");
+
+        // Verify form is still accessible
+        const aliasInput = canvasElement.querySelector('input[name="alias"]') as HTMLInputElement;
+        expect(aliasInput).toBeTruthy();
+        expect(aliasInput.value).toBe("anthropic-model");
+    },
+};
+
```
JKaram: Can you convert these to `unit` tests.

client/webui/frontend/src/lib/components/models/ModelEditPage.tsx
```diff
@@ -49,28 +52,66 @@ export const ModelEditPage = () => {
...
+                setIsLoading(false);
+            }
+        },
+        [isNew, modelToEdit, navigate]
+    );
+
+    const handleSaveAnyway = useCallback(async () => {
+        setShowSaveFailedDialog(false);
+        await performSave(pendingPayloadRef.current!);
+    }, [performSave]);
```
JKaram, line 59: Can we please use react query here, it will simplify this code,

client/webui/frontend/src/stories/models/ModelEditPage.stories.tsx
solace-pjones: done

client/webui/frontend/src/lib/components/models/ModelEditPage.tsx
solace-pjones: done

client/webui/frontend/src/lib/api/models/service.ts
JKaram, line 17: not a biggie. But you could create a types folder in this `/models` directory for this.

client/webui/frontend/src/lib/api/models/hooks.ts
```diff
@@ -11,6 +12,18 @@ export interface SupportedModelsQueryParams {
...
 
+/**
+ * Hook to fetch a single model configuration by ID.
+ */
+export function useModelById(id: string | undefined) {
+    return useQuery({
+        queryKey: modelKeys.detail(id!),
+        queryFn: () => fetchModelById(id!),
+        enabled: !!id,
+        retry: 0,
```
JKaram, line 23: Is there a reason to set the retry to 0?

client/webui/frontend/src/stories/models/ModelEditPage.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+        const user = userEvent.setup();
+
+        await screen.findAllByText("Edit anthropic-model");
+
+        await user.click(screen.getByRole("button", { name: /Save/i }));
+
+        // Verify no dialog appeared — test passed silently and save proceeded
+        await waitFor(() => {
+            expect(screen.queryByRole("dialog")).toBeNull();
+        });
```
JKaram, line 85: It might be useful to check if the UI is sending the correct payload.

JKaram (approved):
Minor comments. LGTM

PR #1393, opened 2026-04-14
chore(DATAGO-131778): Enable Model Config UI feature by default by removing feature flag

solace-pjones:

Enable the Model Configuration UI by default by flipping the model_config_ui feature flag from false to true

Remove legacy LLM_SERVICE_* environment variable placeholders from the Dockerfile, since model configuration is now managed through the UI

Feature flag default (features.yaml)
Changed the model_config_ui feature flag default from false to true. This means the Models tab in the Agent Mesh interface is now visible out of the box without requiring users to set SAM_FEATURE_MODEL_CONFIG_UI=true. Users can still opt out by setting SAM_FEATURE_MODEL_CONFIG_UI=false.

Dockerfile cleanup
Removed four LLM_SERVICE_* placeholder environment variables that are no longer needed:

LLM_SERVICE_ENDPOINT
LLM_SERVICE_API_KEY
LLM_SERVICE_PLANNING_MODEL_NAME
LLM_SERVICE_GENERAL_MODEL_NAME
These were previously required for users to configure LLM access before deploying. With the Model Configuration UI now enabled by default, users configure their model providers directly through the UI instead.


cyrus2281 (approved):
Looks Good

Don't forget to update the enterprise repository dockerfile and features.yaml as well

PR #1395, opened 2026-04-14
fix(DATAGO-131970): azure-gpt-5 model connection fails test connection when it should pass

solace-pjones:

Stop forcing temperature=0.1 during test connection — many reasoning models (e.g. GPT-5) reject non-default temperature values. Temperature is now only sent if the user explicitly configured one.

Increase test connection max_output_tokens from 50 to 256 — reasoning models consume tokens internally for chain-of-thought before producing visible output, causing "No response from LLM" with the previous budget.

Fix indentation bug in _get_completion_inputs where the generation_params = None fallback was inside the parameter extraction loop, causing a NoneType crash when temperature was the first (absent) key.

Removed default temperature in test connection (model_config_service.py):

Increased max_output_tokens from 50 to 256 (model_config_service.py):

Fixed indentation bug in _get_completion_inputs (lite_llm.py):

### DIFF
--- a/src/solace_agent_mesh/agent/adk/models/lite_llm.py
+++ b/src/solace_agent_mesh/agent/adk/models/lite_llm.py
@@ -872,8 +872,8 @@ def _get_completion_inputs(
                 mapped_key = param_mapping.get(key, key)
                 generation_params[mapped_key] = config_dict[key]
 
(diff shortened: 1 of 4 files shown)


cyrus2281 (approved):
LGTM

PR #1397, opened 2026-04-14
feat(DATAGO-131827): add ArtifactMonitor for tracking artifact service operations

cyrus2281:

Add ArtifactMonitor for tracking performance of artifact-related operations. This provides observability metrics for save, load, list, delete and other artifact service functions, helping to identify performance bottlenecks or reliability issues.

- Added new `ArtifactMonitor` class to track artifact service operations with proper metrics
- Applied the monitors with `MonitorLatency` to all methods in `ScopedArtifactServiceWrapper` 
- feat(observability): add ArtifactMonitor for tracking artifact service operations
- Updated frontend package versions from 2.5.0 to 2.4.1
- Removed several unused components/features: model override functionality, inline vision capability, test connection infrastructure, and unused hooks/services

The change uses the existing OperationMonitor pattern to create a consistent monitoring approach. This ensures metrics are properly labeled with component type ("artifact"), component name ("artifact_service") and operation name (e.g. "save", "load"). This monitoring approach was chosen for consistency with existing monitoring practices.

- [x] Manual testing: Testing artifact operations in local environment to verify metrics are recorded
- [x] Unit tests: Added comprehensive unit tests for ArtifactMonitor in test_artifact_observability.py
- [ ] Integration tests: Not applicable
- [ ] Known limitations: Error tracking still relies on generic categories

The PR contains some cleanup of unused features which could be separated in follow-up PRs if needed.

image


src/solace_agent_mesh/common/observability.py
```diff
@@ -81,4 +81,69 @@ def create(cls, name: str) -> MonitorInstance:
             component_type="tool",
             component_name=name,
             operation="execute"
-        )
\ No newline at end of file
+        )
+
+
+class ArtifactMonitor(OperationMonitor):
```
enavitan: I propose using here: RemoteRequestMonitor.

This monitor would include service_name and operation_name, which should be sufficient for this use case.

The reasoning is that OperationMonitor is better suited for aggregating multiple components of the same category, for example:

multiple agents
multiple tools
multiple connectors

In that model, it helps answer questions such as:

How are my agents performing?
How are my tools performing?
How are my connectors performing?

For Artifactory, however, we are dealing with a single external service rather than a group of equivalent components.

src/solace_agent_mesh/agent/adk/services.py
```diff
@@ -170,12 +177,13 @@ async def list_artifact_versions(
...
+        with MonitorLatency(ArtifactMonitor.list_artifact_versions()):
+            return await self.wrapped_service.list_artifact_versions(
+                app_name=scoped_app_name,
+                user_id=user_id,
+                filename=filename,
+                session_id=session_id,
+            )
 
     @override
     async def get_artifact_version(
```
enavitan, line 189: an alternative way would be:

```
@MonitorLatency(ArtifactMonitor.xxx())
async def ...
```

up to you.

src/solace_agent_mesh/common/observability.py
```diff
@@ -81,4 +81,69 @@ def create(cls, name: str) -> MonitorInstance:
...
+        """Create monitor instance for list_versions operation."""
+        return cls._create("list_versions")
+
+    @classmethod
+    def list_artifact_versions(cls) -> MonitorInstance:
+        """Create monitor instance for list_artifact_versions operation."""
+        return cls._create("list_artifact_versions")
+
+    @classmethod
+    def get_version(cls) -> MonitorInstance:
```
enavitan: lets reduce cardinality:
load(),
list() -- this would colapse all list operation - i expect these would be trivial/fast;
delete(), 
save(), 


this way we would drop from 7 to 4 values - which would still reflect artifact perf.

enavitan (commented):
looks good Cyrus, I propose to switch from OperationMonitor to RemoteMonitor.

Thanks
Eugene

src/solace_agent_mesh/agent/adk/services.py
```diff
@@ -196,7 +196,7 @@ async def get_artifact_version(
         version: Optional[int] = None,
     ) -> Optional[ArtifactVersion]:
         scoped_app_name = self._get_scoped_app_name(app_name)
-        with MonitorLatency(ArtifactMonitor.get_version()):
+        with MonitorLatency(ArtifactMonitor.load()):
             return await self.wrapped_service.get_artifact_version(
```
enavitan, line 200: should we treat this one as list() ? - i asusme gettng version, listing version would be similar in speed - and would leave load() - explicitely for reading the actual content from the artifactory WDYT?

enavitan (approved):
left a suggestion below.

Thank you!

PR #1399, opened 2026-04-15
fix(DATAGO-132161): LLM environment variables not seeded 

solace-pjones:

Fix a bug where _ensure_default_aliases attempted to insert duplicate "general" and "planning" records after _seed_from_env_vars had already added them, causing an IntegrityError on commit

invisible to subsequent db.query() calls within the same transaction
Solution: track seeded aliases in-memory (as a set) and pass them to _ensure_default_aliases to skip without querying the DB

_seed_from_env_vars and _seed_from_models_config now return a set of seeded alias names instead of an int count
_ensure_default_aliases accepts a seeded_aliases parameter and skips any alias already in that set (avoiding the uncommitted query problem)
_ensure_default_aliases returns its own set of created placeholders
Final count computed from existing_count + len(seeded_aliases) instead of a DB query (also affected by autoflush=False)

### DIFF
--- a/src/solace_agent_mesh/services/platform/services/model_configuration_seeder.py
+++ b/src/solace_agent_mesh/services/platform/services/model_configuration_seeder.py
@@ -189,32 +189,34 @@ def seed_model_configurations(
 
     existing_count = db.query(func.count(ModelConfiguration.id)).scalar()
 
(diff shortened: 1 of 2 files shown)


cyrus2281 (approved):
LGTM

PR #1414, opened 2026-04-16
refactor(DATAGO-132190): PageLayout + self-controlling ModelWarningBanner

JKaram:

Fix a layout bug where `ModelWarningBanner` pushed the routed `<Outlet />` below the viewport, causing every in-`AppLayout` page to show an unwanted page-level scrollbar. Also tightens the banner's API so `AppLayout` does not need to know whether it will render.

- Added `PageLayout` (`src/lib/components/layout/PageLayout.tsx`) — a flex-column wrapper using `flex-1 min-h-0 overflow-hidden` that each page now uses as its root element, consuming whatever vertical space `AppLayout`'s main leaves.
- Changed `AppLayout` `<main>` from `h-full overflow-auto` to `flex h-full flex-col` so the banner + `<Outlet />` stack naturally.
- Moved all visibility logic into `ModelWarningBanner` itself: it now reads `useBooleanFlagDetails("model_config_ui")`, `useModelConfigStatus()`, and `useChatContext().hasModelConfigWrite` and returns `null` when not applicable. The component takes no props.
- Migrated every page rendered inside `AppLayout` to `<PageLayout>`: `ChatPage`, `SharedChatViewPage`, `PromptsPage`, `AgentMeshPage`, `RecentChatsPage`, `ScheduledTasksPage`, `ArtifactsPage`, `ProjectsPage`, `WorkflowVisualizationPage`, `ModelDetailsPage`, `ModelEditPage`.
- Updated `AppLayout.test.tsx`, `ModelWarningBanner.test.tsx`, and `ModelSetupDialog.stories.tsx` to the new banner API — using `StoryProvider` / `OpenFeatureTestProvider` + MSW for `/api/v1/platform/models/status`, plus async `findByText` for OpenFeature's async flag resolution.

- **Why a shared `PageLayout` instead of fixing sizes per page?** Pages had inconsistent roots (`h-full` vs `h-screen`) which is exactly what caused the overflow when the banner stole ~44px. A single wrapper puts the container contract in one place and lets future siblings of the `<Outlet />` coexist without breaking pages.
- **Why make the banner self-controlling?** `AppLayout` was fetching model config status and threading `showWarning` / `hasModelConfigWrite` as props just to decide whether to render the banner. Keeping that logic inside the banner removes leaky state from `AppLayout` and means callers only need to render `<ModelWarningBanner />`.

- [x] Unit tests: `AppLayout.test.tsx` and `ModelWarningBanner.test.tsx` updated and passing (full suite: 821 tests across 79 files, all green).
- [x] Type check: `tsc --noEmit` clean.
- [ ] Manual testing: did not start dev server for this PR — verified via unit tests and type check only.
- [ ] Known limitations: storybook visual regression for the banner stories not re-run locally; the stories were updated for the new API but should be eyeballed in Chromatic/Storybook CI.

- Review the `PageLayout` migration of each page — confirm none of them relied on `h-full` vs `h-screen` semantics for internal scrolling. The pattern throughout is inner scroll containers, which are unaffected.
- `ModelSetupDialog.stories.tsx` now uses MSW + `OpenFeatureTestProvider`; confirm this matches the convention used in `ChatPage.stories.tsx` and other banner-using stories.

### DIFF
--- a/client/webui/frontend/src/AppLayout.tsx
+++ b/client/webui/frontend/src/AppLayout.tsx
@@ -27,7 +27,6 @@ function AppLayoutContent() {
     const [modelSetupDismissed, setModelSetupDismissed] = useLocalStorage("model-setup-dialog-dismissed", false);
 
     const { data: modelConfigStatus } = useModelConfigStatus();
-    const showModelWarning = modelConfigUiEnabled && modelConfigStatus && !modelConfigStatus.configured;
 
     useEffect(() => {
         if (modelConfigUiEnabled && modelConfigStatus && !modelConfigStatus.configured && !modelSetupDismissed) {
@@ -139,8 +138,8 @@ function AppLayoutContent() {
             ) : (
                 <NavigationSidebar items={topNavItems} bottomItems={bottomNavigationItems} activeItem={getActiveItem()} onItemChange={handleNavItemChange} onHeaderClick={handleHeaderClick} />
             )}
-            <main className="h-full w-full flex-1 overflow-auto">
-                <ModelWarningBanner showWarning={!!showModelWarning} hasModelConfigWrite={hasModelConfigWrite} />
+            <main className="flex h-full w-full flex-1 flex-col">
+                <ModelWarningBanner />
                 <Outlet />
             </main>
             <ToastContainer />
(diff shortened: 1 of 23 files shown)


solace-pjones (approved):
Very nice. Looks good.

PR #1417, opened 2026-04-17
fix(DATAGO-132511): investigate intermittent hanging integration tests

enavitan:

```
Integration tests simulating crashed downstream agents were hanging intermittently for hours instead of completing in ~13 minutes. Tests like test_proxy_remote_crash_no_response would sometimes pass, sometimes    
  hang forever, depending on timing.


  Root Cause                                                                                                                                                                                                           
   
  When simulating an agent crash (agent sends status but no final response), a race condition occurred between:                                                                                                        
                                
  1. Executor closing the event queue to signal "agent stopped"                                                                                                                                                        
  2. SSE Consumer detecting the closure while in a 0.5-second timeout loop
                                                                                                                                                                                                                       
  The Race: Between checking "is queue closed?" and waiting for the next event, timing determined the outcome:                                                                                                         
  - Fast timing: Consumer detects closure and exits cleanly                                                                                                                                                            
  - Slow timing: Consumer starts waiting, timeout fires, infinite loop begins 


This was intermittent because it depended on asyncio scheduler timing, CPU load, and OS scheduling.
  ```

```
  Two changes to DeclarativeAgentExecutor:                                                                                                                                                                             
   
  1. Detect All Terminal Event Types                                                                                                                                                                                   
                                
  Previously only detected Task objects with terminal states. A2A protocol also supports TaskStatusUpdateEvent.final=true to end streams. Now we detect both terminal event patterns, matching the SDK's own           
  definition.                   
                                                                                                                                                                                                                       
  2. Conditional Queue Closing  

  Instead of always using graceful close (which blocks and creates the race):
  - Terminal event sent: Use graceful close() to let consumer process all events
  - No terminal event (crash): Use close(immediate=True) which returns immediately, allowing consumer's next iteration (within 0.5s) to detect closed state and exit     
```

### DIFF
--- a/tests/integration/test_support/a2a_agent/executor.py
+++ b/tests/integration/test_support/a2a_agent/executor.py
@@ -61,8 +61,10 @@ class DeclarativeAgentExecutor(AgentExecutor):
 
                 await event_queue.enqueue_event(event_obj)
 
-                # If the event is a terminal task, stop processing more responses for this request.
-                # Non-terminal states like 'submitted' or 'working' should not stop the processing.
+                # Check if this is a terminal event (ends the stream):
+                # 1. Task with terminal state (completed/failed/canceled/rejected)
+                # 2. TaskStatusUpdateEvent with final=True
+                # These match the SDK's EventConsumer definition of is_final_event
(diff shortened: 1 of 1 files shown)


efunneko (approved):
Looks good!

PR #1438, opened 2026-04-22
fix(DATAGO-131500): Resolve model override for agent's own model alias

bwiebe-solace:

Fixes a bug where per-request model overrides failed when the override alias matched the agent's own default model. The `complete_pending_resolve()` call was inside an `else` branch, so it was skipped when the alias matched the agent's own `model_id`, causing the resolve future to never complete and the request to hang or fail.

Also fixes a secondary issue where non-LiteLLM kwargs (`cache_strategy`, `thinking`, `oauth_*`, `type`) leaked into `litellm.completion()` calls when using per-request model overrides, causing unexpected keyword argument errors.

**`dynamic_model_provider.py`** - Moved `complete_pending_resolve()` out of the `else` block so it runs for all aliases, including the agent's own `model_id`. This ensures resolve futures complete regardless of whether the alias matches the agent's default.

**`lite_llm.py`** - Introduced a `_sanitize_for_completion()` static method that strips keys not valid as `litellm.completion()` kwargs (`cache_strategy`, `thinking`, `type`, `oauth_*`). This method is called:
- During model initialization (consolidating previously scattered `pop()` calls)
- On per-request override configs before they are passed to `litellm.completion()`

**`test_dynamic_model_provider_resolve.py`** - Updated the existing test to assert that `complete_pending_resolve` is called even when the alias matches the agent's own model.

### Key Design Decisions

- **Sanitization as a shared static method**: The same keys need stripping in two code paths (init and per-request override). A shared method avoids duplication and ensures both paths stay in sync.
- **Mutating dict in place**: The sanitization method mutates and returns the dict, matching the existing pattern where callers already work with a `deepcopy`.

- [ ] Manual testing: Verified per-request model override works when alias matches agent's own model and when it differs
- [x] Unit tests: Updated `test_different_model_routes_to_resolve` to verify `complete_pending_resolve` is always called
- [ ] Known limitations: No integration test for the sanitization path with live LiteLLM

- The `complete_pending_resolve()` call now runs unconditionally (not just for "different" aliases). Verify this doesn't cause issues if the same alias triggers both `update_litellm_model` and `complete_pending_resolve` in the same message cycle.
- The `_sanitize_for_completion` key list should be kept in sync if new non-completion kwargs are added to model configs in the future.

---

*Generated by Claude Code*


josephabonasara:
Claude review comments
**Comment 1** — src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py (lines 121-125, the new unconditional complete_pending_resolve call)
When model_config is falsy (the remove_litellm_model() branch above), we now also call complete_pending_resolve(topic_model_id, None). This resolves any in-flight resolve() awaiter with None, which per the docstring means "platform returns no config" — functionally equivalent to a timeout but faster.

Intentional? If a model is being actively removed while a request holds a pending resolve, returning None will likely surface as a 10s-style failure to the caller anyway, so probably fine — but worth confirming this is the desired semantics vs. leaving the future pending to retry. A one-line comment here stating the None case is an explicit "config unavailable" signal would make the intent obvious to future readers.

**Comment 2** — src/solace_agent_mesh/agent/adk/models/lite_llm.py (line 1103, on _sanitize_for_completion)
No direct unit test for this method. Please add one asserting:

cache_strategy, thinking, type are removed
all oauth_* prefixed keys are removed (including ones not in _extract_oauth_config's list, e.g. oauth_audience)
unrelated keys (model, api_key, api_base, num_retries) are preserved
Also worth asserting on the generate_content_async override path (with a mocked acompletion) that completion_args never contains any of these keys when they appear in override_config. This is the exact bug the PR fixes, and without a test it can regress silently.

**Comment 3** — src/solace_agent_mesh/agent/adk/models/lite_llm.py (line 1106, inside _sanitize_for_completion)
The explicit list (cache_strategy, thinking, type) plus the oauth_* prefix scan is fine, but this is the kind of list that silently rots. Two suggestions:

Pull the list into a module/class constant (e.g., _NON_COMPLETION_KEYS = frozenset({"cache_strategy", "thinking", "type"})) so future additions are discoverable at one site.
Consider inverting the approach later: pass through only a known-good allowlist rather than denylisting. Since LiteLLM's accepted kwargs are documented, an allowlist would be safer than chasing new config keys as they're added. Not necessary for this PR — just a note.

**Comment 4** — src/solace_agent_mesh/agent/adk/models/lite_llm.py (lines 1098-1104, the thinking extraction)
You switched from pop("thinking", None) to get("thinking"), relying on the later _sanitize_for_completion() call to remove it. That's correct, but the ordering dependency is now implicit. If someone later moves the sanitize call earlier (or adds an early return/raise between here and the sanitize call), thinking would leak into self._model_config.

A short comment like # thinking is removed by _sanitize_for_completion() below would make the ordering invariant explicit. Same applies to the cache_strategy .get() on line 1085 and _extract_oauth_config on line 1106.

**Comment 5** — tests/unit/agent/adk/models/test_dynamic_model_provider_resolve.py (line 142, test_own_model_updates_litellm)
The test name is now slightly misleading — it verifies both the primary model update path and the resolve completion. Consider renaming to test_own_model_updates_litellm_and_completes_resolve or adding a sibling test specifically for the "override alias == agent's own model_id, with pending resolve" end-to-end scenario (using a real DynamicModelProvider + mocked broker, asserting the resolve() future actually resolves rather than just that complete_pending_resolve is called).

**Comment 6** (optional, nit) — src/solace_agent_mesh/agent/adk/models/lite_llm.py (lines 1084, 1112, 1198)
Minor: trailing-whitespace-only changes on these lines add noise to the diff. Not a blocker, but running ruff format (or equivalent) in a separate commit would keep this PR focused.

josephabonasara (approved):
LGTM

PR #1442, opened 2026-04-23
fix(DATAGO-133094): reducing status calls

lgh-solace:

> User-visible symptom: With multiple concurrent background tasks running, the WebUI becomes unusable — and the slowdown spreads to other users who happen to be served by the same gateway replica. Server logs also show SSE viz queue full for stream viz-stream-… warnings during the same periods.

**Root cause**
The culprit was the frontend's useBackgroundTaskMonitor hook. It existed to poll for background-task completion when SSE wasn't connected to the relevant session, but the implementation amplified load in two ways:

* Per-task fan-out. Every 5 seconds it walked the entire tracked-task list and fired one GET /api/v1/tasks/{id}/status per task sequentially. A user with N tasks = N round-trips per poll = N DB hits every 5 s.
* No re-entrancy guard. setInterval doesn't await the async tick, so when the backend slowed down, rounds stacked on each other — more load created even more load.
* All of this landed on the gateway process's asyncio event loop — one per replica. That loop also runs the visualization message processor and the task-logger loop. Users who share a replica (SSE sessions pin to a replica via affinity because of per-process in-memory state — viz queues, stream maps, metadata caches) compete for the same loop. Polling traffic starved the viz processor, its per-stream queues backed up and started evicting messages, producing the viz-queue-full warnings. Cross-user slowdown was coroutine starvation on a shared loop within the replica.

* Polling fan-out → single batched request. Replaced N-per-user sequential /tasks/{id}/status calls every 5 s with one /tasks/background/active call every 10 s. Tasks missing from the batch get one follow-up /status call. Request volume dropped 10-30× per user.

* Stacking ticks → re-entrancy guard. Added an effect-scoped flag that makes a tick a no-op if a previous one is still in flight. Breaks the feedback loop where backend slowness produced more load.

* Polling now ignores tasks whose session the user is currently viewing. SSE is authoritative there; polling only covers sessions the user isn't looking at. This alone stopped the replayBufferedEvents race that was clobbering in-flight agent responses.

* False completions on transient DB misses → is_running guard. handleTerminalTask now verifies status.is_running === false before firing callbacks. A task transiently absent from the batch response but still running no longer gets incorrectly unregistered.

* Session-ID discarded from server → plumbed through. Server-fetched tasks now preserve session_id instead of hardcoding "", fixing session-scoped indicators and eliminating an extra /status fallback in useAutoGenerateTitle.

* Session-delete polling leak → unregister on delete. ChatProvider.deleteSession now drops any locally-tracked tasks for the deleted session so polling stops immediately for the current tab.

* Abort + auth hygiene. All in-flight fetches are aborted on effect teardown, and the tick bails instead of making unauthenticated calls when userId is null.

N/A


client/webui/frontend/src/lib/hooks/useBackgroundTaskMonitor.ts
```diff
@@ -1,13 +1,52 @@
...
- * Hook for monitoring and reconnecting to background tasks.
- * Stores active background tasks in localStorage and automatically reconnects on session load.
- */
-
 import { useState, useEffect, useCallback, useRef } from "react";
 import { api } from "@/lib/api";
 import type { BackgroundTaskState, BackgroundTaskStatusResponse, ActiveBackgroundTasksResponse, BackgroundTaskNotification } from "@/lib/types/background-tasks";
 
 const STORAGE_KEY = "sam_background_tasks";
+const POLL_INTERVAL_MS = 10_000;
```
lgh-solace: Dropped to 10s but know only polls for chat sessions that are not current and will refresh when the user switches to it.

client/webui/frontend/src/lib/hooks/useBackgroundTaskMonitor.ts
```diff
@@ -1,13 +1,52 @@
...
+const REGISTRATION_GRACE_MS = 3000; // A task that disappears from the active-tasks is treated as terminated only after this period
+
+type TerminationKind = "completed" | "failed" | "timeout";
+
+// ============ Helpers ============
```
lgh-solace, line 12: Extracted to reduce complexity.

amir-ghasemi:
1. Grace-period math is broken for server-fetched tasks —    
  useBackgroundTaskMonitor.ts (selectTerminationCandidates,
  ~line 52)                                                    
  now - task.startTime >= REGISTRATION_GRACE_MS uses the    
  server-supplied task.start_time. For a task the client just  
  discovered on mount but that started hours ago on the server,
   the grace window is already expired, so the is_running guard
   is the only protection against false completion on a     
  transient batch miss. That guard does help, but the grace
  window was supposed to be independent defense-in-depth.
  Fix: add a client-side registeredAt = Date.now() and grace
  against that.                                                
  
  2. Permanent skip when session is active but SSE is actually 
  dead — useBackgroundTaskMonitor.ts (skip logic, ~line 50 of
  diff)                                                        
  Polling skips any task whose sessionId === activeSessionId on
   the premise that "SSE is authoritative." The hook has no    
  signal that SSE is actually connected/healthy — on reconnect
  failure, tab-visibility throttling, or a dropped SSE, the    
  task is skipped on every tick and never gets its terminal 
  callback.
  Fix: gate the skip on a known-alive SSE signal, or add a
  staleness timeout after which polling takes over.                                                         
                                                            
  3. deleteSession race → spurious completion toast —          
  ChatProvider.tsx (deleteSession, ~line 1196)              
  unregisterBackgroundTask is called after await               
  api.webui.delete(...). During that round-trip a poll tick can
   drop the task from the batch and fire onTaskCompleted →
  "Background task completed" notification for a session the   
  user just deleted.                                        
  Fix: unregister optimistically before the DELETE, or filter
  pending-delete sessions in selectTerminationCandidates.      
  
  4. Public checkAllBackgroundTasks has no unmount guard —     
  useBackgroundTaskMonitor.ts:~312                          
  Creates a local AbortController that is never aborted and    
  never registered with any teardown. If the hook unmounts     
  mid-flight, handleTerminalTask still runs setNotifications /
  unregisterBackgroundTask and fires onTaskCompleted on a      
  torn-down consumer. Flagged independently by both the bug and
   concurrency agents.
  Fix: track the controller in a ref that the unmount effect
  aborts, or add a mountedRef bail before setState.            
  
  5. Stale session-check after await —                         
  useBackgroundTaskMonitor.ts (handleTerminalTask)          
  The PR's "we only notify for tasks whose session you aren't  
  viewing" guarantee holds at batch time, but                  
  handleTerminalTask awaits a /status fallback. If the user
  switches into the task's session during that await, the      
  notification still fires.                                 
  Fix: re-check currentSessionIdRef.current !== task.sessionId
  after the await.                                             
  
  6. No immediate kick-off tick — useBackgroundTaskMonitor.ts  
  polling effect (~line 218-234)                            
  The old hook ran an immediate check; the new one waits a full
   POLL_INTERVAL_MS (10s) before the first tick. A task        
  registered just before idle won't be checked for 10s.
  Fix: call tick() once synchronously before setInterval (cheap
   and preserves prior UX).                                    
          
  7. setInterval + re-entrancy guard still queues JS tasks —   
  useBackgroundTaskMonitor.ts                               
  If a tick exceeds 10s, the interval keeps firing and each    
  fires the guard no-op. Functionally correct but wasteful.    
  Converting to recursive setTimeout eliminates the pile-up
  cleanly.

client/webui/frontend/src/lib/hooks/useBackgroundTaskMonitor.ts
```diff
@@ -1,13 +1,53 @@
...
+        return { type: "timeout", message: "Background task timed out" };
+    }
+    return { type: "completed", message: "Background task completed" };
+}
+
+// Returns the subset of tracked tasks that should be treated as terminal based on the following criteria:
+// * not in current session (unless SSE has gone stale)
+// * not in server's list
+// * past grace period
+export function selectTerminationCandidates(tracked: BackgroundTaskState[], activeTaskIds: Set<string>, activeSessionId: string, now: number): BackgroundTaskState[] {
```
lgh-solace, line 38: Exported for testing purposes.

amir-ghasemi:
Finding #2 for reference was: `skip now gated on now - lastEventTimestamp < SSE_STALE_MS (60s),  re-checked post-await     `

Issues:
  1. Removed background-task-completed listener in ChatProvider. The provider no longer consumes
  the event (only dispatches it). Previously this handler replayed buffered SSE events /
  refreshed loadSessionTasks when the user was viewing the task's session. Combined with
  finding 2's fix, polling can now fire a completion callback for the currently-viewed
  session when SSE goes stale — the user gets a toast but no chat replay/refresh. If SSE
  reliably delivers terminal events this is fine; if not, you've just introduced the exact
  scenario 2 was meant to cover but without the UI update. Recommend verifying SSE terminal
  delivery or restoring a lightweight refresh path.

  2. Related stale comment: onTaskCompleted comment in ChatProvider still claims "only fires for
  tasks whose session the user is not currently viewing" — no longer true with the SSE-stale
  fallback.

  3. Test gap: No coverage of same-session completion detected via polling end-to-end through ChatProvider

amir-ghasemi (approved):
LG, left a couple of comments for consideration.

PR #1444, opened 2026-04-23
ci(DATAGO-133169): fix: convert SonarQube hotspot step from uses: docker:// to docker run

johnvincentcorpuz:
## Problem

The `sonarqube-hotspots` job in `release-readiness-check.yaml` used:

\`\`\`yaml
uses: docker://ghcr.io/solacelabs/maas-build-actions:latest
\`\`\`

GitHub Actions pulls `uses: docker://` images during **job setup**, before any steps execute — so there is no opportunity to authenticate first. Cross-repo GHCR packages fail with `unauthorized` at the manifest level even with `packages: read` scope and the repo whitelisted in package settings, because the infrastructure-level pull does not perform an explicit Bearer token exchange.

Failing run: https://github.com/SolaceLabs/solace-agent-mesh/actions/runs/24849687154/job/72753614680

## Fix

Replace `uses: docker://` with explicit `docker login` + `docker run`, using the same DOCKER_CONFIG isolation pattern already established in SPW's `hatch_release_security_checks.yml`:

- Isolated `DOCKER_CONFIG` via `mktemp` (no credential leakage to shared runner config)
- `docker login ghcr.io` with `github.token` before pull
- `docker run` with env vars forwarded via `-e` flags
- Logout + cleanup on `EXIT` trap

## Test

Validated with a dedicated test workflow on this branch before cleanup:
https://github.com/SolaceLabs/solace-agent-mesh/actions/runs/24853555218 ✅


artyom-morozov (approved):
👍

PR #1456, opened 2026-04-26
fix(DATAGO-133775): Fix authentication failure for scheduled SAM reports against Salesforce

amir-ghasemi:
This pull request enhances the scheduled task execution flow in the scheduler service by ensuring that scheduled tasks use a more accurate and user-preserving authentication context, aligning scheduled and interactive user config resolution. The changes improve RBAC, audit, and downstream agent compatibility by passing the real scheduling user and gateway context, and by making the gateway ID configurable.

**User identity and authentication context improvements:**

* The scheduler now constructs a more complete `user_identity` object (including `user_info` with `email`, `user_id`, and authentication method) to match the format used in interactive requests, ensuring downstream agents can properly resolve per-user behavior.
* The authentication mode is explicitly set to `"scheduled"` in the `gateway_context`, and the real scheduling user is preserved (not replaced with a synthetic identity), improving RBAC scope resolution and audit trail accuracy for scheduled tasks.

**Gateway ID handling:**

* The `SchedulerService` now accepts a `gateway_id` parameter, falling back to a synthetic ID only for legacy callers, allowing RBAC and role lookups to consistently hit the correct authorization service. [[1]](diffhunk://#diff-bb5851ef077008c409bc8cbe65fd16b81b8e3d840c96f64adc42d9db56e0110dR83-R92) [[2]](diffhunk://#diff-c0dc5dfe164069e3fe9586ec25f99aed7c56b80e33f6ff2e417fdd8e86444b80R1641)
* The `gateway_id` is now passed from the component to the scheduler service, ensuring consistent context across the system. [[1]](diffhunk://#diff-c0dc5dfe164069e3fe9586ec25f99aed7c56b80e33f6ff2e417fdd8e86444b80R1641) [[2]](diffhunk://#diff-bb5851ef077008c409bc8cbe65fd16b81b8e3d840c96f64adc42d9db56e0110dR83-R92)

**Logging and error handling:**

* Logging has been improved to include the authentication mode and to clarify the consequences of user config resolution failures for scheduled tasks, especially regarding peer-agent calls that require user-delegated credentials.


src/solace_agent_mesh/gateway/http_sse/services/scheduler/scheduler_service.py
```diff
@@ -984,37 +990,70 @@ async def _resolve_user_config_for_task(
...
+                "user_id": effective_user,
+                "email": user_email or effective_user,
+                "name": effective_user,
+                "authenticated": True,
+                "auth_method": "scheduled",
+            }
+            user_identity = {
+                "id": effective_user,
+                "name": effective_user,
+                "email": user_email or effective_user,
```
efunneko: This `is_email` / `user_email` logic is dead — the `or effective_user` fallback collapses both branches to the same value:

- If `is_email` is true: `user_email = effective_user`, so `user_email or effective_user == effective_user`.
- If `is_email` is false: `user_email = None`, so `user_email or effective_user == effective_user`.

So `user_info["email"]` and `user_identity["email"]` are always set to `effective_user` regardless of whether it looks like an email. For non-email user ids (e.g. the `"system-scheduler"` fallback at line 1009, or any UUID-style id), this populates the email field with a non-email string, which is exactly what the Salesforce agent's `_get_user_email` (referenced in the docstring above) is trying to use.

Suggest one of:

```python
"email": user_email,  # None when not an email — let downstream decide
```

or drop the `is_email` check entirely if `effective_user` is always email-shaped in practice. As written the guard is misleading.

efunneko (commented):
Thanks for the fix — the root cause (synthetic `scheduler_<instance>` gateway_id breaking RBAC) and the chosen approach (thread the host gateway_id through and mirror the interactive `_extract_initial_claims` shape) both look right. Two things worth addressing before merge:

**1. Dead `is_email` branch (inline comment below).** The `is_email` / `user_email` computation is effectively a no-op because of the `or effective_user` fallback — non-email user_ids still land in `user_info["email"]`. Since the stated motivation for this change is that the Salesforce agent reads `_user_identity["user_info"]["email"]`, populating that field with a non-email value (e.g. `"system-scheduler"`, a UUID, or a session id) will silently produce wrong behavior downstream rather than a clean "email not available". Either drop the guard or honor it.

**2. No test coverage for the new behavior.** `tests/unit/gateway/http_sse/services/test_scheduler_service.py` doesn't exercise `_resolve_user_config_for_task`, and a grep across the test tree shows nothing else does either. Given this is on an RBAC-adjacent path and the bug being fixed was exactly a wrong value in `gateway_context`, a small unit test that mocks `MiddlewareRegistry.get_config_resolver()` and asserts the shape of `user_identity` and `gateway_context` (specifically: `gateway_id` is the host gateway id, `auth_mode == "scheduled"`, `user_info.email` is populated correctly) would lock in the fix and prevent regression.

**Minor / non-blocking:** worth confirming in the ticket that `task.user_id` is server-authored at create time, not client-supplied — it now lands in `user_info["email"]` and the `userId` user-property, so a tampered value would impersonate another user's RBAC scope on scheduled fires. (This surface existed before for `userId`, but it widens here.)

src/solace_agent_mesh/gateway/http_sse/services/scheduler/session_resolver.py
```diff
@@ -0,0 +1,79 @@
+"""Helpers to map scheduled-task chat session IDs to their artifact storage location.
```
efunneko, line 1: This module appears to be dead code as of this PR.

A repo-wide grep for `session_resolver`, `resolve_execution_context`, `is_execution_session`, and `get_execution_from_db` finds only:

- this file
- `routers/artifacts.py`, which still defines and uses its own *private* copies (`_is_execution_session`, `_get_execution_from_db`, `_resolve_execution_context` at lines 112–174)
- `tests/unit/gateway/http_sse/routers/test_artifact_resolution.py`, which imports and tests the private functions from `artifacts.py`, not this module

So this file is a duplicate of three helpers that already exist and are still used in their original location. It's also unrelated to DATAGO-133775 (the auth-context fix this PR is titled for).

Suggest either:

1. **Finish the refactor in this PR**: update `routers/artifacts.py` to import from `session_resolver`, delete the private duplicates there, and repoint `test_artifact_resolution.py` at the public symbols. Worth doing only if it's truly trivial.
2. **Drop this file from this PR** and land the refactor separately so the auth fix isn't gated on it and the diff stays focused.

Option 2 is probably the cleaner path given this PR is otherwise tightly scoped to the scheduled-tasks auth issue.

efunneko (commented):
Re-reviewed at 689328f. Both prior points are resolved cleanly:

**1. Email logic** — `_resolve_user_config_for_task` now conditionally adds the `email` key only when `effective_user` actually looks like an email. The previous `is_email`/`or effective_user` collapse is gone, and non-email user ids (`system-scheduler`, UUIDs, session ids) no longer pollute `user_info["email"]`. Downstream agents will fail-fast as intended. ✓

**2. Test coverage** — `TestResolveUserConfigForTask` adds 8 focused tests covering exactly the regression-prone surface: host `gateway_id` forwarding, synthetic fallback for legacy callers, `auth_mode`/`scheduling_user_id` signaling, email present iff email-shaped, exception path returns `None`, effective-user precedence (including empty-string handling), and `user_profile` round-trip. The assertions match the implementation. (I couldn't execute them locally due to an unrelated import error in my venv, but CI will cover.) ✓

**One new concern** — the latest commit also added `src/solace_agent_mesh/gateway/http_sse/services/scheduler/session_resolver.py` (79 lines), but a repo-wide grep shows nothing imports it. `routers/artifacts.py` still defines and uses its own private `_is_execution_session`, `_get_execution_from_db`, and `_resolve_execution_context` (lines 112–174), and `test_artifact_resolution.py` tests those private functions, not the new module. So the new file is a dead duplicate. See inline comment for details — either finish the refactor (point `artifacts.py` and its test at the new module, delete the duplicates) or drop the file from this PR, since it's unrelated to DATAGO-133775.

The original auth fix is good to go once the dead-code question is resolved.

amir-ghasemi:
Thanks for the careful look. The dead-code claim was accurate at the snapshot you reviewed (`689328f1e`) but is now stale at HEAD — the consumer refactor landed in `6c164e0e4` ("fix: artifacts in shared scheduled tasks") immediately after.

At HEAD, `session_resolver.py` has three real consumers:

```
routers/artifacts.py:112       from ..services.scheduler.session_resolver import (
                                   get_execution_from_db as _get_execution_from_db,
                                   is_execution_session as _is_execution_session,
                                   resolve_execution_context as _resolve_execution_context,
                               )
routers/share.py:36            from ..services.scheduler.session_resolver import resolve_execution_context
services/share_service.py:42   from .scheduler.session_resolver import resolve_execution_context
```

There are no local definitions of these functions in `artifacts.py` anymore:

```bash
$ grep -nE "^def _is_execution_session|^def _get_execution_from_db|^def _resolve_execution_context" src/solace_agent_mesh/gateway/http_sse/routers/artifacts.py
(no matches)
```

The aliased imports at `artifacts.py:112-115` are exactly what made your grep look like local defs — fair confusion, the underscore prefix makes them indistinguishable from privates at a glance.

**To prevent this confusion next time**, I just pushed a follow-up that retargets `tests/unit/gateway/http_sse/routers/test_artifact_resolution.py` to import directly from `services.scheduler.session_resolver` instead of through `routers.artifacts`'s aliased re-exports. The test now reads:

```python
from solace_agent_mesh.gateway.http_sse.services.scheduler.session_resolver import (
    is_execution_session as _is_execution_session,
    resolve_execution_context as _resolve_execution_context,
)
```

Same coverage (17/17 tests still pass), but the dependency is now obvious to any reader.

On scope: I hear you that the refactor is tangential to DATAGO-133775. It got bundled because the consumer refactor (artifacts/share/share_service) needed the helper extracted to be reachable from multiple places, which surfaced while wiring up scheduled-task ownership checks in `_user_owns_execution_session`. Happy to split it into a separate PR if you'd prefer the auth fix stay strictly minimal — let me know and I'll cherry-pick it out.

efunneko (approved):
Thanks!

PR #1461, opened 2026-04-27
feat(DATAGO-130042): synthetic-monitor auth path for Datadog smoke tests

JKaram:

Adds a config-gated, non-interactive auth path so Datadog Synthetics (and any future machine client) can authenticate against the gateway without driving the SSO UI. Enables an upcoming end-to-end smoke test that sends a chat prompt and asserts a non-empty LLM response.

solace-chat today only supports OIDC Authorization Code flow against Entra ID — every authenticated caller must go through the user login UI. That doesn't work for synthetic monitoring.

New module `src/solace_agent_mesh/shared/auth/synthetic.py`:
- Loads Datadog client app GUIDs, role name, audience, tenant ID, and an endpoint allowlist from component config. Fail-closed (any missing required field, empty appid allowlist, or empty endpoint allowlist disables the path entirely).
- Validates Entra-issued client-credentials JWTs locally via `PyJWKClient` (existing dependency).
- Distinguishes "not a synthetic token" (`SyntheticTokenNotApplicable`, caller falls through to other auth paths) from "synthetic token but invalid" (`SyntheticTokenInvalid`, caller hard-rejects). This boundary prevents an attacker from probing the validator by varying claims.

Modified `src/solace_agent_mesh/shared/auth/middleware.py`:
- Loads synthetic config once at startup; logs a single info line when enabled.
- New `_try_synthetic_auth` runs ahead of the existing `sam_access_token` and IdP branches.
- Maps a valid synthetic token to a fixed principal (`user_id = "synthetic-monitor"`, non-routable email, `is_synthetic` flag) so downstream code that expects a `current_user` keeps working unchanged.
- Endpoint allowlist (method + path regex) enforced at the auth layer before any handler runs. Default deny.

### Key Design Decisions

- **Local JWKS validation** rather than delegating to the external auth service. Synthetic auth is security-sensitive (strict role match, appid allowlist, mixed-claim rejection) and co-locating validation with principal construction and endpoint enforcement makes it easier to audit.
- **Hard reject (don't fall through) on synthetic-shaped invalid tokens.** A token claiming to be synthetic that fails validation is suspicious — falling through to other auth paths would let an attacker probe by varying claims.
- **Strict role equality** (`roles == [config.role_name]`) rather than `contains`. Prevents privilege creep if a future role is granted alongside.
- **Non-routable synthetic email** (`synthetic-monitor@synthetics.invalid`) — can't collide with a real user, can't be probed via account enumeration.
- **Fail-closed config loader** — refuses to enable with empty allowlists. An empty appid allowlist would otherwise accept any tenant app with the role; an empty endpoint allowlist would deny everything.
- **`is_synthetic` schema flag and conversation cleanup are deferred as stretch goals.** Synthetic conversations are identified by the well-known `user_id = "synthetic-monitor"`; cleanup can be added later without touching auth.

- [x] Unit tests: `tests/unit/shared/auth/test_synthetic.py` (29 tests, all passing in 0.44s)
  - Discriminator boundary (NotApplicable vs Invalid)
  - Every rejection path: alg=none, wrong audience/issuer/tenant, appid not allowlisted, extra roles, missing role, expired token, each user-identity claim parameterized
  - Happy path returns expected claims
  - Config fail-closed cases (disabled-by-default, missing field, empty allowlists)
  - Endpoint allowlist incl. method mismatch and path-extension attempts
- [x] Tests use a real RSA keypair generated per session — PyJWT signature/claim validation is actually exercised, not mocked
- [ ] Integration tests against a running gateway: not yet — gated on Phase 2 (Entra setup)
- [ ] Known limitation: rate limiting for the synthetic principal isn't implemented in this PR (lives elsewhere in the stack); needs a separate decision

- **Security review of `synthetic.py`**: each defense-in-depth check (signature, alg allowlist, iss, aud, tid, strict roles, appid allowlist, no-user-claims, fail-closed config) is independently load-bearing. Removing any one check creates a real vulnerability.
- **Ordering in middleware**: the synthetic branch runs first when enabled; it must remain that way so synthetic tokens don't accidentally take the IdP fallback path.
- **Disabled by default.** No behaviour change for existing deployments. Each environment (staging, prod) opts in via config:
  ```yaml
  synthetic_auth_enabled: true
  synthetic_auth_tenant_id: <entra-tenant-guid>
  synthetic_auth_audience: api://<solace-chat-app-id-uri>
  synthetic_auth_role_name: Synthetics.Smoke
  synthetic_auth_appid_allowlist: [<datadog-sp-appid-guid>]
  synthetic_auth_endpoint_allowlist:
    - { method: GET, path: "^/api/v1/sessions$" }
    - { method: POST, path: "^/api/v1/messages$" }
    # plus whatever the chat smoke actually needs
  ```
- **Endpoint allowlist contents are placeholders** — the real list depends on what the chat send-message + response-stream endpoints are exactly. Will be tightened during Phase 2 of the rollout once the smoke test is wired up.
- The middleware lives in the shared `solace-agent-mesh` package, so the synthetic path is technically available to any deployment that opts in via config — not just solace-chat. Default-off behaviour means this is safe, but worth flagging.


src/solace_agent_mesh/shared/auth/synthetic.py
```diff
@@ -0,0 +1,258 @@
...
+            )
+            return None
+
+        compiled_endpoints = tuple(
+            (entry["method"].upper(), re.compile(entry["path"]))
+            for entry in endpoint_list
+        )
+
+        issuer = f"https://login.microsoftonline.com/{tenant_id}/v2.0"
+        jwks_uri = f"https://login.microsoftonline.com/{tenant_id}/discovery/v2.0/keys"
```
Hugo-Pare: Do you think we could allow other OIDC providers too, similar to how it is done here?
https://github.com/SolaceDev/solace-agent-mesh-enterprise/blob/main/examples/oauth2_config.yaml#L44

JKaram: Fair point - But, Going generic across OIDC providers (Okta/Auth0/etc.) is a bigger change than the URL config though — claim names and semantics differ across IdPs. The `oauth2_config.yaml` analogy works for user-OIDC because `sub`/`email`/`name` are pretty standard, but client-credentials tokens are much less standardized. If a second IdP becomes a real requirement we can add a per-provider abstraction at that point — likely cleaner than parameterizing what's currently here.

For now this is solely for our internal Datadog Synthetic test against solace-chat, so I'd rather not preemptively build the generic shape.

amir-ghasemi (approved):

PR #1458, opened 2026-04-27
fix(DATAGO-133911): SSE event replay silently disabled — gateway logs "Database not configured" despite SQL session_service

amir-ghasemi:
Also fixes https://sol-jira.atlassian.net/browse/DATAGO-133912

This pull request introduces a new retrying wrapper for session services to improve reliability during concurrent session updates, and refactors some import patterns in the SSE router to avoid circular dependencies and unnecessary imports. The most important changes are grouped below:

**Session Service Reliability Improvements:**

* Added a new `RetryingSessionService` class in `src/solace_agent_mesh/agent/adk/services.py` that wraps any `BaseSessionService` and transparently retries `append_event` calls on stale-session race conditions, ensuring that internal ADK runner appends are also protected.
* Updated `initialize_session_service` to always wrap the base session service with `RetryingSessionService`, so all session appends benefit from automatic retry logic.

**SSE Router Import Refactoring:**

* Removed the direct import of `SessionLocal` from the top of `src/solace_agent_mesh/gateway/http_sse/routers/sse.py` to prevent unnecessary or circular imports. It is now imported locally within functions where needed.
* Added local imports of `SessionLocal` inside `_prepare_replay_events` and `_get_task_info` functions in `src/solace_agent_mesh/gateway/http_sse/routers/sse.py` for safer and more targeted usage. [[1]](diffhunk://#diff-e2999058ace134a5961032b0806e510fa994c678f3075b90935fd80311a945beR39-R40) [[2]](diffhunk://#diff-e2999058ace134a5961032b0806e510fa994c678f3075b90935fd80311a945beR129-R130)

PR #1462, opened 2026-04-27
fix(DATAGO-133950): forward unknown attrs through ADK's session-service wrappers

amir-ghasemi:
## Summary

`RetryingSessionService` and `FilteringSessionService` only proxy the explicit `BaseSessionService` API, hiding `DatabaseSessionService` instance attrs that two consumers rely on. Add a narrow `__getattr__` to both wrappers so unknown attribute lookups fall through to the wrapped service.

## Why

Two consumers reach past `BaseSessionService` into the underlying SQL service:

| Consumer | Attribute | Failure mode |
|---|---|---|
| Enterprise `SecretSessionManager.__init__` (`solace_agent_mesh_enterprise/auth/secret_session_manager.py:34`) | `database_session_factory` | **Hard** — `AttributeError` blocks agent startup whenever a SQL `session_service` is configured. This is the staging failure. |
| `common/app_base._get_db_engines_from_components` (line 167) | `db_engine` | **Silent** — `hasattr()` returns `False` through the wrapper, so the agent DB has been quietly dropping out of the broker-level DB health check ever since #1458 landed. |

Staging stack trace from the enterprise pod:

```
File "/opt/venv/lib/python3.11/site-packages/solace_agent_mesh_enterprise/auth/secret_session_manager.py", line 34, in __init__
    self._session_factory = self._session_service.database_session_factory
AttributeError: 'RetryingSessionService' object has no attribute 'database_session_factory'
```

## Fix

Single `__getattr__` on each wrapper that delegates to `self._wrapped`. `__getattr__` only fires on attribute-lookup miss, so the explicit wrapper methods (filtering / retry) still take precedence and add their behaviour; only attributes the wrapper itself doesn't define hit the passthrough. Genuinely missing attributes still raise `AttributeError`.

## Audit of other consumers

I swept the codebase for any other consumer reading non-`BaseSessionService` attributes off the session service. None found beyond the two above:

- `agent/protocol/event_handlers.py:2870`, `agent/adk/runner.py:1602/1618` — standard API calls + truthy checks only.
- `gateway/http_sse/dependencies.py:get_adk_session_service` — uses `create_session_service_from_config` directly without wrapping (#1458's design), so its consumers see the raw `DatabaseSessionService`.
- `gateway/http_sse/routers/sessions.py` and `routers/artifacts.py` — different `session_service` (gateway chat-session metadata module, not ADK).

## Test plan

- [x] New regression tests in `tests/unit/agent/adk/test_auto_summarization_services.py`:
  - `database_session_factory` reachable through `FilteringSessionService`
  - `database_session_factory` reachable through `RetryingSessionService`
  - `database_session_factory` reachable through stacked `Filtering(Retrying(SQL))` (the production composition)
  - `db_engine` reachable via `hasattr()` probe through every wrapper combo (the silent-failure case)
  - Genuinely missing attributes still raise `AttributeError`
- [x] All 17 tests in the file pass locally.
- [x] Live-reproduced the exact `service.database_session_factory` lookup that enterprise's `SecretSessionManager` performs — works through the production wrapper stack.
- [ ] Verify on staging after merge that the enterprise pod starts and that the agent DB now appears in the broker DB health check.

### DIFF
--- a/src/solace_agent_mesh/agent/adk/services.py
+++ b/src/solace_agent_mesh/agent/adk/services.py
@@ -369,6 +369,14 @@ class FilteringSessionService(BaseSessionService):
         """
         self._wrapped = wrapped_service
 
+    def __getattr__(self, name: str) -> Any:
+        # Forward backend-specific attributes (e.g. DatabaseSessionService's
+        # database_session_factory, which enterprise's SecretSessionManager
+        # reaches in for) that aren't part of BaseSessionService. Only fires
+        # on lookup miss, so the explicit wrapper methods above still take
+        # precedence and add their filtering/retry behaviour.
+        return getattr(self._wrapped, name)
+
     async def get_session(
         self,
         *,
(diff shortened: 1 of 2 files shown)


efunneko (approved):
LG

PR #1464, opened 2026-04-27
fix(DATAGO-133967): SSE viz queue overflow under high-fan-out tasks — UI progress freezes/skips events

amir-ghasemi:
## Summary
- Build + serialize each viz event **once per A2A message** instead of once per subscribed stream — eliminates O(streams × payload_size) CPU under the viz lock.
- Producer-side short-circuit when `metadata.visualization=false` so excluded events don't enter the fan-out at all.
- Drop the drain-side `_include_for_visualization` filter that re-parsed every dequeued event from JSON only to check the same flag.
- Bump `sse_max_queue_size` default 200 → 1000 as a tail-burst safety margin.

## Why
Datadog (solace-chat prod, 2026-04-26 14:43–15:01 UTC, single pod `i-0d08a319f28ac72e8`) showed 11 active viz streams with sustained per-stream queue overflow:

| Stream | Peak evictions | Active duration |
|---|---|---|
| c4fbcc2f… | 1,180 | 49 min |
| d37a8cf0… | 850 | 50 min |
| 8 streams synchronized at 14:43:49 | 320 each | ~17 min |

Aggregate: **3,380 messages dropped in 11.5 min on one pod** (~4.9/s sustained). The "Visualization message queue is over 90% full" warning never fired — the broker→processor ingest queue is fine. The bottleneck is purely **per-client SSE fan-out**: with multiple dashboards subscribed to one heavy task (138 tool calls + sub-agents), the event loop couldn't push fast enough through the per-stream 200-deep queues, even though the upstream queue was empty.

Root cause in `_visualization_message_processor_loop` (component.py): `sse_event_payload` build + `json.dumps()` ran **inside** the per-stream `for` loop while holding the viz lock. With 10 streams × ~50 KB payloads, that's ~500 KB of synchronous serialization per A2A message, no `await`. The drain side then ran `json.loads()` on every dequeued event just to read `metadata.visualization`. Net: producer and drain CPU both scaled with stream count × payload size.

## Test plan
- [x] `python -m pytest tests/unit/gateway/http_sse/` — **1242 passed, 23 skipped**
- [x] Helper smoke-tested across 9 cases (string `"false"`/`"FALSE"`, bool `False`, `"true"`, missing keys at every level, non-dict, `None`)
- [x] AST parse on all 3 modified files
- [x] Verified no frontend wire change — `viz_event_payload` keys (incl. `full_payload`) are identical, so `TaskProvider.tsx`, `taskVisualizerProcessor.ts`, `ChatSidePanel.tsx` are unaffected
- [ ] Soak in staging under a multi-tab high-tool-count task; confirm no eviction warnings on the same pod

### DIFF
--- a/src/solace_agent_mesh/gateway/http_sse/app.py
+++ b/src/solace_agent_mesh/gateway/http_sse/app.py
@@ -72,7 +72,7 @@ class WebUIBackendApp(BaseGatewayApp):
             "name": "sse_max_queue_size",
             "required": False,
             "type": "integer",
-            "default": 200,
+            "default": 1000,
             "description": "Maximum size of the SSE connection queues. Adjust based on expected load.",
         },
         {
--- a/src/solace_agent_mesh/gateway/http_sse/component.py
+++ b/src/solace_agent_mesh/gateway/http_sse/component.py
@@ -42,27 +42,27 @@ class _CompactionFutureEntry(TypedDict):
     expected_agent_id: str | None
 
 
-def _payload_excluded_from_visualization(payload: Any) -> bool:
-    """Mirror of routers.visualization._include_for_visualization but operating
-    on the raw payload dict before SSE serialization, so we can skip fan-out
-    entirely instead of building+dumping+parsing for every subscriber."""
+def _should_include_for_visualization(payload: Any) -> bool:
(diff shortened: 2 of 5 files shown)


efunneko (approved):
Looks good

PR #1465, opened 2026-04-27
fix(DATAGO-131321): expose useAllArtifacts

zhenyu369:

    Brief summary - what problem does this solve?


client/webui/frontend/src/lib/api/index.ts
```diff
@@ -1,4 +1,5 @@
 export { api, getErrorFromResponse, refreshToken, scheduleProactiveRefresh, cancelProactiveRefresh } from "./client";
+export { useAllArtifacts } from "./artifacts/hooks";
```
lgh-solace: Sorry, I thought this was already exported. Can you use the same style as the examples below?

`export * from "./artifacts";`

lgh-solace (approved):
If you have time, it seems like you could remove this override as I think it should be no longer necessary, but it's non-blocking. Thanks!

```
import { getArtifactContent as getArtifactContentUtil, ... } from "@/lib/utils/file";
```

PR #1468, opened 2026-04-28
fix(DATAGO-134083): Fix inline citations rendering format during response streaming

amir-ghasemi:
This pull request refactors how chat message content with citations is rendered, streamlining and centralizing citation handling in the UI. The main improvement is that citation bubbles are now rendered consistently during both streaming and static display, and the logic for when to show citations is moved into the `MarkdownWrapper` and `StreamingMarkdown` components.

**Rendering and citation handling improvements:**

* The `MarkdownWrapper` component now accepts `citations` and `onCitationClick` props, and automatically decides whether to render citation bubbles inline (using `TextWithCitations`) or fallback to plain markdown, both during and after streaming.
* The `StreamingMarkdown` component is updated to support citations, rendering citation bubbles inline during streaming if citations are provided.
* The chat message rendering logic in `ChatMessage.tsx` is simplified: instead of branching between `MarkdownWrapper`, `TextWithCitations`, and `MarkdownHTMLConverter`, it now always uses `MarkdownWrapper` and passes citation data as needed.
* The now-unnecessary direct import of `TextWithCitations` in `ChatMessage.tsx` is removed, since citation rendering is handled inside the markdown wrapper components.

### DIFF
--- a/client/webui/frontend/src/lib/components/chat/ChatMessage.tsx
+++ b/client/webui/frontend/src/lib/components/chat/ChatMessage.tsx
@@ -17,7 +17,6 @@ import { Sources } from "@/lib/components/web/Sources";
 import { ImageSearchGrid } from "@/lib/components/research";
 import { isDeepResearchReportFilename } from "@/lib/utils/deepResearchUtils";
 import { hasDocumentSearchResults } from "@/lib/utils/sourceUrlHelpers";
-import { TextWithCitations } from "./Citation";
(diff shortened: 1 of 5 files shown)


lgh-solace (approved):
It'd be nice if we could squeeze another test into this area?

amir-ghasemi:
> It'd be nice if we could squeeze another test into this area?

Done

PR #1469, opened 2026-04-28
fix(DATAGO-134087): Address laggy loading of recent-chats page

amir-ghasemi:
The /recent-chats page was loading sessions via an unbounded full-table scan: SessionService.get_user_sessions called find_background_tasks_by_status with no filters, pulling every background task in the system into Python before filtering by user. Cost grew with total task history, not page size.

Replace with a single SQL query joining ChatTaskModel and TaskModel, scoped to the user, the page's session_ids, execution_mode=background, end_time IS NULL, and non-terminal status. The terminal-status check on task_metadata JSON now runs only on the matching rows. Same fix benefits the Scheduled Tasks tab, which shares this code path.

Frontend: gate useInfiniteSessions and useSharedWithMe on activeTab so each tab pays only for its own data instead of always firing both.

### DIFF
--- a/client/webui/frontend/src/lib/api/sessions/hooks.ts
+++ b/client/webui/frontend/src/lib/api/sessions/hooks.ts
@@ -38,15 +38,17 @@ export function useRecentSessions(maxItems: number = MAX_RECENT_CHATS) {
  * Hook to fetch paginated sessions with infinite scroll support.
  * Automatically invalidates on session events (new session, session updated, title updated, background task completed).
  */
-export function useInfiniteSessions(pageSize: number = 20, source?: string) {
+export function useInfiniteSessions(pageSize: number = 20, source?: string, options?: { enabled?: boolean }) {
     const queryClient = useQueryClient();
+    const enabled = options?.enabled ?? true;
 
     useEffect(() => {
+        if (!enabled) return;
         const invalidate = () => queryClient.invalidateQueries({ queryKey: sessionKeys.lists() });
         const events = ["new-chat-session", "session-updated", "session-title-updated", "background-task-completed"];
         events.forEach(e => window.addEventListener(e, invalidate));
         return () => events.forEach(e => window.removeEventListener(e, invalidate));
-    }, [queryClient]);
+    }, [queryClient, enabled]);
 
     return useInfiniteQuery({
         queryKey: [...sessionKeys.lists(), "infinite", pageSize, source],
@@ -54,6 +56,7 @@ export function useInfiniteSessions(pageSize: number = 20, source?: string) {
         getNextPageParam: lastPage => lastPage.meta.pagination.nextPage ?? undefined,
         initialPageParam: 1,
         refetchOnMount: "always",
+        enabled,
     });
 }
 
--- a/client/webui/frontend/src/lib/api/share/hooks.ts
+++ b/client/webui/frontend/src/lib/api/share/hooks.ts
@@ -26,10 +26,11 @@ export function useShareUsers(shareId: string | undefined) {
     });
 }
 
-export function useSharedWithMe() {
+export function useSharedWithMe(options?: { enabled?: boolean }) {
     return useQuery({
         queryKey: shareKeys.sharedWithMe(),
         queryFn: shareService.listSharedWithMe,
+        enabled: options?.enabled ?? true,
     });
 }
 
--- a/client/webui/frontend/src/lib/components/pages/RecentChatsPage.tsx
+++ b/client/webui/frontend/src/lib/components/pages/RecentChatsPage.tsx
@@ -140,8 +140,8 @@ export const RecentChatsPage: React.FC = () => {
 
     // Sessions fetched only for chat/scheduler tabs; shared tab uses its own query.
     const sessionSource = activeTab === "shared" ? "chat" : activeTab;
-    const { data, fetchNextPage, hasNextPage, isFetchingNextPage } = useInfiniteSessions(PAGE_SIZE, sessionSource);
-    const { data: sharedChats = [], isLoading: isLoadingShared } = useSharedWithMe();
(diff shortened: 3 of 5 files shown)


efunneko:
## PR Review

### Summary
Replaces an unbounded full-table scan in `SessionService.get_user_sessions` (which loaded every background task into Python before filtering by user) with a single SQL query joined and scoped to the page's session IDs, the user, `execution_mode='background'`, and non-terminal status. Frontend gates `useInfiniteSessions` and `useSharedWithMe` on `activeTab` so only one fires per tab.

### Architecture
- **Significant gap: same anti-pattern still present in `search_sessions`** (`session_service.py:474-529`). It still calls `task_repo.find_background_tasks_by_status(db, status=None)` and filters in Python — the exact pre-fix code that was just removed from `get_user_sessions`. Search results pages will keep paying the original cost. The PR description calls out that the Scheduled Tasks tab \"shares this code path,\" but in practice search has its own copy that wasn't refactored. Fixing `search_sessions` to call the new `_find_sessions_with_running_background_tasks` helper is a one-line change and should be in this PR.
- The new helper (`session_service.py:45-97`) sits on `SessionService` but is purely a repository-shaped query against `ChatTaskModel` + `TaskModel`. It would fit more naturally on `TaskRepository`/`ChatTaskRepository` (e.g. `find_running_background_session_ids(user_id, session_ids)`), keeping SQL concerns out of the service layer. Not blocking.
- Frontend gating in `RecentChatsPage.tsx:143-144` and the `enabled` pattern in `hooks.ts:41-60` / `share/hooks.ts:29-34` is a clean, idiomatic TanStack Query approach — short-circuiting the event-listener effect when disabled is the right call.

### Code Quality
- Return type `dict[str, bool]` is awkward: the helper only ever stores `True`. A `set[str]` would convey intent more directly. Minor.
- Helper docstring at `session_service.py:54` reads \"chat-side completion can lead TaskModel.status.\" — missing a word.
- `if not session_id: continue` at `session_service.py:81` is dead defensive code — `ChatTaskModel.session_id` is `nullable=False`. Carryover from the original.
- Stray debug log in `session_service.py:151-152` inside the per-session loop is preserved from the old code; with the SQL pre-filter, every entry in `session_task_map` produces a debug line — no longer providing the \"I just chose to mark this\" signal it used to.
- PR title says \"Audit and optimize Database indexing for performance\" but no migrations or indexes are added. The change is pure query reshape. Either the JIRA title is out of sync with scope, or index work is missing from this PR.

### Security
- No new injection surface: parameters flow through SQLAlchemy parameter binding.
- User scoping is preserved at the SQL layer — this is actually a slight authorization-hardening improvement vs. the original Python-side filter.
- JSON parse remains wrapped in `try/except` and only logs a warning — no information disclosure of `raw_metadata` contents.
- No secrets, transport, or auth changes.

### Test Coverage
- **No backend tests** for `_find_sessions_with_running_background_tasks` or the `get_user_sessions` background-task-status behavior. The fix changes the result-equivalence claim (old: in-memory filter; new: SQL filter joined to TaskModel) — verifying parity (running task in non-terminal status surfaces; terminal-metadata task does not; cross-user task does not) under integration test would be appropriate.
- **No frontend test** asserting the `enabled` gating actually prevents the disabled query from firing. `RecentChatsPage.test.tsx:30-43` stubs both hooks unconditionally, so the new behavior is invisible to tests. A test that switches `activeTab` and asserts the wrong service function wasn't called would lock in the win this PR is making.
- Behavior change to `useInfiniteSessions` event listeners (skip add/remove when disabled) is also untested.

### Recommendations
1. **Blocking:** apply the same refactor to `search_sessions` — the unbounded scan is still there and undermines the stated goal of the PR.
2. **Should fix:** add a backend integration test that exercises `get_user_sessions` with running, terminal-by-metadata, and cross-user background tasks to lock in the new SQL-side filtering semantics.
3. **Should fix:** add a frontend test asserting the inactive tab's query is disabled (e.g. spy on `getPaginatedSessions` / `listSharedWithMe`).
4. **Nice-to-have:** return `set[str]` from the helper, fix the docstring typo, and consider moving the helper to `TaskRepository`/`ChatTaskRepository`.
5. **Nice-to-have:** reconcile PR title vs. actual scope (indexing vs. query reshape).

### Verdict
**Request Changes** — the core fix is correct and well-implemented, but the duplicate hot path in `search_sessions` was missed and there's no test coverage for either the backend SQL change or the frontend gating. Both are small follow-ups on top of an otherwise clean PR.

efunneko:
## Re-review on 6b11a78b3

Both blocking items resolved.

### Verified addressed
- `search_sessions` (`session_service.py:466-475`) now calls the same `_find_sessions_with_running_background_tasks` helper — duplicate scan gone.
- Backend coverage added in `tests/unit/gateway/http_sse/services/test_session_service_background_tasks.py` (9 cases): running surfaces, terminal-by-metadata excluded, foreground excluded, end_time excluded, cross-user no-leak, empty short-circuit, pending surfaces, null-status surfaces, page-bound subset. Locks in the SQL semantics nicely.
- Helper now returns `set[str]`, dead `if not session_id: continue` removed, stray per-session debug logs removed.

(Local pytest collection blocked by an unrelated `solace_ai_connector.common.observability` import error that hits all tests in this tree, not just the new ones — flagging only so it isn't mistaken for a test-quality issue.)

### Still open (non-blocking)
- Docstring at `session_service.py:54` still reads \"chat-side completion can lead TaskModel.status\" — there's a missing word (likely \"...can lead [the update of] TaskModel.status\" or \"...can outpace TaskModel.status\"). Worth one more pass.
- Frontend `enabled` gating (the other half of the perf win) still lacks a test that asserts the inactive tab's service function isn't called. Mentioning so it doesn't get lost — fine to defer.
- PR title (\"Audit and optimize Database indexing for performance\") still doesn't match the actual change (query reshape, no indexes added).

### Verdict
**Approve.** Core fix is correct, parity-tested, and now consistently applied across both code paths. The remaining nits are cosmetic.

efunneko (approved):
Looks good

JKaram (approved):

PR #1471, opened 2026-04-28
fix(DATAGO-134114): replace hardcoded max_llm_calls_per_task with constant and increase value to 30

cyrus2281:

    Replace hardcoded max LLM calls per task value with a constant to improve code consistency and maintainability.

    - Added a new constant DEFAULT_MAX_LLM_CALLS_PER_TASK in the common/constants.py file
    - Replaced all hardcoded instances of the value "20" with the new constant across multiple files:
      - fix: replace hardcoded max_llm_calls_per_task with constant DEFAULT_MAX_LLM_CALLS_PER_TASK

    Using a constant makes the code more maintainable by centralizing the configuration value in one place, making future changes easier and reducing the risk of inconsistency.

- [x] Manual testing: Verified agent behavior remains unchanged with the constant
- [ ] Unit tests: Not applicable - constant replacement only
- [ ] Integration tests: Not applicable
- [ ] Known limitations: None - direct value replacement with no functional changes

    Ensure all instances of the hardcoded value have been replaced correctly.


enavitan (approved):
looks good, wonder if we log the failures and at least try to understand where things went south, and perhaps have a follow up task to tweak prompts to reduce these failures case per case basis

PR #1473, opened 2026-04-28
feat(DATAGO-131493): User can run the experiment and view the result

josephabonasara:

Unblocks the enterprise eval pipeline's per-example **activity diagram** by exposing two reusable primitives from community SAM:

1. A pure utility for the live-SSE-stream → visualizer mapping (so the enterprise eval pipeline can persist events in the same shape chat streams).
2. A local-filesystem object-storage backend (so eval storage works in dev environments with no S3 infra).

Without these two additions, enterprise either has to fork/duplicate community mapping logic or require every developer to spin up MinIO before they can use eval features that rely on persisted artifacts.

**Public visualization mapper utility**
- New `src/solace_agent_mesh/gateway/http_sse/visualization_mapper.py` exporting the pure function `infer_visualization_event_details(topic, payload, log_identifier="")`. Body lifted verbatim from `WebUIBackendComponent._infer_visualization_event_details`; the only change is `self.log_identifier` becomes a parameter.
- `src/solace_agent_mesh/gateway/http_sse/component.py` — the existing method becomes a one-line delegating wrapper. Zero behavior change for chat.
- The enterprise eval pipeline calls this utility at upload time so the persisted event shape matches what chat streams live, eliminating the risk of a frontend mapping shim drifting from the backend mapper.

**Filesystem object-storage backend**
- New `src/solace_agent_mesh/services/platform/storage/filesystem_client.py` — `FileSystemStorageClient` implementing `ObjectStorageClient` against a local directory. One file per object (`{root}/{bucket}/{key}`). `get_public_url` / `generate_presigned_url` return `file://` URIs.
- `src/solace_agent_mesh/services/platform/storage/factory.py` — adds `"filesystem"` / `"fs"` / `"local"` aliases. Default is still `"s3"`; existing callers are unaffected.

### Key Design Decisions

- **Why extract the mapper as a pure function rather than persist `{topic, payload}` and rebuild on the frontend.** Keeping mapping logic in one place (Python) avoids drift between a backend implementation and a TS port. The persisted shape matches what chat already produces, so the visualizer code path is shared between live and replay.
- **Why a separate filesystem backend rather than an in-memory mock for dev.** The existing `EvalStorageService` round-trips bytes through the `ObjectStorageClient` interface; a filesystem implementation is the lowest-friction way to make the dev experience match production semantics without requiring MinIO. Real disk also makes failure-mode testing (size limits, permissions, prefix delete) realistic.
- **Why a separate test module for filesystem storage.** The existing `tests/unit/shared/storage/test_object_storage.py` imports `azure.storage.blob` at module top, so it requires the optional Azure SDK to even collect. Filesystem tests live in their own module so they run in any minimal env.

- [x] **Unit tests:** `tests/unit/gateway/http_sse/test_visualization_mapper.py` — 7 tests pinning the standalone function's contract (system events, topic-fallback for request/response/status, payload-summary truncation, serialization-failure handling).
- [x] **Unit tests:** `tests/unit/shared/storage/test_filesystem_storage.py` — 11 tests exercising the real disk path: round-trip put/get, factory aliases (`filesystem`/`fs`/`local`/`FILESYSTEM`), delete-prefix subtree behavior, missing-key `StorageNotFoundError`, `list_objects`, and `file://` public URLs.
- [x] **Manual testing:** ran an end-to-end eval experiment in enterprise with the filesystem backend (`~/.sam-data/object-storage/sam-eval-data/...`); confirmed the per-example activity diagram renders the same flow chart the chat panel produces for the same agent.
- [x] **Regression:** chat behavior is untouched. The component method now delegates to the extracted utility; existing tests for `WebUIBackendComponent` continue to pass.
- [ ] **Integration tests:** none added — both new pieces are pure-function / pure-IO units already covered comprehensively at the unit level.

- **`component.py` diff** is a behavior-preserving extraction. The 178-line removal is the original mapper body; the addition is a one-line `return infer_visualization_event_details(topic, payload, self.log_identifier)`. Worth a quick eyeball to confirm no logic was dropped.
- **`FileSystemStorageClient.delete_prefix`** has three code paths (target is dir / file / key-prefix string match). The third branch covers the case where the prefix doesn't correspond to a real directory but is a true key-substring — matches the behavior real S3 clients implement. All three are exercised by the new tests.
- **`generate_presigned_url`** silently ignores `expires_in`. Filesystem can't honor expiry; this is documented as a dev-only backend.
- **`put_object` `metadata`** is accepted and silently dropped (S3/Azure persist it). Acceptable for a dev backend; flagged here in case reviewers want a sidecar metadata file added before merge.

### DIFF
--- a/src/solace_agent_mesh/gateway/http_sse/component.py
+++ b/src/solace_agent_mesh/gateway/http_sse/component.py
@@ -26,6 +26,7 @@ from ...gateway.base.component import BaseGatewayComponent
 from ...gateway.http_sse.session_manager import SessionManager
 from ...gateway.http_sse.sse_manager import SSEManager
 from . import dependencies
+from .visualization_mapper import infer_visualization_event_details
 from .components import VisualizationForwarderComponent
 from .components.task_logger_forwarder import TaskLoggerForwarderComponent
 from .components.scheduler_result_forwarder import SchedulerResultForwarderComponent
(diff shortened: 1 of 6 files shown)


ziyanwan (approved):

bwiebe-solace (approved):
LGTM

PR #1474, opened 2026-04-29
fix(DATAGO-134139): stop cascading auth prompts in solace-chat's Atlassian heavy-query flow

amir-ghasemi:
## Summary

`EmbedResolvingMCPTool._is_auth_error` flagged any MCP tool result whose text contained both `"401"` and `"unauthorized"` (or `"403"` and `"forbidden"`) as an authentication failure. For free-form tool output — Jira issues, Confluence pages, search results — this triggers cascading "Pending user authorization." prompts mid-conversation whenever legitimate response content happens to include those substrings.

## Diagnosis

Diagnosed against the Atlassian MCP server on staging. A single `searchJiraIssuesUsingJql` response (~700KB, top-level keys `['issues','nextPageToken','isLast']`) matched the substring check four times:

| Substring | Where it actually appeared |
|---|---|
| `"401 Unauthorized"` | Inside a Jira issue description quoting an upstream SEMP API error log |
| `"403 Forbidden"` | Inside an issue description discussing S3 HeadBucket errors behind a corporate proxy |
| `"401"` | Jira's own status REST URL: `.../rest/api/3/status/10401` (the status ID literally is 10401) |
| `"403"` | Inside a Gravatar avatar URL hash that happened to contain `403` |

None of these were real auth failures. The result had:
- No top-level `isError`
- No top-level `code` / `status` of 401 / 403
- No nested 401 / 403 anywhere in the JSON tree (verified by walking the full structure)

The user saw two `"The 'JiraConfluenceAgent' agent requires authentication."` prompts during one heavy DATAGO bug query, with no underlying auth issue.

## Fix

Tighten `_is_auth_error`:

1. **Always trust structured signals** — text items that parse as JSON with top-level `code` or `status` equal to 401 / 403. Behavior unchanged.
2. **Gate the substring fallback** behind two conditions:
   - The MCP `CallToolResult.isError` flag is true. We only consider substring matches when we're already inside an error response, not a legitimate large data response.
   - The text item is at most 2 KB (`_SUBSTRING_AUTH_CHECK_MAX_TEXT_LEN`). Real auth error payloads are short; substring matches in larger payloads are dominated by incidental content.

A separate enterprise-side fix (PR #1080 in `solace-agent-mesh-enterprise`) made `save_credential(None)` a no-op so the false-positive trigger no longer destroys the persisted credential — but the user-visible re-auth prompts still fire from this path until this detector is fixed.

## Test plan

- [x] Existing tests updated where they implicitly relied on substring matching working without `isError` (now set `isError=True` to match real MCP shape).
- [x] New regression test reproducing the staging case: a large Jira-style response containing the offending substrings inside legitimate content. Asserts `_is_auth_error` returns `False`.
- [x] New test: large payload with `isError=True` and an embedded substring match — still `False` because it's beyond the size gate.
- [x] New test: structured `{"code": 401}` is detected even in a large payload (the structured signal is not gated).
- [x] New test: `isError=True` with non-auth-related text doesn't trigger re-auth.
- [x] `tests/unit/agent/adk/test_mcp_auth_error_detection.py` — 22 tests pass.

### DIFF
--- a/src/solace_agent_mesh/agent/adk/embed_resolving_mcp_toolset.py
+++ b/src/solace_agent_mesh/agent/adk/embed_resolving_mcp_toolset.py
@@ -337,10 +337,33 @@ class EmbedResolvingMCPTool(_BaseMcpToolClass):
         return data
 
 
+    # Substring matching is only trusted for response payloads that are this
+    # small. Free-form tool output (Jira issues, Confluence pages, search
+    # results) routinely embeds substrings like "401 Unauthorized" or
+    # "403 Forbidden" in legitimate content (logs, error reports, status URLs
+    # like .../rest/api/3/status/10401, gravatar hashes that contain "403").
+    # Triggering re-auth on those was the cause of cascading "Pending user
+    # authorization" prompts mid-conversation.
+    _SUBSTRING_AUTH_CHECK_MAX_TEXT_LEN = 2000
+
     def _is_auth_error(self, result: dict) -> bool:
-        """Check if an MCP tool result indicates an authentication/authorization error."""
+        """Check if an MCP tool result indicates an authentication/authorization error.
+
+        Authoritative signals (always trusted):
+        - The MCP CallToolResult-level ``isError`` flag.
+        - A text item that parses as a JSON object with top-level
+          ``code`` or ``status`` equal to 401 / 403.
+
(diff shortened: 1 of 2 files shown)


efunneko (approved):
LG

PR #1476, opened 2026-04-29
fix(DATAGO-134263): honour per-node timeout in workflow agent calls

efunneko:

The per-node `timeout` field on workflow `AgentNode` and `WorkflowInvokeNode` was declared in the Pydantic schema but **never consulted at runtime** — the agent caller always used the workflow-level `default_node_timeout_seconds`, silently discarding per-node overrides like `timeout: "5m"`.

Reported on the CBRE POC (SAM Ent v1.143.0) — a `classify_errors` node configured with `timeout: "5m"` was timing out after 60s because the workflow's `default_node_timeout_seconds: 60` was winning.

Jira: [DATAGO-134263](https://sol-jira.atlassian.net/browse/DATAGO-134263)

- `agent_caller.py`: new `_resolve_node_timeout()` helper parses the node's `timeout` string via `parse_duration` and falls back to `default_node_timeout_seconds`. Both `call_agent` and `call_workflow` now thread the resolved timeout into `_publish_agent_request`, which uses it for the cache-expiry that drives node-level timeouts. Invalid per-node duration strings log a warning and fall back to the default rather than failing the call.
- `workflow_execution_context.py`: added a `sub_task_timeouts` map with thread-safe `set_sub_task_timeout` / `get_sub_task_timeout` so the timeout handler can recover the actual value that fired.
- `component.py`: timeout error message now reports the resolved timeout (read from context), not the workflow default.

### Key Design Decisions

- Invalid per-node timeout strings fall back to the default with a warning — a malformed `timeout` shouldn't take down a workflow.
- Tracking is keyed by `sub_task_id` and stored on `WorkflowExecutionContext` (which already owns sub-task lifetime), so cleanup happens automatically when the context dies.

- [x] Unit tests added in `tests/unit/workflow/test_agent_caller.py`:
  - `TestResolveNodeTimeout` — per-node timeout overrides default; missing/empty falls back; invalid string falls back.
  - `TestPublishAgentRequestTimeout` — cache expiry uses the passed timeout; context tracks it for later error reporting.
- [x] Manual testing in CBRE workflow environment (recommended)

Local `.venv` couldn't run the workflow unit suite (pre-existing — the installed `solace-ai-connector==3.3.4` is missing `common.observability`, which blocks collection of `tests/unit/workflow/*` on `main` too). CI should run them cleanly.

- The fallback on invalid duration strings — reviewers may prefer hard-failing instead. I went with warn-and-fallback because the schema doesn't validate the format and a typo shouldn't take down a running workflow.
- `_publish_agent_request` is now invoked from two call sites (`call_agent`, `call_workflow`) — confirm both pass the resolved timeout correctly.

[DATAGO-134263]: https://sol-jira.atlassian.net/browse/DATAGO-134263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/solace_agent_mesh/workflow/agent_caller.py
+++ b/src/solace_agent_mesh/workflow/agent_caller.py
@@ -20,6 +20,7 @@ from ..agent.utils.artifact_helpers import (
     format_artifact_uri,
 )
 from .app import WorkflowNode, WorkflowInvokeNode
+from .utils import parse_duration
 from .workflow_execution_context import WorkflowExecutionContext, WorkflowExecutionState
 
 if TYPE_CHECKING:
@@ -138,7 +139,11 @@ class AgentCaller:
 
         # Publish request
         await self._publish_agent_request(
-            node.agent_name, message, sub_task_id, workflow_context
+            node.agent_name,
+            message,
+            sub_task_id,
+            workflow_context,
+            self._resolve_node_timeout(node.timeout),
         )
 
         # Track in workflow context
@@ -227,7 +232,11 @@ class AgentCaller:
 
         # Publish request to the sub-workflow
         await self._publish_agent_request(
-            node.workflow_name, message, sub_task_id, workflow_context
+            node.workflow_name,
+            message,
+            sub_task_id,
+            workflow_context,
+            self._resolve_node_timeout(node.timeout),
         )
 
         # Track in workflow context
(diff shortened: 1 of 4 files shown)


solace-pjones (approved):

solace-pjones:
Did a manual test of cbre use case, verified the bug, applied Ed's fix and verified the timeout no longer occurs. We are now honouring the agent timeout.

PR #1484, opened 2026-04-30
fix(DATAGO-122712): Enhance mention detection and add tests for multi-word queries

cyrus2281:

The mention detection logic previously only supported single-word queries, meaning users couldn't search for people by their full name (e.g. "Michael Du Plessis"). This change enables multi-word mention queries so users can type a first and last name after the `@` trigger and still get relevant search results.

- fix: enhance mention detection and add tests for multi-word queries
  - Updated `detectMentionTrigger` in `mentionUtils.ts` to allow spaces within a mention query, replacing the previous rule that closed the dropdown on any space. The dropdown now only closes on double-spaces, leading whitespace, newlines, or queries exceeding 50 characters.
  - Non-breaking spaces (commonly inserted by `contenteditable` elements) are now normalized to regular spaces before detection and search, ensuring consistent behaviour.
  - The search query is trimmed before being sent to the API in `MentionsCommand.tsx` to avoid trailing-space issues in requests.

Double-space was chosen as the signal to close the mention dropdown rather than a single space, as a single space is now needed to support multi-word name searches. A maximum query length of 50 characters is enforced to prevent the dropdown from staying open across unintended long sentences.

- [ ] Manual testing: Typing a multi-word query after `@` in the chat input to verify search results appear and the correct URL-encoded query is sent to the backend.
- [ ] Unit tests: New test file `mentionUtils.test.ts` added with comprehensive coverage of `detectMentionTrigger`, including edge cases for multi-word queries, non-breaking spaces, double-spaces, newlines, leading whitespace, length limits, and cursor position scoping. Existing `MentionsCommand.test.tsx` extended with a multi-word query test verifying URL encoding.
- [ ] Known limitations: A `console.log` debug statement was left in `mentionUtils.ts` and should be removed before merge.

A `console.log` debug statement in `mentionUtils.ts` needs to be removed. Reviewers should also verify the double-space-to-close behaviour feels natural in practice, as it is a behaviour change from the previous single-space rule.

### DIFF
--- a/client/webui/frontend/src/lib/components/chat/MentionsCommand.tsx
+++ b/client/webui/frontend/src/lib/components/chat/MentionsCommand.tsx
@@ -107,7 +107,7 @@ export const MentionsCommand: React.FC<MentionsCommandProps> = ({ isOpen, onClos
         const fetchPeople = async () => {
             setIsLoading(true);
             try {
-                const data: PeopleSearchResponse = await api.webui.get(`/api/v1/people/search?q=${encodeURIComponent(searchQuery)}&limit=10`);
+                const data: PeopleSearchResponse = await api.webui.get(`/api/v1/people/search?q=${encodeURIComponent(searchQuery.trim())}&limit=10`);
                 setPeople(data.data || []);
             } catch (error) {
                 console.error("Failed to fetch people:", error);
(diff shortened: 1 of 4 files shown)


amir-ghasemi (approved):
Please remove debug console.log  `mentionUtils.ts:50`     
   
```
  console.log("text", text, "cursorPosition", cursorPosition, "textBeforeCursor",
   textBeforeCursor);   
```

PR #1489, opened 2026-04-30
feat(DATAGO-134503): Experiment runs shows agent generated artifacts

josephabonasara:

Lets library consumers reuse the chat artifact preview UI outside of a chat session — specifically, the SAM Enterprise eval-experiment results page, which renders saved artifacts but has no `ChatProvider` ancestor. Today, importing `ArtifactDetails` and `ContentRenderer` from `@SolaceLabs/solace-agent-mesh-ui` either isn't possible (not exported) or throws at runtime (`useChatContext` is required).

- **`ArtifactDetails.tsx`** — switched from `useChatContext()` (throws when no provider) to `useContext(ChatContext)` (returns `undefined` when absent). The version-selector now hides instead of throwing when there's no chat session, and a `&& navigateArtifactVersion` render gate keeps TS strict mode happy. **Behavior inside a `ChatProvider` is unchanged** — same context values, same render output.
- **`chat/index.ts`** — explicit re-exports of `ArtifactBar`, `ArtifactDetails`, `ContentRenderer`, and the preview helpers (`getRenderType`, `canPreviewArtifact`, `decodeBase64Content`, `encodeBase64Content`, `getFileContent`). Sibling artifact components that still depend on `useChatContext` are intentionally not re-exported (called out in a comment).
- **`ui/index.ts`** — re-exports the `Sheet` primitive family (`Sheet`, `SheetClose`, `SheetContent`, …) so consumers can build slide-out side-panels.

Considered moving `ArtifactDetails` to a context-agnostic location with an explicit `versions` prop, but that would break the existing in-chat call site (`ArtifactPanel.tsx`) and force a prop-drilling refactor. Optional context consumption is a smaller, safer change with no behavior delta for current consumers.

Considered adopting `FileMessage`'s `readOnly` prop + `hasContext` guard pattern. Skipped because `ArtifactDetails` doesn't render differently for read-only state — the simpler optional-context check is a better fit here.

- [x] Manual testing: type-checked enterprise frontend (`tsc --noEmit -p tsconfig.app.json`) against a freshly-rebuilt local tgz of this branch — clean.
- [x] Unit tests: new `src/stories/chat/ArtifactDetails.test.tsx` (7 tests) covering both with-`ChatProvider` and without-`ChatProvider` render paths, version-selector visibility, and action-button callbacks.
- [x] Full unit suite: `npx vitest run --project=unit` — 87 files / 921 tests pass.
- [x] Lint: `npx eslint` clean on changed files.
- [x] Type-check: `npx tsc --noEmit -p tsconfig.app.json` clean.
- [ ] Storybook test: not added — no existing storybook story for `ArtifactPanel` to attach to, and `ArtifactDetails` already has full unit coverage. Worth adding a play-function story in a follow-up.

- **Bundle size**: `ContentRenderer` statically imports all 10 renderers (PDF, Mermaid, Office, Markdown, etc.). Tree-shaking should drop unused ones for consumers using named imports — verified no `import * as Chat` exists in the enterprise repo. If your downstream uses wildcard imports, the chat barrel is now wider.
- **Runtime peer-deps**: `ContentRenderer`'s renderers carry implicit peer-dep requirements (PDF.js worker, Mermaid runtime, etc.). The current consumer (enterprise eval results) already wires these up; broader consumers may need to.

### DIFF
--- a/client/webui/frontend/src/lib/components/chat/artifact/ArtifactDetails.tsx
+++ b/client/webui/frontend/src/lib/components/chat/artifact/ArtifactDetails.tsx
@@ -1,9 +1,9 @@
-import React, { useMemo } from "react";
+import React, { useContext, useMemo } from "react";
 
 import { Download, Info, Trash } from "lucide-react";
 
 import { Button, Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from "@/lib/components/ui";
-import { useChatContext } from "@/lib/hooks";
+import { ChatContext } from "@/lib/contexts/ChatContext";
 import { formatRelativeTime } from "@/lib/utils/format";
 import type { ArtifactInfo } from "@/lib/types";
 
@@ -23,8 +23,13 @@ interface ArtifactDetailsProps {
 }
 
 export const ArtifactDetails: React.FC<ArtifactDetailsProps> = ({ artifactInfo, isPreview = false, isExpanded = false, onDelete, onDownload, setIsExpanded, badge }) => {
-    const { previewedArtifactAvailableVersions, currentPreviewedVersionNumber, navigateArtifactVersion } = useChatContext();
-    const versions = useMemo(() => previewedArtifactAvailableVersions ?? [], [previewedArtifactAvailableVersions]);
+    // Use the chat context optionally — when ArtifactDetails is rendered outside a
+    // ChatProvider (e.g. eval experiment results) the version-selector simply
+    // hides itself instead of throwing.
+    const ctx = useContext(ChatContext);
+    const versions = useMemo(() => ctx?.previewedArtifactAvailableVersions ?? [], [ctx?.previewedArtifactAvailableVersions]);
+    const currentPreviewedVersionNumber = ctx?.currentPreviewedVersionNumber;
+    const navigateArtifactVersion = ctx?.navigateArtifactVersion;
 
     return (
         <div className="flex flex-row justify-between gap-1">
@@ -41,7 +46,7 @@ export const ArtifactDetails: React.FC<ArtifactDetailsProps> = ({ artifactInfo, 
                     </div>
                 </div>
 
-                {isPreview && versions.length > 1 && (
+                {isPreview && versions.length > 1 && navigateArtifactVersion && (
                     <div className="align-right">
                         <Select
                             value={currentPreviewedVersionNumber?.toString()}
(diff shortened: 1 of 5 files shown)


client/webui/frontend/src/lib/components/chat/artifact/ArtifactDetails.tsx
```diff
@@ -23,8 +23,13 @@ interface ArtifactDetailsProps {
 }
 
 export const ArtifactDetails: React.FC<ArtifactDetailsProps> = ({ artifactInfo, isPreview = false, isExpanded = false, onDelete, onDownload, setIsExpanded, badge }) => {
-    const { previewedArtifactAvailableVersions, currentPreviewedVersionNumber, navigateArtifactVersion } = useChatContext();
-    const versions = useMemo(() => previewedArtifactAvailableVersions ?? [], [previewedArtifactAvailableVersions]);
+    // Use the chat context optionally — when ArtifactDetails is rendered outside a
+    // ChatProvider (e.g. eval experiment results) the version-selector simply
+    // hides itself instead of throwing.
+    const ctx = useContext(ChatContext);
```
bwiebe-solace, line 29: `FileMessage.tsx` already does optional `ChatContext` consumption (`useContext(ChatContext) as ChatContextValue | undefined` + a `hasContext` gate). With this PR there are two call sites doing the same thing in slightly different shapes. Not a blocker, but worth a small follow-up to unify behind a `useOptionalChatContext()` hook so the "this component opts out of context" intent is explicit and consistent across components.

---
*[Added by Claude Code]*

client/webui/frontend/src/stories/chat/ArtifactDetails.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+describe("ArtifactDetails outside a ChatProvider", () => {
+    test("renders filename without throwing when no ChatContext is present", () => {
+        render(<ArtifactDetails artifactInfo={baseArtifact} isPreview />);
+        expect(screen.getByText("report.pdf")).toBeInTheDocument();
+    });
+
+    test("hides version selector when no ChatContext is present even with isPreview=true", () => {
+        render(<ArtifactDetails artifactInfo={baseArtifact} isPreview />);
+        expect(screen.queryByText("Version")).not.toBeInTheDocument();
```
bwiebe-solace, line 33: Inconsistent selector — the with-`ChatProvider` tests in this file use `queryByRole("combobox")` for the equivalent assertion. Recommend matching that here for robustness against placeholder text changes.

---
*[Added by Claude Code]*

bwiebe-solace (approved):
A couple of non-blocking nits, feel free to merge as-is in the interest of time

PR #1491, opened 2026-04-30
fix(DATAGO-134599): Pre-filter empty sessions on /api/v1/artifacts/all

amir-ghasemi:
## Summary
`list_all_artifacts` does one S3 `list_artifact_keys` call per user session, even though for power users most sessions have **no artifacts in storage at all** (one-message chats, abandoned sessions, test runs). The endpoint clips the 60s upstream timeout for users with many sessions.

This PR adds a single-call user-prefix S3 listing and uses it to filter the session list **before** the per-session scan. Project-scoped sources pass through unchanged, and search queries bypass the filter (they need to scan everything).

The per-session loop runs `list_artifact_keys` (one S3 ListObjectsV2 each, gated to 10 concurrent), then loads metadata.json per artifact. Most sessions are empty so most calls return zero — pure waste.

S3 keys are structured `{app_name}/{user_id}/{session_id}/{filename}/{version}`, so a single ListObjectsV2 with prefix `{app_name}/{user_id}/` enumerates every session the user has artifacts in. 

## Compatibility
- `S3ArtifactService` gets the new method.
- `BaseArtifactService` is unchanged (it's the upstream google.adk class).
- Other backends (`AzureArtifactService`, `FilesystemArtifactService`) — unchanged. The router uses `getattr(..., None)` and falls back to the existing per-session iteration when the method is absent, so non-S3 deployments are unaffected.
- Search queries (`?search=...`) skip the optimization to preserve the cross-session search semantics that already exist.

## Test plan
- [x] `pytest tests/unit/gateway/http_sse/routers/test_artifacts.py tests/unit/agent/utils/test_artifact_helpers.py tests/unit/agent/adk/` — **828 passed, 1 skipped**
- [x] Existing `TestListAllArtifacts` suite (13 tests) still passes — mocks use `spec=BaseArtifactService` so they exercise the fallback path.
- [x] New regression test `test_list_all_artifacts_prefilters_empty_sessions` stubs the new method and asserts that only the artifact-bearing session triggers `get_artifact_info_list_fast`.
- [ ] Soak in staging post-deploy: amir's `/api/v1/artifacts/all` should drop from clipping the 60s timeout to <5s.
- [ ] Soak in prod for power users (`zhenyu.chen` 561 sessions, `david.marcuccio` 315, `prasanth.pulipakala` 260): same speedup expected.

## Out of scope (follow-ups)
- Fully O(1) S3-prefix-driven listing that also derives `mime_type`/`size` from list output instead of per-artifact `metadata.json` GETs. This PR keeps the per-artifact metadata fetch — it's only run for sessions that actually have artifacts now, so the cost is bounded by artifact count, not session count.
- Maintaining a `sessions.has_artifacts` boolean column for non-S3 backends. Same effect via DB instead of S3 prefix list, useful if more deployments use Azure/filesystem.


efunneko (commented):
## PR Review

### Summary
Adds a single-call user-prefix S3 listing (`list_sessions_with_artifacts_for_user`) on `S3ArtifactService` and uses it in `routers/artifacts.py` to filter session rows before the per-session `list_artifact_keys` scan. Other backends and search queries fall through to the existing path via `getattr(..., None)`. Well-targeted optimization.

### Architecture
- Good fit. New method lives on the concrete `S3ArtifactService`, keeping the upstream Google ADK base untouched.
- `getattr(..., None)` fallback in the router is the right call — non-S3 backends are unaffected with no spec changes leaking across implementations.
- Search bypass and project carve-out (`startswith(\"project-\")`) are correct.

### Code Quality
- Method naming, log prefixes, and the `asyncio.to_thread` paginator pattern mirror the adjacent `list_artifact_keys` — consistent.
- Error path uses `ClientError` + warning + empty-set fallback, matching the existing pattern.
- **Nit:** the comment block in `artifacts.py` (Step 1.5) leans long and bakes in user-specific anecdotes (\"amir had 6,752 sessions...\"). Concrete prod numbers age poorly — consider trimming to the *why* and leaving the metrics in this PR description.

### Security
- No new authn/authz surface.
- Prefix is built from config-controlled `app_name` and an already-authenticated `user_id`; `app_name.strip('/')` matches the existing method. No injection risk.
- No secrets logged.

### Test Coverage
- New regression test correctly stubs the method and asserts that only the artifact-bearing session triggers `get_artifact_info_list_fast` — good behavioural assertion (checks `call_args_list`, not just outcome).
- Fallback path is implicitly covered via existing `MagicMock(spec=BaseArtifactService)` fixtures. Worth one explicit test asserting \"no method on service → all sessions still scanned\" to lock the contract.
- **Missing:** failure-path test — when `list_sessions_with_artifacts_for_user` raises, the router should fall through to the per-session scan (lines 987-992). No test asserts this.
- **Missing:** unit tests for `list_sessions_with_artifacts_for_user` itself — S3 key parsing, the `ClientError` → empty-set path, and behaviour when no objects exist.

### Behavioural Concern Worth a Decision
**User-scoped artifacts edge case.** S3 keys for user-scoped artifacts are `{app}/{user_id}/user/{filename}/{version}` (literal `user` segment — see `list_artifact_keys` around line 343). The new method's `parts[2]` will pick up the literal string `\"user\"` and add it to the returned set.

- If the user has any session-scoped artifact, at least one real `session_id` survives the filter, that session's `list_artifact_keys` call still returns the user-scoped artifacts (with `user:` prefix), and they surface as before. No regression.
- **However**, if a user has *only* user-scoped artifacts (no session-scoped) plus DB sessions, the prefix listing returns `{\"user\"}` (truthy), the filter strips every real `session_id`, and the user-scoped artifacts disappear from `/api/v1/artifacts/all`.

Pre-PR behaviour: every session's `list_artifact_keys` returned user-scoped artifacts (duplicated per session, but visible). Whether `/all` is *meant* to surface user-scoped artifacts is unclear from the code — they aren't deduplicated, so prior behaviour was arguably buggy. But this PR silently changes it. Worth either an explicit \"skip filter if the only entry is 'user'\" guard, or a comment documenting the intent.

### Recommendations
1. (Discussion) Decide what to do about user-scoped-only artifacts — keep prior behaviour or document the exclusion.
2. (Suggestion) Add a unit test for `list_sessions_with_artifacts_for_user` (key parsing, ClientError → empty set).
3. (Suggestion) Add a router test for the fallback when the new method raises.
4. (Nit) Trim the prod-anecdote comment in `artifacts.py`.

### Verdict
**Approve with minor follow-ups.** Optimization is well-targeted, fallback design is defensive, regression test pins the new behaviour. The user-scoped-only edge case is the one real concern — likely a corner that doesn't bite in practice, but worth a confirmed decision before merge.

efunneko (approved):
Approving — see prior review comment for details and minor follow-up suggestions.

PR #1496, opened 2026-05-04
fix(agent): lazy-init DynamicModelProvider for per-request model_override

josephabonasara:

Per-request `model_override` (used by enterprise eval to swap an agent's
model on a per-task basis) was being rejected with `"model config not
available"` for any agent whose YAML doesn't declare a `model_provider`
field — including the bundled OrchestratorAgent in `custom_llm_agent_app`.
That broke eval runs against those agents end-to-end: every example
failed at the entry agent before the LLM was even called.

Concretely:
- `SamComponentBase.__init__` reads `model_provider` from the YAML; if
  absent, `self.model_provider = None`.
- `_async_setup_and_run` only starts the model listener when both
  `_lazy_model_mode` AND `model_provider` are truthy, so the orchestrator
  never starts one — `_dynamic_model_provider` stays `None`.
- `_resolve_model_override_metadata` then takes the `if not
  dynamic_model_provider` branch and returns `"model config not
  available"`, rejecting the task.

Forcing every existing and future agent's YAML/helm template to declare
`model_provider` is brittle ongoing maintenance, especially for built-in
mesh agents.

The override path now lazy-instantiates an override-only
`DynamicModelProvider` on the first per-request override, instead of
relying on a startup-time provider tied to a specific configured alias.

**`agent/adk/models/dynamic_model_provider.py`**
- New `skip_bootstrap` constructor flag. When `True`, `initialize()`
  wires up the broker subscription via `listen_for_model_config_change`
  and returns immediately — it does **not** call `request_model_config`.
- Skipping the bootstrap is what makes this safe: the wildcard
  subscription `configuration/model/response/*/{component_id}` already
  catches one-shot resolves for any alias/UUID via
  `complete_pending_resolve`, but never receives a topic that matches
  `self.model_provider.model_id`, so `update_litellm_model` /
  `remove_litellm_model` (the auto-update path) can never fire and
  clobber the agent's static `model:` config.

**`agent/protocol/event_handlers.py`**
- New `_ensure_override_provider(component)` helper: under an
  `asyncio.Lock` on the component, constructs a `DynamicModelProvider`
  with `skip_bootstrap=True` keyed on a sentinel `model_id`
  (`"__lazy_override_resolver__"`) that no real platform alias can match,
  awaits its `initialize()` so the subscription is live before returning,
  and stores it on `component._dynamic_model_provider`.
- `_resolve_model_override_metadata` now takes `component` instead of
  the bare provider; if no provider exists when an override comes in, it
  calls `_ensure_override_provider` and resolves through that. If the
  component has no litellm instance to back resolution, it falls back
  to the existing `"model config not available"` rejection rather than
  constructing a dead provider.

**`common/sac/sam_component_base.py`**
- Adds `_dynamic_model_provider_init_lock: Optional[asyncio.Lock] =
  None` so concurrent first-override requests don't race two providers
  into existence.

### Key Design Decisions

**Why a sentinel `model_id` instead of the override's own id?** The
receiver routes by `topic_model_id == self.model_provider.model_id` —
match → auto-update litellm; mismatch → complete a one-shot resolve
future. If we keyed the provider on the override's id, the platform's
bootstrap response (or any subsequent config-update broadcast for that
alias) would mutate the agent's static litellm permanently. A sentinel
guarantees the auto-update branch is unreachable while leaving one-shot
resolves fully functional.

**Why lazy-init rather than always-on at startup?** Always-on requires
either (a) every agent template declares `model_provider` — fragile and
easy to forget — or (b) starting a provider with no real alias for every
agent at boot, paying broker subscription cost even when overrides are
never used. Lazy-init pays only on demand and self-heals for new agents.

**Why `asyncio.Lock` rather than a thread lock?** The override path is
fully async; a thread lock would force `await` paths off the event loop
inside the critical section. `asyncio.Lock` keeps the await semantics
intact while still serialising init.

- [x] Unit tests:
  - `tests/unit/agent/protocol/test_event_handlers_model_override.py` —
    existing 13 tests reworked to pass a mock component; 2 new tests
    cover `TestLazyOverrideProviderInit::test_lazy_init_when_provider_missing`
    (verifies `skip_bootstrap=True` and provider stored on component) and
    `test_lazy_init_skipped_when_litellm_unavailable` (rejection when
    component has no litellm).
  - `tests/unit/agent/adk/models/test_dynamic_model_provider.py` —
    fixture initializes `_skip_bootstrap=False`; new
    `test_initialize_skip_bootstrap_subscribes_only` confirms the listener
    is set up but `request_model_config` is never called.
- [x] Full unit suite: 5206 passed, 14 skipped, 6 xfailed (pre-existing).
- [ ] Manual end-to-end against test cluster — pending merge + community
  release + enterprise `manifest.json` bump.

- The lazy-init runs on the request thread the first time an override
  arrives. The SAC flow setup (queue creation + subscription) typically
  takes ~200 ms; subsequent overrides reuse the same provider. Worth a
  look from anyone who knows the SAC flow startup cost on the production
  brokers — if that latency is problematic, an alternative is to lazy-init
  earlier (e.g. on first chat request), but that wasn't necessary for the
  bug at hand.
- The enterprise eval broker currently sends the model **UUID** rather
  than its alias (`eval_run_service.py:312` prefers `modelConfigId`).
  Platform's `ModelConfigurationRepository.get_by_alias_or_id` accepts
  both, so this works — but it's a latent coupling worth noting if either
  side ever tightens its input handling.
- Sentinel string `"__lazy_override_resolver__"` is unlikely to collide
  with a real model alias, but if you'd prefer a UUID-shaped sentinel for
  belt-and-braces I can switch it.

### DIFF
--- a/src/solace_agent_mesh/agent/protocol/event_handlers.py
+++ b/src/solace_agent_mesh/agent/protocol/event_handlers.py
@@ -1084,9 +1084,63 @@ def _send_error_response_and_nack(
     component.handle_error(exception, Event(EventType.MESSAGE, message))
 
 
+_LAZY_OVERRIDE_PROVIDER_SENTINEL = "__lazy_override_resolver__"
+
+
+async def _ensure_override_provider(component: "SamAgentComponent") -> Any:
+    """Lazy-init an override-only ``DynamicModelProvider`` on the component.
+
+    Agents whose YAML lacks a ``model_provider`` field don't start a model
+    listener at boot, so ``component._dynamic_model_provider`` is ``None`` and
+    per-request ``model_override`` resolution would fail with "model config
+    not available". Instead of forcing every agent's YAML to declare
+    ``model_provider``, we spin up an override-only provider on the first
+    override request: it shares the wildcard bootstrap subscription so
+    ``resolve()`` works for any alias/UUID, but skips the auto-update path
+    (``skip_bootstrap=True``) so the agent's static ``model:`` config is
+    never replaced.
+
+    Returns the provider, or ``None`` if the component has no litellm
+    instance to back the resolution flow.
+    """
+    if component._dynamic_model_provider is not None:
+        return component._dynamic_model_provider
+
+    if component._dynamic_model_provider_init_lock is None:
+        component._dynamic_model_provider_init_lock = asyncio.Lock()
+
+    async with component._dynamic_model_provider_init_lock:
+        if component._dynamic_model_provider is not None:
+            return component._dynamic_model_provider
+
+        litellm_instance = component.get_lite_llm_model()
+        if litellm_instance is None:
+            return None
+
+        # Inline import keeps event_handlers free of an agent->adk dependency
+        # at module load and matches the pattern used elsewhere in this file.
+        from ..adk.models.dynamic_model_provider import DynamicModelProvider
+
+        provider = DynamicModelProvider(
+            component,
+            litellm_instance,
+            _LAZY_OVERRIDE_PROVIDER_SENTINEL,
+            skip_bootstrap=True,
+        )
+        # Wait for the broker subscription to be live before returning so the
+        # immediately-following resolve() doesn't race the listener setup.
+        await provider.initialize()
+        component._dynamic_model_provider = provider
+        log.info(
+            "%s Lazy-initialised override-only DynamicModelProvider",
+            component.log_identifier,
+        )
+        return provider
+
+
 async def _resolve_model_override_metadata(
     task_metadata: Dict[str, Any],
-    dynamic_model_provider: Any,
+    component: "SamAgentComponent",
     log_identifier: str,
 ) -> Optional[str]:
     """Resolve model_override alias in task metadata to a raw LiteLLM config dict.
@@ -1126,6 +1180,10 @@ async def _resolve_model_override_metadata(
         return None
 
     model_id = model_override["model_id"]
+    dynamic_model_provider = component._dynamic_model_provider
+    if dynamic_model_provider is None:
+        dynamic_model_provider = await _ensure_override_provider(component)
+
     resolved = None
     if dynamic_model_provider:
         resolved = await dynamic_model_provider.resolve(model_id)
@@ -1231,7 +1289,7 @@ async def _handle_send_message_request(
 
     override_error = await _resolve_model_override_metadata(
         task_metadata,
-        component._dynamic_model_provider,
+        component,
         component.log_identifier,
     )
     if override_error:
--- a/src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py
+++ b/src/solace_agent_mesh/agent/adk/models/dynamic_model_provider.py
@@ -143,10 +143,20 @@ class ModelConfigReceiverComponent(ComponentBase):
 
 class DynamicModelProvider:
 
-    def __init__(self, component: ComponentBase, litellm_instance: LiteLlm, model_id: str):
+    def __init__(
+        self,
+        component: ComponentBase,
+        litellm_instance: LiteLlm,
+        model_id: str,
+        skip_bootstrap: bool = False,
+    ):
         self._component = component
         self._litellm_instance = litellm_instance
         self._model_id = model_id
+        # When True, set up the broker subscription only — don't send a bootstrap
+        # request for ``model_id``. Used by the override-only path so the agent's
+        # static litellm model isn't replaced via the auto-update flow.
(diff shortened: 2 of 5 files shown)


src/solace_agent_mesh/agent/protocol/event_handlers.py
```diff
@@ -1084,9 +1084,63 @@ def _send_error_response_and_nack(
...
+
+        provider = DynamicModelProvider(
+            component,
+            litellm_instance,
+            _LAZY_OVERRIDE_PROVIDER_SENTINEL,
+            skip_bootstrap=True,
+        )
+        # Wait for the broker subscription to be live before returning so the
+        # immediately-following resolve() doesn't race the listener setup.
+        await provider.initialize()
```
bwiebe-solace, line 1132: 🟡 **Important: `initialize()` runs twice on this code path**

`DynamicModelProvider.__init__` already does `asyncio.create_task(self.initialize())`. This explicit `await provider.initialize()` invokes it a second time. Today it's benign by accident — the explicit call's only awaited coroutine (`listen_for_model_config_change`) has no inner awaits, so it runs to completion before the create_task version is scheduled, and by the time create_task runs `_internal_app` is set and the skip_bootstrap branch returns immediately. But the contract is fragile: adding any `await` inside `listen_for_model_config_change` would either cause a real race or doubled SAC-flow setup.

Recommend picking one initialization path. Cleanest is to remove `asyncio.create_task(self.initialize())` from `__init__` and update the existing `start_model_listener` to also `await provider.initialize()` explicitly — that way the contract is obvious to future callers and there's no fire-and-forget magic in the constructor.

---
*[Added by Claude Code]*

src/solace_agent_mesh/agent/protocol/event_handlers.py
```diff
@@ -1084,9 +1084,63 @@ def _send_error_response_and_nack(
     component.handle_error(exception, Event(EventType.MESSAGE, message))
 
 
+_LAZY_OVERRIDE_PROVIDER_SENTINEL = "__lazy_override_resolver__"
```
bwiebe-solace, line 1087: 🟡 **Important: sentinel can be collision-named via the platform UI**

The PR's safety property — that `topic_model_id == self.model_provider.model_id` is unreachable for the lazy provider — relies on no real alias matching this string. But the listener also subscribes to `configuration/model/{model_id}` (the per-model_id update topic) in `_ensure_config_listener_flow_is_running`. If anyone creates a model alias literally named `__lazy_override_resolver__`, platform updates on that alias will call `update_litellm_model` on the agent's static LiteLlm — exactly the outcome the sentinel is supposed to prevent.

You already offered the fix in the PR body: switch to a UUID-shaped sentinel. Recommend taking it. Cost is one line; collision becomes mathematically impossible. e.g.:

```python
from uuid import uuid4
_LAZY_OVERRIDE_PROVIDER_SENTINEL = f"__override_resolver_{uuid4().hex}__"
```

Even better as a follow-up (separate change): conditionally skip the per-model_id subscription in `_ensure_config_listener_flow_is_running` when `skip_bootstrap=True`, which removes the collision footgun at the source rather than relying on the UUID's improbability.

---
*[Added by Claude Code]*

src/solace_agent_mesh/agent/protocol/event_handlers.py
```diff
@@ -1084,9 +1084,63 @@ def _send_error_response_and_nack(
...
+    never replaced.
+
+    Returns the provider, or ``None`` if the component has no litellm
+    instance to back the resolution flow.
+    """
+    if component._dynamic_model_provider is not None:
+        return component._dynamic_model_provider
+
+    if component._dynamic_model_provider_init_lock is None:
+        component._dynamic_model_provider_init_lock = asyncio.Lock()
```
bwiebe-solace, line 1110: 🟢 **Suggestion: document or eliminate the lock-init pattern**

This reads as a check-then-act race but is actually safe in asyncio because there's no `await` between the check and the assignment, so two coroutines on the same loop can't both observe `None`. Worth either a one-line comment ("Safe under asyncio: no `await` between check and set"), or eagerly initializing `_dynamic_model_provider_init_lock = asyncio.Lock()` in `SamComponentBase.__init__` so this branch goes away entirely. The latter is slightly cleaner and removes a subtle correctness argument from this hot path.

---
*[Added by Claude Code]*

bwiebe-solace (approved):
Ignoring Claude's comments for now. The issues flagged are not problematic, more just correctness things, and we want to be able to run our blitz so let's just go as-is

PR #1501, opened 2026-05-05
docs(DATAGO-131495): add Offline Evaluations UI documentation

ziyanwan:
## Summary

- Adds a new enterprise documentation page for the Offline Evaluations UI feature (`docs/docs/documentation/enterprise/offline-evaluations.md`)
- Covers storage configuration, dataset management, evaluator types, experiment setup, triggering runs, viewing results, exporting, and troubleshooting
- All field constraints, env var defaults, run statuses, RBAC scopes, and UI behavior verified directly against the enterprise source code

## What the page covers

- **Storage Configuration** — env vars and defaults for object storage, dataset generation tuning, and run execution tuning, plus the single-replica deployment constraint
- **Datasets** — manual creation, CSV/JSON import (with format examples), AI-assisted generation, and export
- **Evaluators** — heuristic types (rouge, levenshtein, valid_json, json_diff) and LLM-as-judge creation with prompt template guidance
- **Experiments** — form walkthrough including the Advanced Options accordion for runs per example
- **Viewing Results** — experiment report grid, example inspector side panel, Open Full Details full-page view, and the Files & Activity panel (A2A flow chart)
- **Exporting** — CSV vs JSON export field differences (reasoning in JSON only)
- **Troubleshooting** — six concise entries covering the most common failure modes

### DIFF
--- a/docs/docs/documentation/enterprise/agent-builder.md
+++ b/docs/docs/documentation/enterprise/agent-builder.md
@@ -89,6 +89,8 @@ When you deploy an agent, the system records its configuration. If you later edi
 
 Downloading agents as YAML files provides portability and version control. These files support backing up agent configurations, sharing configurations between deployments, tracking configuration changes in version control systems, and deploying agents using infrastructure-as-code tools.
 
+## Offline Evaluations
+
 Once an agent is deployed and available in Chat, you can set up offline evaluations to measure and track its quality over time. Evaluations let you run a curated set of prompts against the live agent, score the responses, and compare results across model updates or configuration changes — without modifying the running system. See [Offline Evaluations](offline-evaluations.md).
 
 ## Downloading Agent Configurations
--- a/docs/docs/documentation/enterprise/agent-builder.md
+++ b/docs/docs/documentation/enterprise/agent-builder.md
@@ -89,7 +89,7 @@ When you deploy an agent, the system records its configuration. If you later edi
 
 Downloading agents as YAML files provides portability and version control. These files support backing up agent configurations, sharing configurations between deployments, tracking configuration changes in version control systems, and deploying agents using infrastructure-as-code tools.
 
-Once an agent is deployed and available in Chat, you can set up offline evaluations to measure and track its quality over time. Evaluations let you run a curated set of prompts against the live agent, score the responses, and compare results across model updates or configuration changes — without modifying the running system. See [Offline Evaluations UI](offline-evaluations.md).
+Once an agent is deployed and available in Chat, you can set up offline evaluations to measure and track its quality over time. Evaluations let you run a curated set of prompts against the live agent, score the responses, and compare results across model updates or configuration changes — without modifying the running system. See [Offline Evaluations](offline-evaluations.md).
 
 ## Downloading Agent Configurations
 
@@ -145,4 +145,4 @@ For information about connector-related capabilities, see [Connectors](connector
 
 In addition to creating agents through Agent Builder, you can connect external A2A agents that run on separate infrastructure. External agents communicate over HTTPS and integrate with your mesh through a proxy. For more information, see [Connect External Agents](connect-external-agents.md).
 
-Once your agents are deployed, use the [Offline Evaluations UI](offline-evaluations.md) to measure their quality and track performance over time.
+Once your agents are deployed, use [Offline Evaluations](offline-evaluations.md) to measure their quality and track performance over time.
--- a/docs/docs/documentation/enterprise/agent-builder.md
+++ b/docs/docs/documentation/enterprise/agent-builder.md
@@ -89,6 +89,8 @@ When you deploy an agent, the system records its configuration. If you later edi
 
 Downloading agents as YAML files provides portability and version control. These files support backing up agent configurations, sharing configurations between deployments, tracking configuration changes in version control systems, and deploying agents using infrastructure-as-code tools.
 
+Once an agent is deployed and available in Chat, you can set up offline evaluations to measure and track its quality over time. Evaluations let you run a curated set of prompts against the live agent, score the responses, and compare results across model updates or configuration changes — without modifying the running system. See [Offline Evaluations UI](offline-evaluations.md).
+
 ## Downloading Agent Configurations
 
 The Download button allows you to export agent configurations as YAML files. These files are designed for the **Agent Deployer** running in Kubernetes/Helm environments.
@@ -142,3 +144,5 @@ For information about connector-related capabilities, see [Connectors](connector
 ## Related Topics
 
 In addition to creating agents through Agent Builder, you can connect external A2A agents that run on separate infrastructure. External agents communicate over HTTPS and integrate with your mesh through a proxy. For more information, see [Connect External Agents](connect-external-agents.md).
+
+Once your agents are deployed, use the [Offline Evaluations UI](offline-evaluations.md) to measure their quality and track performance over time.
--- a/docs/docs/documentation/enterprise/enterprise.md
+++ b/docs/docs/documentation/enterprise/enterprise.md
@@ -5,27 +5,19 @@ sidebar_position: 700
 
 # Agent Mesh Enterprise
 
-Agent Mesh Enterprise extends the open-source framework with features that enterprise environments require. This version provides:
-
-- Security through single sign-on integration
-- Access control through role-based permissions
-- Data management for cost optimization
-- Observability tools for monitoring agent workflows and system performance
-- Offline evaluations for measuring and tracking agent quality over time
-
-Enterprise is available as a self-managed container image that you can deploy in your own infrastructure. You can obtain access by joining the pilot program at [solace.com/solace-agent-mesh-pilot-registration](https://solace.com/solace-agent-mesh-pilot-registration/).
+Agent Mesh Enterprise extends the open-source framework with features that enterprise environments require. It is available as a self-managed container image that you can deploy in your own infrastructure. You can obtain access by joining the pilot program at [solace.com/solace-agent-mesh-pilot-registration](https://solace.com/solace-agent-mesh-pilot-registration/).
 
 ## Enterprise Features
 
-The Enterprise version includes several capabilities beyond the Community edition.
+Agent Mesh Enterprise provides the following capabilities beyond the Community edition:
 
-Authentication and authorization integrate with your existing identity systems through SSO, eliminating the need for separate credentials and maintaining security standards. You can configure role-based access control to implement authorization policies that determine which agents and resources each user can access through the Agent Mesh Gateways. The Platform Service shares this same authentication and authorization stack with the WebUI Gateway, so a single configuration secures both services. For details, see [Authentication and Authorization](platform-service-auth.md).
+- **Authentication and authorization** — Integrates with your existing identity systems through SSO, eliminating the need for separate credentials and maintaining security standards. You can configure role-based access control to implement authorization policies that determine which agents and resources each user can access through the Agent Mesh Gateways. The Platform Service shares this same authentication and authorization stack with the WebUI Gateway, so a single configuration secures both services. For details, see [Authentication and Authorization](platform-service-auth.md).
 
-Data management features help you optimize costs and improve accuracy. Filtering capabilities reduce unnecessary compute expenses, and data governance helps prevent hallucinations by controlling what information reaches your language models.
+- **Data management** — Helps you optimize costs and improve accuracy. Filtering capabilities reduce unnecessary compute expenses, and data governance helps prevent hallucinations by controlling what information reaches your language models.
 
-Observability tools provide visibility into your agent ecosystem. The built-in workflow viewer tracks LLM interactions and agent communications in real time, allowing you to monitor performance, diagnose issues, and understand system behavior.
+- **Observability** — Provides visibility into your agent ecosystem. The built-in workflow viewer tracks LLM interactions and agent communications in real time, allowing you to monitor performance, diagnose issues, and understand system behavior.
 
-Offline evaluations let you measure and track agent quality against a prepared set of prompts without touching the running system. You define datasets of example prompts, create reusable evaluators that score responses, combine them into experiments that target specific agents, and trigger runs on demand. Results are stored in the database and can be compared across model updates, prompt changes, or configuration adjustments over time.
+- **Offline evaluations** — Lets you measure and track agent quality against a prepared set of prompts without touching the running system. You define datasets of example prompts, create reusable evaluators that score responses, combine them into experiments that target specific agents, and trigger runs on demand. Results are stored in the database and can be compared across model updates, prompt changes, or configuration adjustments over time.
 
 ## Getting Started with Enterprise
 
--- a/docs/docs/documentation/enterprise/enterprise.md
+++ b/docs/docs/documentation/enterprise/enterprise.md
@@ -5,7 +5,7 @@ sidebar_position: 700
 
 # Agent Mesh Enterprise
 
-Agent Mesh Enterprise extends the open-source framework with features that enterprise environments require. This version provides security through single sign-on integration, access control through role-based permissions, data management for cost optimization, and observability tools for monitoring agent workflows and system performance.
+Agent Mesh Enterprise extends the open-source framework with features that enterprise environments require. This version provides security through single sign-on integration, access control through role-based permissions, data management for cost optimization, observability tools for monitoring agent workflows and system performance, and offline evaluations for measuring and tracking agent quality over time.
 
 Enterprise is available as a self-managed container image that you can deploy in your own infrastructure. You can obtain access by joining the pilot program at [solace.com/solace-agent-mesh-pilot-registration](https://solace.com/solace-agent-mesh-pilot-registration/).
 
@@ -18,6 +18,8 @@ Authentication and authorization integrate with your existing identity systems t
 Data management features help you optimize costs and improve accuracy. Filtering capabilities reduce unnecessary compute expenses, and data governance helps prevent hallucinations by controlling what information reaches your language models.
 
 Observability tools provide visibility into your agent ecosystem. The built-in workflow viewer tracks LLM interactions and agent communications in real time, allowing you to monitor performance, diagnose issues, and understand system behavior.
+
+Offline evaluations let you measure and track agent quality against a prepared set of prompts without touching the running system. You define datasets of example prompts, create reusable evaluators that score responses, combine them into experiments that target specific agents, and trigger runs on demand. Results are stored in the database and can be compared across model updates, prompt changes, or configuration adjustments over time.
 
 ## Getting Started with Enterprise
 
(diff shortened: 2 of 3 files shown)


docs/docs/documentation/enterprise/enterprise.md
```diff
@@ -5,7 +5,7 @@ sidebar_position: 700
 
 # Agent Mesh Enterprise
 
-Agent Mesh Enterprise extends the open-source framework with features that enterprise environments require. This version provides security through single sign-on integration, access control through role-based permissions, data management for cost optimization, and observability tools for monitoring agent workflows and system performance.
+Agent Mesh Enterprise extends the open-source framework with features that enterprise environments require. This version provides security through single sign-on integration, access control through role-based permissions, data management for cost optimization, observability tools for monitoring agent workflows and system performance, and offline evaluations for measuring and tracking agent quality over time.
```
bwiebe-solace: Suggestion: this is starting to become quite a long list-sentence, should we consider bullet-points for easier reading?

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,327 @@
+---
+title: Offline Evaluations UI
+sidebar_position: 850
+---
+
+# Offline Evaluations UI
+
+:::warning Experimental Feature
+This feature is under active development. Configuration schemas and behavior may change in future releases. Do not use in production without validating against the current release notes.
```
bwiebe-solace, line 9: I know we said in the past that we don't want to document feature flags, but will having this page (and the other mentions in the docs) potentially confuse users when they don't see this feature in their environment? If we don't want to document the exact flag, do we want to add a note saying something like "This feature is under active development and may not be available in all deployments."? That way we keep the docs available for the customers that we're going to ship this to, but note that not everyone will see it right now.

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,327 @@
...
+:::
+
+Offline evaluations let you measure agent quality by sending a curated set of prompts to a deployed agent, scoring the responses, and tracking results over time. "Offline" means the inputs are prepared in advance — not sampled from live production traffic. This makes it safe to run evaluations repeatedly against the same prompt set and compare results across model changes, prompt updates, or configuration adjustments.
```
bwiebe-solace: The word "safe" feels wrong in "...makes it **safe** to run evaluations repeatedly..." . Suggestion: "...makes it **possible** to run evaluations repeatedly..."

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,327 @@
...
+- At least one agent deployed and reachable through the event broker.
+- The Platform Service running with a database configured. SQLite works for development; use PostgreSQL for production. For database configuration guidance, see [Installing Agent Mesh Enterprise](installation.md).
+- Object storage configured if you want to persist execution data and artifact snapshots beyond the local filesystem. See [Storage Configuration](#storage-configuration) below.
+
+## Storage Configuration
+
+The evaluation execution service stores execution data and agent-produced artifact snapshots in object storage. Two backends are available, selected automatically based on environment variables.
+
+:::warning
+The evaluation execution service assumes a single active instance at a time. Running multiple replicas of the Platform Service may cause one instance to mark another's in-progress runs as failed. Set `replicas: 1` (or the equivalent in your deployment platform) for the Platform Service.
```
bwiebe-solace: This shouldn't be true: multiple executors should be supported and not step on each other.

Each experiment run has a timeout attached based on the start time, which is stored in the DB. If multiple platform service replicas are running, then one won't (well, maybe "shouldn't" is the right word - this path hasn't been well-tested) automatically mark anything it doesn't own as failed. There is a scheduled task in each replica that will check for experiments that exceed their timeout period and fail those (terminating the execution if that replica has the running experiment task, or just marking in the DB to be picked-up by the other next time).

Similarly, if there are multiple replicas then only one should receive a task and there's guarding in-place to ensure that only one will end up running a task.

I'll leave it up to you on if you want to leave this note as just a safeguard for now, since it's not well-tested like I said, but it should technically be incorrect.

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Object Storage
+
+| Environment Variable | Required | Default | Description |
+|---|---|---|---|
+| `EVAL_DATA_BUCKET_NAME` | No | — | S3, GCS, or Azure Blob container name. When set, the service uses cloud object storage. |
+| `OBJECT_STORAGE_FS_ROOT` | No | `<cwd>/tmp/eval-storage` | Filesystem root when no bucket is configured. Defaults to a `tmp/eval-storage` directory relative to the process working directory. |
+
+:::warning
+The local filesystem backend is suitable for development only. Data stored under the default `tmp/eval-storage` path is not replicated and will be lost if the container restarts without a persistent volume. Set `EVAL_DATA_BUCKET_NAME` for any deployment where you want to retain execution traces and artifact snapshots.
```
bwiebe-solace: "...suitable for **development** only." - Is "development" the right word here?

I'd suggest maybe "experimentation" or rephrasing this sentence to something like "The local filesystem backend is not suitable for production deployments and is only intended to provide a quick way to try-out offline evaluations with minimal infrastructure."

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,327 @@
...
+
+The Reports page is the entry point for the Evaluations section. It shows a paginated table of all run groups — one row per time you clicked "Run" on an experiment. From this page you can:
+
+- See the status, start time, duration, and result count for each run group.
+- Select a run group row to open the experiment report for that run.
+- Filter the run group list by experiment name or target agent.
+
+A run group corresponds to one trigger of a specific experiment. If the experiment targets multiple language model configurations, each configuration produces one run within the group.
+
+The watchlist section at the top of the Reports page shows score trend charts for up to five agents you pin. Select agents to watch from the Edit Watchlist dialog, which appears when you click the edit control. Each agent's chart shows one series per experiment that targets it, plotting the primary evaluator score over time. Use the watchlist to spot regressions across deploys or model changes.
```
bwiebe-solace: Suggestion: "pin" -> "select" in "...five agents you **pin**."

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,327 @@
...
+- `rouge`: Weighted ROUGE F1 score (20% R1, 30% R2, 50% RL) measuring text overlap with the expected response.
+- `levenshtein`: Normalized edit-distance similarity (0.0–1.0) comparing the response to the expected response.
+- `valid_json`: Returns 1.0 if the response is valid JSON, 0.0 otherwise.
+- `json_diff`: Recursive structural similarity (0.0–1.0) between two JSON objects.
+
+Heuristic evaluators that use an expected response produce a score of 0.0 when the example has no expected response.
+
+System-seeded evaluators cannot be modified or deleted. The name column shows which evaluators are system-seeded.
+
+#### LLM-as-judge evaluators
```
bwiebe-solace: We mention the built-ins for heuristic evaluators, do we need to mention the built-in LLM-as-judge evaluators here?

bwiebe-solace: We've used "LLM-as-a-judge" in the UI, but there are instances in non-user-facing code which uses "LLM-as-judge". I think we should align our docs on the one with "-a-" for now, then we can re-visit the labeling post-MVP if there are objections.

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,327 @@
...
+- **Describe the full range of quality.** Your criteria text pairs with the choice score labels you define below. Writing criteria that naturally maps to those labels (for example, calling out "complete", "partial", and "missing") makes the judge's scoring more consistent.
+- **Keep it focused on one dimension per evaluator.** An evaluator that tries to assess accuracy, tone, and format simultaneously produces scores that are hard to act on. Create separate evaluators for separate concerns and attach all of them to the experiment.
+
+##### Choice scores {#choice-scores}
+
+Choice scores map natural-language outcome labels to numeric scores. The judge LLM picks one label per evaluation; its associated score becomes the numeric result for that example.
+
+**How it works:** The platform presents your labels to the judge as lettered options `(A)`, `(B)`, `(C)`… in the order they are defined. The judge returns a letter; the platform maps it back to the label and then to the score. The label (not the letter) is stored in the result for the audit trail.
```
bwiebe-solace: Do we need to expose this level of detail? We chose to make this internal mapping transparent to the user as it was confusing when we had to use explicit single-letters in our criteria and add a corresponding choice. We also only do this to help improve the output quality from the LLM (ie. so that it only has to choose a single token to return instead of a compound set of tokens), it doesn't really impact the user experience at all since we map everything back to friendly labels.

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,327 @@
...
+The inspector header has an **Open Full Details** button that opens the example on a dedicated full-page view. The full-page view shows the same prompt, expected response, and one Run card per trial. Each model column in a Run card has a network icon button in its header. Clicking it opens a side panel with two tabs:
+
+- **Files** — artifacts produced by the agent for that specific run and model, with the same preview and download behavior as the main inspector.
+- **Activity** — the captured A2A event stream rendered as a flow chart, showing the message flow between the orchestrator, the target agent, and any peer agents involved in producing the response.
+
+If no task events were captured for a result, the Activity tab shows an "Activity diagram unavailable" message.
+
+### Exporting results
+
+You can export the results for a completed run as CSV or JSON. Navigate to the run group, expand the run you want to export, and use the export button. Both formats include the prompt, expected response, agent response, duration, and a score and pass/fail per evaluator for each example. The JSON export additionally includes the evaluator reasoning for each score.
```
bwiebe-solace: We're removing this for MVP if I understand correctly

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,327 @@
...
+
+### AI-assisted dataset generation fails
+
+**Invalid JSON response:** The orchestrator's model produced output the platform could not parse. Reduce the number of requested examples and retry, or switch to a model with stronger instruction-following.
+
+**Timeout:** The orchestrator did not respond within `EVAL_DATASET_GEN_TIMEOUT` seconds. Increase this value (300–600 seconds for large datasets), or generate a smaller batch and append additional examples in a second pass.
+
+### Run shows "completed_with_warnings"
+
+At least one example timed out or returned an error from the agent. Open the result inspector for an errored example and check `errorMessage`. If many examples error, consider increasing `EVAL_AGENT_TIMEOUT_SECONDS` or reducing the experiment's concurrent load.
```
bwiebe-solace: "...reducing the experiment's concurrent load." - What does that mean to a user since we got rid of the "experiment speed" concept? We should probably explain it better with options, such as: "...reducing the experiment's number of runs per example or dataset size."

bwiebe-solace (commented):
Looks good overall, left some comments to consider

docs/docs/documentation/enterprise/offline-evaluations.md
bwiebe-solace: This has been put to the epic channel, will leave it as-is for now pending that decision

docs/docs/documentation/enterprise/agent-builder.md
jessieli-ux, line 1: I feel like Evals deserves its own titled section, not just under "Manage Deployed Agents". For Manage Deployed Agents, I was expecting things that are more CRUD-based.

docs/docs/documentation/enterprise/enterprise.md
jessieli-ux, line 1: I like how the features are bulletted in the first section, can they be bolded or titled in the Enterprise Features section as well so that it's easier to scan?

jessieli-ux: Or combine them, cause right now Enterprise Features is kind of overlapping with both the section above and below.

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,351 @@
...
+
+Offline evaluations provide a browser-based interface for managing the building blocks of this workflow: datasets of example prompts, evaluators that define how responses are scored, and experiments that combine them into reusable run configurations. For the CLI-based evaluation workflow that works with Community and Enterprise deployments alike, see [Evaluating Agents](../developing/evaluations.md).
+
+## What Offline Evaluations Do
+
+Offline evaluations let you continuously measure the quality of your deployed agents without modifying their configuration or writing test suite files. You define a dataset of prompts, select or create an LLM-as-a-judge evaluator, combine them into an experiment that targets a specific agent, and trigger runs on demand. The platform sends each prompt to the agent through the event broker, captures the response, scores it using the evaluator, and stores the results for inspection and comparison.
```
jessieli-ux: We have more than just LLM-as-a-judge evaluators. Should it say something like "create an LLM-as-a-judge evaluator or choose from pre-set LLM and heuristic evaluators"

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,351 @@
...
+| `EVAL_AGENT_TIMEOUT_SECONDS` | `480` | Maximum seconds to wait for the target agent to respond to a single example prompt. |
+| `EVAL_EXAMPLE_BUDGET_SECONDS` | `600` | Total budget per example, including the agent timeout and scoring time. |
+| `EVAL_WATCHDOG_INTERVAL_SECONDS` | `120` | Interval at which the watchdog checks for stale or stuck runs. |
+| `EVAL_SCORING_THREADS` | `8` | Thread pool size for parallel LLM-as-a-judge scoring calls. |
+
+## Reports
+
+The Reports page is the entry point for the Evaluations section. It shows a paginated table of all run groups — one row per time you clicked "Run" on an experiment. From this page you can:
+
+- See the status, start time, duration, and result count for each run group.
```
jessieli-ux: Let's keep the name consistent and call these objects "reports" to match the UI.

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,351 @@
...
+
+The watchlist section at the top of the Reports page shows score trend charts for up to five agents. Select agents to watch from the Edit Watchlist dialog, which appears when you click the edit control. Each agent's chart shows one series per experiment that targets it, plotting the primary evaluator score over time. Use the watchlist to spot regressions across deploys or model changes.
+
+## Experiment Lab
+
+The Experiment Lab contains three tabs — Datasets, Evaluators, and Experiments — where you manage the building blocks of evaluations.
+
+### Datasets {#datasets}
+
+A dataset gives your agents a benchmark for experiments. It contains a set of prompts — optionally paired with expected responses — that are sent to the agent during a run. Add examples manually or upload a file to get started, then use the dataset across as many experiments as you need.
```
jessieli-ux: We also have the AI generated option too right?

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,351 @@
...
+```json
+[
+  { "prompt": "What is the capital of France?", "expected_response": "Paris" },
+  { "prompt": "Summarize this quarter's sales trends." }
+]
+```
+
+#### AI-assisted dataset generation
+
+The "Generate" option creates a dataset populated by the orchestrator agent based on the target agent's metadata. Select a target agent, specify the number of examples (1–100), and optionally provide a focus hint that the orchestrator treats as a hard requirement for every generated example.
```
jessieli-ux: What's a focus hint? Is that what it's called in the UI?

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Evaluators {#evaluators}
+
+An evaluator defines how the agent's response is scored. Define your criteria, and an LLM automatically scores every agent response in your experiments against the provided dataset — giving you clear, actionable insight into how well your agents are performing.
+
+Two types are available.
+
+#### Heuristic evaluators
+
+Heuristic evaluators are built-in and cannot be created or deleted through the UI. They are seeded by the platform at startup. Available heuristic types are:
```
jessieli-ux: "created"?

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,351 @@
...
+
+- `rouge`: Weighted ROUGE F1 score (20% R1, 30% R2, 50% RL) measuring text overlap with the expected response.
+- `levenshtein`: Normalized edit-distance similarity (0.0–1.0) comparing the response to the expected response.
+- `valid_json`: Returns 1.0 if the response is valid JSON, 0.0 otherwise.
+- `json_diff`: Recursive structural similarity (0.0–1.0) between two JSON objects.
+
+Heuristic evaluators that use an expected response produce a score of 0.0 when the example has no expected response.
+
+System-seeded evaluators cannot be modified or deleted. The name column shows which evaluators are system-seeded.
```
jessieli-ux: System-created

jessieli-ux (changes requested):

docs/docs/documentation/enterprise/agent-builder.md
solace-mdupls: Agreed

docs/docs/documentation/enterprise/enterprise.md
```diff
@@ -5,7 +5,13 @@ sidebar_position: 700
 
 # Agent Mesh Enterprise
 
-Agent Mesh Enterprise extends the open-source framework with features that enterprise environments require. This version provides security through single sign-on integration, access control through role-based permissions, data management for cost optimization, and observability tools for monitoring agent workflows and system performance.
+Agent Mesh Enterprise extends the open-source framework with features that enterprise environments require. This version provides:
```
solace-mdupls: "version" is maybe overloaded here. I this version is being used as a way to describe the Enterprise vs. Community edition of the product. More clarity would be "Agent Mesh Enteprise provides:"

docs/docs/documentation/enterprise/enterprise.md
```diff
@@ -5,7 +5,13 @@ sidebar_position: 700
...
+
+- Security through single sign-on integration
+- Access control through role-based permissions
+- Data management for cost optimization
+- Observability tools for monitoring agent workflows and system performance
+- Offline evaluations for measuring and tracking agent quality over time
```
solace-mdupls: A list of features here - seems like it would be best added to the "Enterprise features" section. It's a bit strange that it's called out immediately and then again right after.  There's some duplication like "This version provides" and "The Enterprise version includes several capabilities beyond the Community edition."

It could be simplified by consolidating the two sections

docs/docs/documentation/enterprise/offline-evaluations.md
```diff
@@ -0,0 +1,351 @@
+---
+title: Offline Evaluations
```
solace-mdupls, line 2: Similar to the "Application Metrics" (OTEL) documentation, the suggestion from @michellefthomas was to simplify by splitting the docs into smaller targeted pages. Can you consider doing that here as well? That may provide better clarity.

docs/docs/documentation/enterprise/agent-builder.md
ziyanwan: Separate section is added in the new commit.

docs/docs/documentation/enterprise/enterprise.md
ziyanwan: They are bolded in both sections now.

docs/docs/documentation/enterprise/offline-evaluations.md
ziyanwan: Addressed.

docs/docs/documentation/enterprise/offline-evaluations.md
ziyanwan: It is "additional instructions". Corrected.

docs/docs/documentation/enterprise/enterprise.md
ziyanwan: I will leave this part for now, and only do addition on top of the current structure.  Style was enhanced per Jessie's suggestion.

docs/docs/documentation/enterprise/offline-evaluations.md
ziyanwan: I tried, and it doesn't look nice, as each object relates to each other. 
Also, this will change the url of the sections. The URL is already used in the UI. 
We can revisit this after the release.

PR #1531, opened 2026-05-13
docs(DATAGO-136350): default EVAL_DATA_BUCKET_NAME so Quickstart K8s installs retain eval artifacts

josephabonasara:

Two docs follow-ups from the offline-evaluations Slack thread:

1. **Mahmood's ask** (via Ziyang): the offline-evaluations docs should reflect the offlineEvals storage configurations — particularly that on Kubernetes the Helm chart now sets `EVAL_DATA_BUCKET_NAME` automatically, so the existing warning ("set EVAL_DATA_BUCKET_NAME for any deployment…") was over-broad for K8s operators.
2. **Hugo's ask**: the production docs were missing S3 bucket setup instructions for evals — parallel to the existing *S3 Buckets for OpenAPI Connector Specs* section under External services.

Sequenced after [SolaceDev/sam-kubernetes#228](https://github.com/SolaceDev/sam-kubernetes/pull/228) — the chart change that introduces the `EVAL_DATA_BUCKET_NAME` auto-injection these docs now describe.

- `docker-installation.md`: added a new **Infrastructure Setup: S3 Bucket for Eval Data (Offline Evaluations)** section parallel to the existing OpenAPI one. Structure mirrors the OpenAPI section (When required / Setup Instructions / Security Notes) so the doc reads consistently.
- `offline-evaluations.md`: added an inline cross-link to the new installation section, matching the established `openapi-connectors.md → docker-installation.md` pattern (relative `.md` link, single sentence with K8s-handled-automatically note). Reworded the filesystem-fallback warning so it explicitly targets non-Kubernetes deployments.

Files changed: 2. Net diff: `+31/-1`.

### Key Design Decisions

- **Matched the OpenAPI section's rhythm, not invented a new shape.** The OpenAPI section is concise (no "Why share by default?" subsection, no separate Azure/GCS block). Earlier drafts of the eval section had both — I dropped them to keep the structure parallel and the doc maintainable.
- **Inline prose for the K8s callout, not an admonition.** The existing convention (e.g., `openapi-connectors.md:42`) uses a one-sentence note linking to the canonical infrastructure-setup section. Using `:::info Kubernetes deployments` admonitions would have introduced a style the codebase doesn't use here.
- **Linked to `docker-installation.md`, not `production-kubernetes.md`.** The OpenAPI cross-reference does the same — `docker-installation.md` is the canonical "platform admin infrastructure setup" doc. K8s users read the inline note ("Helm charts handle this automatically") and don't need to follow the standalone recipe.
- **Optional bucket framing.** Unlike OpenAPI specs (which *require* a separate bucket for the public-read pattern), the eval bucket is *optional* — by default eval data shares the artifacts bucket via the `{namespace}/eval/runs/...` key prefix. The doc leads with this so administrators don't think they have to provision a new bucket.

- [x] Manual testing: Both files render in the Docusaurus dev server (`docs/`); cross-link from `offline-evaluations.md` resolves to the new anchor in `docker-installation.md`.
- [x] Anchor verification: GitHub auto-generates the slug `infrastructure-setup-s3-bucket-for-eval-data-offline-evaluations` from the heading; the inline link uses the matching anchor.
- [ ] Unit tests: N/A (docs-only change)
- [ ] Integration tests: N/A
- [ ] Known limitations: I did not add a parallel mention to `production-kubernetes.md`. That doc has a much more detailed K8s-specific OpenAPI section (line 321); a parallel eval-data block could be useful for operators reading that doc cover-to-cover, but the cross-link convention already directs both K8s and standalone administrators to `docker-installation.md`. Happy to add one if reviewers prefer.

- The link target in `offline-evaluations.md` (`docker-installation.md#infrastructure-setup-s3-bucket-for-eval-data-offline-evaluations`) depends on the heading slug. If the heading wording changes in review, please update the anchor in `offline-evaluations.md` to match.
- This PR is **safe to land independently** of [SolaceDev/sam-kubernetes#228](https://github.com/SolaceDev/sam-kubernetes/pull/228), but the K8s callout ("Helm charts handle this configuration automatically") is technically only accurate *after* that PR ships in a chart version. If the chart PR slips, the callout slightly overstates current behavior; if it's a concern, hold this PR until the chart bump.

### DIFF
--- a/docs/docs/documentation/enterprise/docker-installation.md
+++ b/docs/docs/documentation/enterprise/docker-installation.md
@@ -244,34 +244,6 @@ Some enterprise features require additional infrastructure setup. If you plan to
 - Public read is safe for API schemas only
 - Write access should be restricted to the platform
 
-## Infrastructure Setup: S3 Bucket for Eval Data (Offline Evaluations)
-
-If you plan to use the [Offline Evaluations](offline-evaluations.md) feature in production, the platform service should store execution data and artifact snapshots in object storage. Standalone deployments can either set `EVAL_DATA_BUCKET_NAME` to a bucket, or run with the default local filesystem backend at `<cwd>/tmp/eval-storage` — the latter works correctly for development and local testing, but data is wiped if the container restarts without a persistent volume. Kubernetes deployments handle this configuration automatically via Helm charts.
-
-### When do you need a separate eval data bucket?
-- **Recommended for production standalone deployments** to retain eval data across container restarts (the local filesystem fallback is ephemeral in containerized environments without a persistent volume)
-- **Not required for Kubernetes by default** — Helm charts handle this automatically and share the artifacts bucket unless you set a dedicated eval data bucket (requires sam-kubernetes chart version 1.501.0 or later)
-- **Optional in any deployment** when you want different IAM, retention, or replication policies for eval data than for artifacts
-
-### Setup Instructions
-
-1. **Create the eval data S3 bucket** (private, authenticated read/write):
-   ```bash
-   aws s3 mb s3://my-eval-data-bucket --region us-west-2
-   ```
-
-2. **IAM permissions for the SAM service**:
-   - Grant `s3:PutObject`, `s3:GetObject`, `s3:DeleteObject`, `s3:ListBucket` for this bucket.
-
-3. **Kubernetes deployments**: Set `dataStores.s3.evalDataBucketName` (or the Azure / GCS equivalent) in your Helm chart values. Leave empty to share the artifacts bucket.
-
-4. **Standalone deployments**: Set `EVAL_DATA_BUCKET_NAME` on the platform service. When unset, eval data falls back to an ephemeral local filesystem path.
-
-### Security Notes
-- Keep the bucket private — no public read access is needed
-- Apply lifecycle rules to age out historical eval data if desired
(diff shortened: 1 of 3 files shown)


docs/docs/documentation/enterprise/docker-installation.md
```diff
@@ -244,6 +244,34 @@ Some enterprise features require additional infrastructure setup. If you plan to
 - Public read is safe for API schemas only
 - Write access should be restricted to the platform
 
+## Infrastructure Setup: S3 Bucket for Eval Data (Offline Evaluations)
```
moodiRealist: @josephabonasara I don't think we need to document under the docker-installation.md , please instead update the docs under https://solacelabs.github.io/solace-agent-mesh/docs/documentation/enterprise/production-kubernetes , if you search for `connectorSpecBucketName` in that page, you'd get a good idea the level of documentation we'd need for offline evals bucket.

moodiRealist (changes requested):

docs/docs/documentation/enterprise/docker-installation.md
moodiRealist: We can have a chat btw, and I can show you the 4 new docs we added for the Z2H milestone.

moodiRealist (approved):

PR #1542, opened 2026-05-15
fix(DATAGO-121653): reject prompt imports with invalid command characters

solace-jkhalife:

When importing a prompt with a `command` field containing characters outside `[a-zA-Z0-9_-]` (e.g. spaces, dots, `@`), the import would silently succeed but the resulting row would be dropped from all list responses — making the prompt invisible after import.  The fix validates the command pattern at the API layer and in the import dialog UI before the request is submitted.

- **Backend** (`routers/prompts.py`): Added a regex check against `^[a-zA-Z0-9_-]+$` on the `command` field in the `import_prompt` endpoint. Returns HTTP 400 with the offending value in the error message if validation fails. Only applied when `preserve_command=true`; ignored otherwise.
- **Frontend** (`PromptImportDialog.tsx`): Added `COMMAND_PATTERN` regex to the Zod form schema for the command field. Added `initialCommandInvalid` state so that files with an invalid command immediately open the field in editable mode (same UX as conflict handling). Added `isValid` to the `isEnabled` gate on the Import button. Fixed `setValue` calls to pass `shouldValidate: true` so react-hook-form recalculates `isValid` after programmatic population.
- **Tests**: Backend integration tests covering rejection of 6 invalid command patterns and pass-through when `preserve_command=false`. Frontend unit tests covering editable input appearance, button disabled state, button re-enable after correction, and read-only display for valid commands.

### Key Design Decisions 

The validation pattern (`^[a-zA-Z0-9_-]+$`) mirrors the existing `PromptGroupResponse` filter — the same characters that cause silent drops from list queries. 

- [ ] Manual testing: import a prompt JSON with a command containing a space or dot — should see 400 / inline error; import with a valid command — should succeed
- [ ] Unit tests: 5 new frontend tests in `PromptImportDialog.test.tsx` (all passing)
- [ ] Integration tests: 2 new backend tests in `test_prompts_validation_and_auth.py` (all passing on SQLite)
- [ ] Known limitations: PostgreSQL integration tests require Docker (not run locally)


lgh-solace (commented):
Looks good!  Can you check out the below comments?

### Please review and fix if confirmed

- **The same `$`-before-newline quirk likely affects the create/update path too.** Python's `re` (without `re.MULTILINE`) treats `$` as matching either end-of-string *or* right before a trailing `\n`, and Pydantic v2's string `pattern` uses those semantics. The Field constraints at [prompt_dto.py:69](src/solace_agent_mesh/gateway/http_sse/routers/dto/prompt_dto.py#L69) and [prompt_dto.py:93](src/solace_agent_mesh/gateway/http_sse/routers/dto/prompt_dto.py#L93) still use `pattern="^[a-zA-Z0-9_-]+$"` — did you consider whether `POST /groups` / `PATCH /groups/{id}` with `"command": "valid-cmd\n"` slips through the same way the import path used to? A quick integration probe would confirm. If it does, the cleanest fix is a `field_validator` using `re.fullmatch` against a shared constant — which would also collapse the three regex copies.

### Other areas to consider

- **One regex, three copies.** `re.fullmatch` literal at [prompts.py:1257](src/solace_agent_mesh/gateway/http_sse/routers/prompts.py#L1257), Pydantic `pattern` at [prompt_dto.py:69](src/solace_agent_mesh/gateway/http_sse/routers/dto/prompt_dto.py#L69) / [:93](src/solace_agent_mesh/gateway/http_sse/routers/dto/prompt_dto.py#L93), and `COMMAND_PATTERN` at [PromptImportDialog.tsx:12](client/webui/frontend/src/lib/components/prompts/PromptImportDialog.tsx#L12). The new code-comment already calls out the coupling ("must match PromptGroupResponse") — extracting one Python constant + one FE constant alongside `PROMPT_FIELD_LIMITS` would prevent the next drift.
- **`promptImportCommandSchema` is unreferenced and duplicates the dialog's form schema.** [prompt-import.schema.ts:62-64](client/webui/frontend/src/lib/schemas/prompt-import.schema.ts#L62-L64) is dead code, and the regex change was made to the inline schema in the dialog instead. Worth deleting it or moving the dialog's `promptImportFormSchema` into the schema file so the rule lives in one place.
- **Backend truncates *before* it validates** at [prompts.py:1251-1257](src/solace_agent_mesh/gateway/http_sse/routers/prompts.py#L1251-L1257) — a 51-char command with an invalid char at position 51 will be sliced off and pass silently. Worth either swapping the order or adding a `# why` note.
- [PromptImportDialog.test.tsx:11-21](client/webui/frontend/src/stories/Prompts/PromptImportDialog.test.tsx#L11-L21) polyfills `Blob.prototype.text` per-file. If jsdom really lacks it in this version, this belongs in `vitest.setup.ts` once rather than copied into every new dialog test — `PromptBuilderChat.test.tsx` doesn't need it, which suggests it may not be necessary here either.
11).

PR #1570, opened 2026-05-31
feat(DATAGO-138153): add Agent Mode — a chat-only view pinned to a single agent

ghaithdallaali:
## Summary
Adds **Agent Mode**: a focused, chat-only rendering of the web UI locked to a single
agent, activated via hash-route query params:

    /#/chat?agentMode=true&agent=<agentName>

When active, the UI pins the conversation to the named agent and presents a
stripped-down chat surface (no agent picker / navigation chrome), suitable for
embedding a single agent's chat experience.

## What changed

- **URL-param entry point** — `getHashQueryParams()` reads Agent Mode params from
  the hash query (not `window.location.search`) so they travel with the hash
  route. `agentMode` is surfaced on the config context and never persisted.
- **Pinned-agent selection** — `selectInitialAgent` extracted as a pure helper
  covering pinned vs priority-order selection. An unresolved `?agent=` shows a
  "Connecting…" state rather than silently falling back to another agent, and
  sending is gated until the pinned agent resolves.
- **Deterministic pinning by wire name** — `?agent=` and message routing resolve
  the agent by its unique wire name only (no `displayName` fallback), so two
  agents sharing a display name can't mislabel the hero or misroute messages.
- **Re-pin on session load** — loading a session no longer lets its stored agent
  override the pinned `?agent=` (the selector is hidden, so a silent re-route
  would be undetectable). Falls back to the stored agent only when the hash query
  is absent.
- **Timeout fallback** — a bad `?agent=` shows a terminal "unavailable" state
  after 15s instead of spinning forever.
- **Chrome stripped in Agent Mode** — Activity panel hidden (with a fallback to
  the Files tab when Activity was active), session move/share actions removed,
  a centered per-agent greeting hero replaces the welcome bubble, and contextual
  prompts are disabled.
- **Neutral login screen** — the login view now shows a bare centered Login
  button with no product branding, so white-labeled deployments stay unbranded
  through the auth flow.

## Tests

- `getHashQueryParams` parsing, including ignoring `window.location.search`
- `selectInitialAgent` pinned vs priority-order coverage; `?agent=` pins by wire
  name even on a shared display name and never resolves a display-name-only match
- ChatPage empty states (connecting / unavailable / ready) with a Full-UI control
- ChatInputArea no-agent-no-send gate
- Nav-hiding tests covering Full UI vs Agent Mode
- AgentModeWelcome hero, session action allowlist

### DIFF
--- a/client/webui/frontend/src/auth/authCallback.tsx
+++ b/client/webui/frontend/src/auth/authCallback.tsx
@@ -1,5 +1,7 @@
 import { useEffect } from "react";
 
+import { consumePostLoginRedirect } from "@/lib/utils/url";
+
 function AuthCallback() {
     useEffect(() => {
         const hash = window.location.hash.substring(1);
@@ -16,8 +18,9 @@ function AuthCallback() {
             if (refreshToken) {
                 localStorage.setItem("refresh_token", refreshToken);
             }
-            // Redirect to the main application page
-            window.location.href = "/";
+            // Redirect back to the URL the user left from (restores Agent Mode
+            // params dropped by the IdP round-trip), or "/" if none was stashed.
+            window.location.href = consumePostLoginRedirect();
         } else {
             console.error("AuthCallback: No access token found in URL hash.");
         }
--- a/client/webui/frontend/src/AppLayout.tsx
+++ b/client/webui/frontend/src/AppLayout.tsx
@@ -2,6 +2,7 @@ import { useEffect, useState, useCallback } from "react";
 import { Outlet, useLocation, useNavigate } from "react-router-dom";
 
 import { NavigationSidebar, CollapsibleNavigationSidebar, ToastContainer, bottomNavigationItems, getTopNavigationItems, EmptyState } from "@/lib/components";
+import { Button } from "@/lib/components/ui/button";
 import { SelectionContextMenu, useTextSelection } from "@/lib/components/chat/selection";
 import { MoveSessionDialog } from "@/lib/components/chat/MoveSessionDialog";
 import { ModelSetupDialog } from "@/lib/components/models/ModelSetupDialog";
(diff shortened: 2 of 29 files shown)


ghaithdallaali:
This PR also attempts to fix an important URL redirect issue that would need to be further tested. For visibility:

```
Problem

The login flow does an OAuth round-trip: the app redirects to the IdP, which redirects back to a fixed callback URL. That round-trip drops the URL you started from — including the ?agentMode=true&agent=X params that pin the UI to a specific agent. Without handling this, you'd land on the default route after login and lose your pinned agent.

Fix: stash-before-leave, restore-after-callback

A one-shot localStorage round-trip in src/lib/utils/url.ts:

- stashPostLoginRedirect() — saves the full window.location.href (hash route + query params and all) to localStorage key sam_post_login_redirect right before redirecting to the IdP.
- consumePostLoginRedirect() — reads it back exactly once (deletes on read), validates it's same-origin (open-redirect protection), and returns it; falls back to "/" if nothing was stashed or the value is unusable.

Where it's wired

Stash is called at every exit-to-login site, so the current URL is preserved no matter how the user reaches login:
- AuthProvider.tsx — the explicit login() action.
- client.ts — when a refresh-token refresh fails.
- client.ts — when a request still returns 401 after a successful refresh.

Restore happens in the callback (authCallback.tsx): after storing the tokens, instead of hard-navigating to /, it navigates back to the stashed URL:
window.location.href = consumePostLoginRedirect();

Net effect

Logging in from a pinned link → URL is stashed → IdP round-trip → callback restores the exact URL → the app reads the params from the hash at load and re-pins the agent. The same-origin guard means a tampered stash value can only ever redirect to "/", never off-site.

Covered in url.test.ts: stash/restore happy path, missing-stash fallback, cross-origin rejection, and one-shot consumption.
```

client/webui/frontend/src/auth/authCallback.tsx
```diff
@@ -16,8 +18,9 @@ function AuthCallback() {
             if (refreshToken) {
                 localStorage.setItem("refresh_token", refreshToken);
             }
-            // Redirect to the main application page
-            window.location.href = "/";
+            // Redirect back to the URL the user left from (restores Agent Mode
+            // params dropped by the IdP round-trip), or "/" if none was stashed.
+            window.location.href = consumePostLoginRedirect();
```
DevSol-UQ, line 23: Enterprise has it's own copy of this and will need to be updated for the fix to work full

ghaithdallaali:
Dupe of https://github.com/SolaceLabs/solace-agent-mesh/pull/1574 

Closing.

PR #1584, opened 2026-06-04
fix(auth): skip sentinel claim values when extracting user identifier

Hugo-Pare:

Authenticated users whose IdP/OAuth proxy returns the literal string `"Unknown"` (or `"null"`/`"none"`/`""`) for one of the high-priority OIDC claims were being silently re-identified as the `sam_dev_user` dev fixture, with their real identity discarded.

**Concrete repro:** an OAuth proxy that substitutes `"Unknown"` for missing claims is enough to trigger this. Azure AD guest accounts often have one or more populated claims missing; the proxy fills in `"Unknown"`; SAM accepts it as a valid identifier; everything downstream then sees `sam_dev_user`. Sessions, artifacts, share permissions, and audit attribution all get logged under the wrong identity.

**Impact:** identity confusion / collision across users whose IdP responses contain any sentinel value. In the worst case, two distinct authenticated users end up sharing the same `sam_dev_user` namespace.

### Root cause

`_extract_user_identifier` (`src/solace_agent_mesh/shared/auth/middleware.py:132-152`) walked the standard OIDC claim priority list with `or`-chaining, then checked **only** if the resulting value was `"Unknown"` — and on that check, returned `"sam_dev_user"` outright instead of continuing the walk:

```python
user_identifier = user_info.get("sub") or user_info.get("client_id") or ... or user_info.get("user_id")
if user_identifier and user_identifier.lower() == "unknown":
    return "sam_dev_user"
return user_identifier
```

So if `sub == "Unknown"` but `email == "alice@example.com"`, the function locked in on `"Unknown"` (because `or`-chaining stops at the first truthy value), saw it was the sentinel, and bailed to the dev user — never looking at the populated `email` claim sitting further down the list. The same trap exists in `_create_user_state` (only the *final* identifier was checked against `{"unknown", "null", "none", ""}`, not each individual claim) and in `_extract_user_details` (the email/display-name chains had no sentinel filtering at all).

Added a `_claim()` helper that returns `None` for sentinel values, and rewrote the three OAuth-claim extraction functions to use it:

- **`_extract_user_identifier`** — now walks the priority list explicitly and *skips* sentinel values instead of accepting them. A real claim further down the list (e.g. `email` or `upn`) can still resolve. Only falls back to `sam_dev_user` when no claim resolves at all, and logs the keys that *were* present so future regressions surface immediately.
- **`_extract_user_details`** — email/display-name chains now use `_claim()` instead of raw `.get()`, so a sentinel `email` value yields to `preferred_username`/`upn` rather than locking in.
- **`_create_user_state`** — same sentinel filter applied to the final assembled identifier.

Sentinel set is centralised as `_INVALID_CLAIM_VALUES = {"unknown", "null", "none", ""}` and the priority list is named `_USER_IDENTIFIER_CLAIM_ORDER`.

- [x] Unit tests: existing `tests/unit/gateway/http_sse/test_user_id_extraction.py` covers the priority walk. Happy to add an explicit `{"email": "Unknown", "user_id": "real-sub"}` → `"real-sub"` case inline if reviewers want it on this PR.
- [x] Manual: by inspection — behavioural change is small and local to the three helper functions.

### DIFF
--- a/src/solace_agent_mesh/shared/auth/middleware.py
+++ b/src/solace_agent_mesh/shared/auth/middleware.py
@@ -151,23 +151,26 @@ def _claim(user_info: dict, key: str) -> str | None:
     value = user_info.get(key)
     if not isinstance(value, str):
         return None
-    if value.lower() in _INVALID_CLAIM_VALUES:
+    value = value.strip()
+    if not value or value.lower() in _INVALID_CLAIM_VALUES:
         return None
     return value
 
 
 def _extract_user_identifier(user_info: dict) -> str:
     """Extract user identifier from OAuth user info."""
-    # Sentinel values like "Unknown" (used by some upstream proxies to fill
-    # missing claims) are skipped so the walk continues to the next claim.
+    # Sentinel values like "Unknown" are skipped so the walk continues.
     for key in _USER_IDENTIFIER_CLAIM_ORDER:
         candidate = _claim(user_info, key)
         if candidate:
             return candidate
 
+    rejected = sorted(k for k in _USER_IDENTIFIER_CLAIM_ORDER if k in user_info)
     log.warning(
-        "AuthMiddleware: No valid user identifier in user_info (keys=%s). Using fallback.",
+        "AuthMiddleware: No valid user identifier in user_info "
+        "(keys=%s, rejected=%s). Using fallback.",
         sorted(user_info.keys()),
+        rejected,
     )
     return "sam_dev_user"
 
@@ -197,20 +200,11 @@ async def _create_user_state(
     user_identifier: str, email_from_auth: str, display_name: str
 ) -> dict:
     """Create user state dictionary from OAuth info."""
-    final_user_id = user_identifier or email_from_auth
-    if (
-        not final_user_id
-        or not isinstance(final_user_id, str)
-        or final_user_id.lower() in _INVALID_CLAIM_VALUES
-    ):
-        final_user_id = "sam_dev_user"
-        log.warning("AuthMiddleware: Had to use fallback user ID due to invalid identifier")
-
     return {
-        "id": final_user_id,
-        "user_id": final_user_id,
-        "email": email_from_auth or final_user_id,
-        "name": display_name or final_user_id,
+        "id": user_identifier,
+        "user_id": user_identifier,
+        "email": email_from_auth or user_identifier,
+        "name": display_name or user_identifier,
         "authenticated": True,
         "auth_method": "oidc",
     }
--- a/src/solace_agent_mesh/shared/auth/middleware.py
+++ b/src/solace_agent_mesh/shared/auth/middleware.py
@@ -129,44 +129,66 @@ async def _get_user_info(
     return userinfo_response.json()
 
 
+_INVALID_CLAIM_VALUES = frozenset({"unknown", "null", "none", ""})
+
+_USER_IDENTIFIER_CLAIM_ORDER = (
+    "sub",
+    "client_id",
+    "username",
+    "oid",
+    "preferred_username",
+    "upn",
+    "unique_name",
+    "email",
+    "name",
+    "azp",
+    "user_id",
+)
+
+
+def _claim(user_info: dict, key: str) -> str | None:
+    """Return the claim at ``key``, treating sentinel strings as missing."""
+    value = user_info.get(key)
+    if not isinstance(value, str):
+        return None
+    if value.lower() in _INVALID_CLAIM_VALUES:
+        return None
+    return value
+
+
 def _extract_user_identifier(user_info: dict) -> str:
     """Extract user identifier from OAuth user info."""
-    user_identifier = (
-        user_info.get("sub")
-        or user_info.get("client_id")
-        or user_info.get("username")
-        or user_info.get("oid")
-        or user_info.get("preferred_username")
-        or user_info.get("upn")
-        or user_info.get("unique_name")
-        or user_info.get("email")
-        or user_info.get("name")
-        or user_info.get("azp")
-        or user_info.get("user_id")
+    # Sentinel values like "Unknown" (used by some upstream proxies to fill
+    # missing claims) are skipped so the walk continues to the next claim.
+    for key in _USER_IDENTIFIER_CLAIM_ORDER:
+        candidate = _claim(user_info, key)
+        if candidate:
+            return candidate
+
+    log.warning(
+        "AuthMiddleware: No valid user identifier in user_info (keys=%s). Using fallback.",
+        sorted(user_info.keys()),
     )
-
-    if user_identifier and user_identifier.lower() == "unknown":
-        log.warning("AuthMiddleware: IDP returned 'Unknown' as user identifier. Using fallback.")
-        return "sam_dev_user"
-
-    return user_identifier
+    return "sam_dev_user"
 
 
 def _extract_user_details(user_info: dict, user_identifier: str) -> tuple:
     """Extract email and display name from OAuth user info."""
     email_from_auth = (
-        user_info.get("email")
-        or user_info.get("preferred_username")
-        or user_info.get("upn")
+        _claim(user_info, "email")
+        or _claim(user_info, "preferred_username")
+        or _claim(user_info, "upn")
         or user_identifier
     )
 
+    given = _claim(user_info, "given_name") or ""
+    family = _claim(user_info, "family_name") or ""
     display_name = (
-        user_info.get("name")
-        or user_info.get("given_name", "") + " " + user_info.get("family_name", "")
-        or user_info.get("preferred_username")
+        _claim(user_info, "name")
+        or (f"{given} {family}".strip() or None)
+        or _claim(user_info, "preferred_username")
         or user_identifier
-    ).strip()
+    )
 
     return email_from_auth, display_name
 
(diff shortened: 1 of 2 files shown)


src/solace_agent_mesh/shared/auth/middleware.py
```diff
@@ -129,44 +129,66 @@ async def _get_user_info(
...
-        or user_info.get("azp")
-        or user_info.get("user_id")
+    # Sentinel values like "Unknown" (used by some upstream proxies to fill
+    # missing claims) are skipped so the walk continues to the next claim.
+    for key in _USER_IDENTIFIER_CLAIM_ORDER:
+        candidate = _claim(user_info, key)
+        if candidate:
+            return candidate
+
+    log.warning(
```
Hugo-Pare, line 169: 🟡 Nit: The fallback warning logs only `sorted(user_info.keys())`. When all expected claim keys are present but every value was filtered as a sentinel — which is the exact scenario `_INVALID_CLAIM_VALUES` was introduced to defend against — operators see a key list that's indistinguishable from "claims genuinely absent." A sanitized hint that values were filtered would make this much easier to triage.

Lightweight improvement (no PII leak; just confirms whether *any* value was rejected as a sentinel):

```suggestion
    rejected_keys = [k for k in _USER_IDENTIFIER_CLAIM_ORDER if k in user_info]
    log.warning(
        "AuthMiddleware: No valid user identifier in user_info (keys=%s, rejected_as_sentinel=%s). Using fallback.",
        sorted(user_info.keys()),
        sorted(rejected_keys),
    )
```

<details>
<summary>Why this matters</summary>

**What the gap is:** keys-only logs collapse two very different failures into one log line — "the IdP didn't send sub/email/upn at all" vs "the IdP sent `sub='Unknown'` and we filtered it." The latter is precisely what this PR is hardening against; surfacing it in the log is the difference between "fix the IdP" and "fix the proxy."

**How to fix:** the suggestion above lists the priority-ordered claim names actually present in `user_info`. No claim *values* (so no PII), but enough signal to confirm sentinel filtering fired.

**Alternative:** plumb the request through and log `request.url.path` / `request.method`. More invasive — only worth it if operators need to correlate to specific requests. Probably defer.
</details>

src/solace_agent_mesh/shared/auth/middleware.py
```diff
@@ -175,8 +197,12 @@ async def _create_user_state(
...
 ) -> dict:
     """Create user state dictionary from OAuth info."""
-    final_user_id = user_identifier or email_from_auth or "sam_dev_user"
-    if not final_user_id or final_user_id.lower() in ["unknown", "null", "none", ""]:
+    final_user_id = user_identifier or email_from_auth
+    if (
+        not final_user_id
+        or not isinstance(final_user_id, str)
+        or final_user_id.lower() in _INVALID_CLAIM_VALUES
+    ):
```
Hugo-Pare: 🟡 Nit: After this PR, `_extract_user_identifier` is guaranteed to return a non-empty string (either a real claim or the literal `"sam_dev_user"`). That makes the `or email_from_auth` branch unreachable and the sentinel re-check below it dead — `user_identifier` can no longer be any of `{"unknown", "null", "none", ""}` after the new walk.

The dead branches are confusing because they imply `_create_user_state` can be called with invalid inputs, which contradicts the invariant `_extract_user_identifier` now enforces. Simplest fix:

```suggestion
    final_user_id = user_identifier
```

<details>
<summary>Verification</summary>

`_extract_user_identifier`'s new shape returns `candidate` (a non-empty, non-sentinel string from `_claim`) or `"sam_dev_user"` — no falsy or sentinel return path remains. The only caller of `_create_user_state` is line 520, which sources `user_identifier` directly from `_extract_user_identifier`. The `final_user_id.lower() in _INVALID_CLAIM_VALUES` check can therefore never fire.

If you'd rather keep `_create_user_state` defensive against future callers, leave the guard but at minimum drop the dead `or email_from_auth` to match what's actually reachable. Tiny readability win either way.
</details>

src/solace_agent_mesh/shared/auth/middleware.py
```diff
@@ -129,44 +129,66 @@ async def _get_user_info(
     return userinfo_response.json()
 
 
+_INVALID_CLAIM_VALUES = frozenset({"unknown", "null", "none", ""})
+
+_USER_IDENTIFIER_CLAIM_ORDER = (
+    "sub",
+    "client_id",
+    "username",
```
Hugo-Pare, line 137: 🟡 Nit: The `username` claim (position 3 in this tuple) is not covered by the existing tests in `tests/unit/gateway/http_sse/test_user_id_extraction.py`. Those tests **reimplement** the priority walk locally rather than importing the real `_extract_user_identifier`, and the local list omits `username` entirely — so a regression that drops `username` from this priority order would never trip CI.

This is distinct from the broader "add sentinel-skipping coverage" point already raised by Copilot — that one is about the new sentinel logic; this one is about the test scaffolding itself not exercising the real implementation.

Suggested minimal addition (in `tests/unit/shared/auth/test_middleware.py` or alongside the existing tests):

```python
from solace_agent_mesh.shared.auth.middleware import _extract_user_identifier

def test_username_claim_preferred_over_oid():
    assert _extract_user_identifier({"username": "proxy-user", "oid": "object-id-123"}) == "proxy-user"

def test_username_sentinel_skipped_falls_to_oid():
    assert _extract_user_identifier({"username": "Unknown", "oid": "object-id-123"}) == "object-id-123"
```

Anchors the priority order *and* sentinel-skipping against the real function in one shot.

src/solace_agent_mesh/shared/auth/middleware.py
```diff
@@ -129,44 +129,66 @@ async def _get_user_info(
...
+        "AuthMiddleware: No valid user identifier in user_info (keys=%s). Using fallback.",
+        sorted(user_info.keys()),
     )
-
-    if user_identifier and user_identifier.lower() == "unknown":
-        log.warning("AuthMiddleware: IDP returned 'Unknown' as user identifier. Using fallback.")
-        return "sam_dev_user"
-
-    return user_identifier
+    return "sam_dev_user"
```
Hugo-Pare, line 175: 🟣 Pre-existing: This PR strictly improves the OIDC fallback (more claim shapes are now accepted as a real identity), but the residual case — *every* claim in `_USER_IDENTIFIER_CLAIM_ORDER` is sentinel/missing — still ends with `request.state.user = {"id": "sam_dev_user", "auth_method": "oidc", "authenticated": True, ...}`. That collides with the dev-mode identity used when `frontend_use_authorization=False` (line ~299) and is consumed by downstream scoping (`session_manager.py:54`, `dependencies.py:303`, `share.py:67/82/242/296/337`, `users.py:38`). Two distinct OIDC users whose claims all fail to resolve still collapse into the same logical principal.

The PR description targets exactly this collision, so worth flagging that the residual case remains. Not a blocker for this PR — it's a pre-existing behavior the PR narrows but doesn't eliminate.

Follow-up options (separate PR):
- Hard-fail with 401 when no real claim resolves in the auth-enabled path (`use_auth=True`), keeping `sam_dev_user` only for the explicit dev branch.
- Or tag the fallback distinctly: `auth_method="oidc_fallback"` + `id=f"sam_oidc_anon_{uuid4().hex[:12]}"` so different unresolvable tokens don't collapse and `auth_method` doesn't misreport.

The `_ROLE_FORWARD_ALLOWED_AUTH_METHODS` gate at `component.py:44` already excludes `"oidc"`, so this is **not** a privilege-escalation path — purely an identity-attribution / audit-fidelity concern.

Hugo-Pare (commented):
## PR Review

**Tier:** Standard (≈80 lines changed in a single file, but touches the OAuth identity-extraction path — auth-critical code).

3 🟡 Nit, 1 🟣 Pre-existing — see inline comments for details.

The core change is sound: bug-detection and cross-cutting agents both confirmed the dict shape is preserved, downstream consumers are unaffected, and the rewrite incidentally fixes a latent issue in the old `display_name` chain where `" "` (one-space concatenation of empty given+family names) short-circuited past the `preferred_username` fallback.

### Already raised by other reviewers — not duplicated here

- Copilot (line 156): `_claim()` doesn't normalize whitespace; `' Unknown '` / `'\nnull'` slip through. Also `name` claims with leading/trailing spaces will now flow through untrimmed.
- Copilot (line 172): no unit tests for the sentinel-skipping behavior.

The test-scaffolding finding in this review is a *separate* point from Copilot's broader test-coverage note: it's about the existing tests reimplementing the priority logic locally instead of importing the real function — so `username` and other claims could regress invisibly.

### Verification notes

Four candidate findings were retracted or downgraded during verification:

- **Retracted:** `_get_user_info` returns `None` on non-200 HTTP without logging the status (pre-existing, low signal).
- **Retracted:** `sam_user_id` claim could be `None` after token verification (pre-existing; the enterprise `trust_manager` likely enforces presence — can't confirm from this repo).
- **Retracted:** bare `except Exception` in the `sam_access_token` block (pre-existing, intentional fall-through pattern with WARNING-level logging already in place).
- **Downgraded 🔴 → 🟣:** identity-collision residue when *all* claims are sentinels (the PR strictly reduces frequency but doesn't introduce; out of scope for this change).

### Non-code observation

The commit (`dee0f68`) is missing a `Signed-off-by:` trailer. CLAUDE.md's "Before Committing" section requires DCO sign-off — the DCO check on this PR will fail until amended. Fix: `git commit --amend --signoff --no-edit && git push --force-with-lease`.

### Merge readiness

**Ready to merge with fixes** — no 🔴 findings, but the DCO trailer is required for the PR check to pass. The three 🟡 nits are non-blocking.

<details>
<summary>Run diagnostics — 15 agents, all completed</summary>

| Agent | Model | Status | Findings |
|---|---|---|---|
| haiku-gate | haiku | complete | 0 |
| fetch-pr-file-list | sonnet | complete | (1 file) |
| compliance-claude-md-root | sonnet | complete | 1 |
| bug-detection | opus | complete | 0 |
| logic-security-reliability | opus | complete | 3 |
| cross-cutting | opus | complete | 2 |
| ext-silent-failure-hunter | sonnet | complete | 4 |
| ext-pr-test-analyzer | sonnet | complete | 1 |
| verify-f3-identity-collision | sonnet | complete | downgrade |
| verify-f4-log-context | sonnet | complete | keep |
| verify-f5-dead-code | sonnet | complete | keep |
| verify-f8-http-silent | sonnet | complete | retract |
| verify-f10-sam-user-id-none | sonnet | complete | retract |
| verify-f11-bare-except | sonnet | complete | retract |
| verify-f12-username-test-gap | sonnet | complete | keep |

For cost/token breakdown, run `/pr-toolkit:summarize 1584`. Do NOT use the RUN_ID from this block — that routes sentinels, not execution artifacts.
</details>

RohanSreelesh (approved):

PR #1592, opened 2026-06-19
fix(DATAGO-141037): re-validate SSRF guard on every connection in web_request

ziyanwan:
## Summary

- The `web_request` agent tool validated URLs once via `_is_safe_url`, then fetched with `httpx.AsyncClient(follow_redirects=True)` — redirect hops were not re-checked, so an attacker-controlled public host could `302` the agent into loopback / RFC1918 / cloud-metadata addresses (CWE-918, redirect-revalidation bypass). The same path also had a resolve-then-connect TOCTOU window.
- Move IP-block helpers into a shared `common/utils/ssrf` module and install `SSRFSafeTransport` on the `AsyncClient`. Validation now happens inside the transport, which httpx calls for every request including each redirect hop, closing both gaps.
- The webhook scheduler's `notification_service` already used the same transport pattern locally; both call sites now share one implementation. Existing private aliases (`_check_ip_blocked`, `_SSRFSafeTransport`) are preserved as re-exports so existing tests stay green.

## Why a transport, not `follow_redirects=False`

Disabling redirects would close the disclosed bypass but break a lot of normal LLM-driven web fetches (HTTPS upgrades, www → bare-domain, paywalls, DOI/archive.org redirects, CDN canonicalization). It also wouldn't fix the resolve-then-connect TOCTOU. A connection-time check fixes both with one mechanism, and matches what the webhook scheduler already does.

## Test plan

- [x] New unit tests in `tests/unit/common/test_ssrf.py` cover the IP blocklist and `SSRFSafeTransport` against loopback, the cloud-metadata endpoint, and a DNS-rebinding simulation (15 tests, all pass locally).
- [x] New regression test `tests/unit/agent/tools/test_web_request_ssrf.py` reproduces the disclosed attack shape: a real localhost listener returns `302` → loopback, `_is_safe_url` is patched to `True` to simulate "attacker host appears public", and `web_request` is asserted to return the safety error with no artifact produced. Also covers direct loopback, `169.254.169.254`, and the `allow_loopback` testing escape hatch.
- [x] Pre-existing `notification_service` SSRF tests remain unchanged — the helpers they import are still resolvable through `notification_service` via re-exports.
- [ ] Reviewers: please run `hatch test -m security` to confirm the full suite stays green in a properly-provisioned env (local venv was missing `apscheduler` and a current `solace_ai_connector`, both unrelated to this change).

## Notes for reviewers

- Latent bug fixed in passing: `notification_service.py` previously compared `url.scheme == b"https"` (httpx URL scheme is a `str`, not bytes). The shared transport uses the correct `str` comparison.
- `web_request`'s pre-flight `_is_safe_url` is preserved for fast rejection / log clarity; the transport is now the authoritative guard.
- The `allow_loopback` test-only knob still disables both pre-flight and transport, matching prior semantics. Tightening that to "private but not metadata" is a possible follow-up.
- Out of scope: adding a scheme allowlist to `web_request` (the scheduler does this; the agent tool currently does not). Worth a follow-up for parity.

### DIFF
--- a/src/solace_agent_mesh/agent/tools/web_tools.py
+++ b/src/solace_agent_mesh/agent/tools/web_tools.py
@@ -140,6 +140,13 @@ async def web_request(
                 client_kwargs: Dict[str, Any] = {
                     "timeout": 30.0,
                     "follow_redirects": True,
+                    # Ignore HTTP_PROXY/HTTPS_PROXY/NO_PROXY env. With those
+                    # set, httpx would connect to the proxy host instead of
+                    # the origin, and SSRFSafeTransport only validates the
+                    # origin's DNS — leaving an unvalidated proxy as a
+                    # bypass. If proxy support is needed later, plumb it
+                    # through tool_config with explicit destination checks.
+                    "trust_env": False,
                 }
                 if not allow_loopback:
                     client_kwargs["transport"] = SSRFSafeTransport()
--- a/src/solace_agent_mesh/agent/tools/web_tools.py
+++ b/src/solace_agent_mesh/agent/tools/web_tools.py
@@ -130,14 +130,19 @@ async def web_request(
         # Retry logic with exponential backoff. The SSRF-safe transport
         # re-validates each redirect hop at connect time, closing the
         # follow_redirects bypass and the resolve-then-connect TOCTOU window.
-        client_kwargs: Dict[str, Any] = {"timeout": 30.0, "follow_redirects": True}
-        if not allow_loopback:
-            client_kwargs["transport"] = SSRFSafeTransport()
-
+        # A fresh transport is built per attempt because AsyncClient closes
+        # its transport on context-manager exit, and a closed transport's
+        # connection pool cannot be reused on the next retry.
         last_error = None
         response = None
         for attempt in range(1, max_retries + 1):
             try:
+                client_kwargs: Dict[str, Any] = {
+                    "timeout": 30.0,
+                    "follow_redirects": True,
+                }
+                if not allow_loopback:
+                    client_kwargs["transport"] = SSRFSafeTransport()
                 async with httpx.AsyncClient(**client_kwargs) as client:
                     log.info(
                         f"{log_identifier} Attempt {attempt}/{max_retries}: Making {method} request to {url}"
--- a/src/solace_agent_mesh/agent/tools/web_tools.py
+++ b/src/solace_agent_mesh/agent/tools/web_tools.py
@@ -33,11 +33,14 @@ CATEGORY_DESCRIPTION = "Access the web to find information to complete user requ
 DEFAULT_MAX_RESPONSE_SIZE = 10 * 1024 * 1024  # 10 MB default
 ABSOLUTE_MAX_RESPONSE_SIZE = 50 * 1024 * 1024  # 50 MB hard cap (cannot be exceeded even via config)
 
-def _is_safe_url(url: str) -> bool:
+def _is_safe_url(url: str, *, allow_loopback: bool = False) -> bool:
     """
     Pre-flight check that rejects obviously unsafe URLs before opening a
     connection. The authoritative SSRF guard is :class:`SSRFSafeTransport`,
     which re-validates every hop (including HTTP redirects) at connect time.
+
+    ``allow_loopback`` exempts only loopback addresses; RFC1918 and cloud
+    metadata stay blocked.
     """
     try:
         hostname = urlparse(url).hostname
@@ -47,7 +50,7 @@ def _is_safe_url(url: str) -> bool:
 
         try:
             for *_, sockaddr in socket.getaddrinfo(hostname, None):
-                check_ip_blocked(sockaddr[0])
+                check_ip_blocked(sockaddr[0], allow_loopback=allow_loopback)
         except socket.gaierror:
             log.warning(f"Could not resolve hostname: {hostname}")
             return False
@@ -108,7 +111,7 @@ async def web_request(
             max_response_size = min(int(configured_size), ABSOLUTE_MAX_RESPONSE_SIZE)
             log.debug(f"{log_identifier} Using configured max_response_size: {max_response_size} bytes")
 
-    if not allow_loopback and not _is_safe_url(url):
+    if not _is_safe_url(url, allow_loopback=allow_loopback):
         log.error(f"{log_identifier} URL is not safe to request: {url}")
         return ToolResult.error("URL is not safe to request.")
 
@@ -147,9 +150,15 @@ async def web_request(
                     # bypass. If proxy support is needed later, plumb it
                     # through tool_config with explicit destination checks.
                     "trust_env": False,
+                    # Always install the transport. allow_loopback narrows
+                    # what it permits (loopback only); RFC1918, cloud
+                    # metadata, IPv6 ULA, and redirect re-validation stay
+                    # active even in dev mode. Dropping the transport
+                    # entirely — which an earlier revision of this PR did
+                    # for allow_loopback=True — silently disabled the rest
+                    # of the SSRF guarantee.
+                    "transport": SSRFSafeTransport(allow_loopback=allow_loopback),
                 }
-                if not allow_loopback:
-                    client_kwargs["transport"] = SSRFSafeTransport()
                 async with httpx.AsyncClient(**client_kwargs) as client:
                     log.info(
                         f"{log_identifier} Attempt {attempt}/{max_retries}: Making {method} request to {url}"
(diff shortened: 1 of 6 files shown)


ziyanwan:
@claude review once

src/solace_agent_mesh/gateway/http_sse/services/scheduler/notification_service.py
```diff
@@ -13,24 +12,16 @@
 import httpx
 from sqlalchemy.orm import Session as DBSession
 
+from .....common.utils.ssrf import (
+    SSRFSafeTransport as _SSRFSafeTransport,
```
gregmeldrum, line 16: 🟡 Nit: The webhook client that consumes this shared transport (`__init__`, lines 93–96) does **not** set `trust_env=False`, so the two consumers of the new shared module now have asymmetric SSRF postures.

`web_tools.py` deliberately added `trust_env=False` (line 152) with a comment explaining why: with `HTTP_PROXY`/`HTTPS_PROXY` set, httpx connects to the *proxy* host instead of the origin, and `SSRFSafeTransport` only validates the **origin's** DNS — leaving an unvalidated proxy as a bypass. The webhook client here:

```python
self.http_client = httpx.AsyncClient(
    timeout=30.0,
    transport=_SSRFSafeTransport(),
)
```

still defaults to `trust_env=True`, so in any gateway environment with proxy env vars set, an attacker-controlled webhook URL can be routed through the proxy without IP validation — the exact bypass class this PR closes for `web_tools`.

Suggest adding `trust_env=False` here to keep both call sites consistent:

```python
self.http_client = httpx.AsyncClient(
    timeout=30.0,
    trust_env=False,
    transport=_SSRFSafeTransport(),
)
```

<details>
<summary>Extended reasoning…</summary>

**What it is:** The PR's stated goal is "one implementation" shared by both SSRF consumers, but only `web_tools` got the proxy-env hardening. The webhook delivery client retains the default `trust_env=True`.

**Why existing code doesn't prevent it:** `SSRFSafeTransport.handle_async_request` validates `request.url.host`. When httpx is configured to use a proxy (via env), the TCP connection target is the proxy, not the validated origin — the transport never sees or validates the proxy endpoint.

**Impact:** Only exploitable when proxy env vars are present in the gateway's environment, so this is a 🟡 rather than 🔴 — but it's the identical bypass the PR explicitly fixes one file over, and webhook URLs are user-influenced via scheduled-task config.

**How to fix:** Add `trust_env=False` to the `AsyncClient(...)` here, mirroring `web_tools.py:152`.
</details>

src/solace_agent_mesh/agent/tools/web_tools.py
```diff
@@ -33,27 +33,29 @@
...
-        hostname = parsed_url.hostname
+        hostname = urlparse(url).hostname
         if not hostname:
             log.warning(f"URL has no hostname: {url}")
             return False
 
         try:
-            ip_str = socket.gethostbyname(hostname)
-            ip = ipaddress.ip_address(ip_str)
+            for *_, sockaddr in socket.getaddrinfo(hostname, None):
```
gregmeldrum, line 52: 🟡 Nit: This pre-flight `socket.getaddrinfo` runs synchronously on the event loop, but `web_request` is a coroutine. You deliberately wrapped the transport's identical lookup in `asyncio.to_thread` (`ssrf.py:116`, "so a slow resolver doesn't stall the event loop") — the same hazard applies here, just left unwrapped.

A slow or hanging DNS resolver during the pre-flight check stalls every other coroutine in the agent process for the duration of the lookup. Not a crash, just degraded concurrency, and the prior code (`socket.gethostbyname`) was already blocking — so this is a carried-forward pattern rather than newly introduced. But since the line was rewritten here and the transport already shows the fix, it's worth aligning.

Make `_is_safe_url` async and `await asyncio.to_thread(socket.getaddrinfo, hostname, None)`, then `await _is_safe_url(...)` at the call site (line 114).

gregmeldrum (commented):
## PR Review — SSRF redirect re-validation (DATAGO-141037)

**Depth: Deep** — security/SSRF path, 6 files / ~502 additions, touches a shared module with two consumers.

This is a well-constructed fix. Moving the IP-block predicate + `SSRFSafeTransport` into `common/utils/ssrf` and validating inside the transport genuinely closes the disclosed redirect-revalidation bypass and the resolve-then-connect TOCTOU — every hop (including each 302) now passes through `handle_async_request`. The `check_ip_blocked` logic is strong: the `is_private | is_reserved | is_loopback | is_link_local | is_multicast | is_unspecified` predicate plus the CIDR backstop is strictly broader than the old CIDR-only list, the IPv4-mapped IPv6 unwrap is correct, `BlockedIPError` (a `ValueError` subclass) is caught separately and explicitly not retried, and the `trust_env=False` + per-attempt-transport decisions are sound and well-commented. Re-exports keep `notification_service` and its tests green.

**Findings: 2 🟡 Nit — see inline comments.** Both are non-blocking.

- 🟡 `notification_service.py` webhook client omits `trust_env=False` (the proxy-env bypass `web_tools` explicitly guards against one file over) — asymmetric SSRF posture between the two consumers of the new shared module.
- 🟡 `_is_safe_url` pre-flight `getaddrinfo` runs blocking on the event loop, while the transport's identical lookup is correctly wrapped in `asyncio.to_thread`.

### Already raised by Copilot (not re-posted)
Copilot filed 11 threads; most are resolved/addressed in code (transport reuse, gaierror→ConnectError, IPv4-mapped IPv6, blocking transport DNS, async test markers, ruff E241). Two worth keeping in mind:
- **DNS-rebinding TOCTOU residual** — the transport validates the IPs from its own `getaddrinfo`, but httpx/anyio re-resolves independently at connect time, so a narrow rebinding window remains. This is an acknowledged limitation in the PR description and is **not worse** than disclosed; closing it fully needs a custom network backend that couples the validated sockaddr to the actual connect. Fine as a follow-up.
- The unresolved `trust_env`/proxy thread on `web_tools` is **now addressed in code** (line 152) — the same fix is just missing from `notification_service` (my inline finding above).

### Non-blocking notes (not posted inline — pre-existing, outside changed hunks)
- 🟣 **Silent truncation in `web_request`** (`web_tools.py:191–217`): when the streamed body exceeds `max_response_size`, the loop logs a warning, `break`s, and returns `ToolResult.ok("Successfully fetched…")` with the partial body — whereas the Content-Length pre-check returns a hard error. The model/user gets a truncated document reported as a complete success. Pre-existing (this PR doesn't touch the streaming logic) but worth a follow-up; consider a `truncated: true` flag or a hard error for consistency.
- **Test coverage gap (security-relevant):** neither new test exercises the production `SSRFSafeTransport.handle_async_request` on a *true* public→loopback redirect hop. `test_web_request_ssrf.py`'s redirect test uses a loopback **initial** URL (blocked on the first connect, 302 listener never reached — Copilot noted the docstring is misleading), and `test_ssrf.py`'s redirect test overrides `handle_async_request` with a reimplementation. A regression that broke per-hop validation in the real transport would pass the suite green. Recommend one test: initial host mocked to resolve public, 302 → loopback, driven through the unmodified production transport. Also untested: `trust_env=False` and the `BlockedIPError`-doesn't-retry branch in `web_request`.

### CI status
SonarQube reports **4 security hotspots** and FOSSA flags 6 vulnerability issues (3 Medium / 3 Unknown) vs base — worth confirming whether any are attributable to this change (likely the `getaddrinfo`/SSRF hotspots are expected for security code, but verify before merge).

### Verification
Both inline findings were verified by reading the actual files in a local checkout of the PR head. Test analyzer ran `test_ssrf.py` (39 pass); `test_web_request_ssrf.py` was uncollectable locally due to a missing `solace_ai_connector` dependency (environmental, unrelated to the change — matches the PR's own note).

**Merge readiness: Ready to merge with fixes** — no 🔴 findings. The `trust_env` asymmetry is a small, worthwhile fix for parity; everything else is non-blocking.

<details><summary>Run diagnostics — 5 agents, all succeeded</summary>

| Agent | Model | Status | Findings |
|---|---|---|---|
| bug-detection | opus | complete | 1 |
| logic-security-reliability | opus | complete | 3 |
| cross-cutting | opus | complete | 1 |
| ext-silent-failure-hunter | opus | complete | 7 |
| ext-pr-test-analyzer | opus | complete | 5 |

Findings were consolidated and deduplicated against Copilot's existing review; 2 net-new actionable items posted inline, the rest summarized above. For cost/token breakdown, run `/pr-toolkit:summarize 1592`.
</details>

gregmeldrum:
I don't think those nits are blocking. I'm approving the PR.

gregmeldrum (approved):
LG

PR #1595, opened 2026-06-19
fix(security): upgrade turbo-stream to 3.0.0 (CVE-2026-3407)

ziyanwan:
## Summary

- Adds `turbo-stream: 3.0.0` to the npm `overrides` in `config_portal/frontend/package.json`
- Upgrades the resolved version in `package-lock.json` from 2.4.1 → 3.0.0
- `turbo-stream` is a transitive dependency of `@remix-run/node` and `@remix-run/react`

**CVE:** CVE-2026-3407
**Affected package:** `turbo-stream` 2.4.1
**Fix:** Upgrade to 3.0.0

## Test plan

- [ ] `npm ci` in `config_portal/frontend/` installs without errors
- [ ] `node -e "require('./node_modules/turbo-stream/package.json').version"` prints `3.0.0`
- [ ] `npm run build` succeeds

### DIFF
--- a/config_portal/frontend/package-lock.json
+++ b/config_portal/frontend/package-lock.json
@@ -2158,12 +2158,14 @@
       "version": "15.7.15",
       "resolved": "https://registry.npmjs.org/@types/prop-types/-/prop-types-15.7.15.tgz",
       "integrity": "sha512-F6bEyamV9jKGAFBEmlQnesRPGOQqS2+Uwi0Em15xenOxHaf2hv6L8YCVn3rPdPJOiJfPiCnLIRyvwVaqMY3MIw==",
+      "dev": true,
       "license": "MIT"
     },
     "node_modules/@types/react": {
       "version": "18.3.28",
       "resolved": "https://registry.npmjs.org/@types/react/-/react-18.3.28.tgz",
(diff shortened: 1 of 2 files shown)


ziyanwan:
Thanks for the flag — this was a fair concern worth validating explicitly.

**Compatibility validation: Remix 2.17.5 + turbo-stream 3.0.0**

The breaking API changes in turbo-stream 3.x are:
- `decode` changed stream type from `ReadableStream<Uint8Array>` → `ReadableStream<string>`, and return type from `{done: Promise<void>, value: unknown}` → `Promise<T>` directly
- `encode` changed output from `ReadableStream<Uint8Array>` → `ReadableStream<string>`; removed `postPlugins` option; added `redactErrors`

These are only exercised by Remix's **deferred SSR streaming** path — the mechanism that pipes server loader data to the client mid-render.

**This app runs in SPA mode (`ssr: false`).**  `vite.config.ts` sets `ssr: false`, so Remix generates a static `index.html` with no Node server runtime. There are no server-side loaders or actions that could invoke `encode`/`decode` at runtime. The code is bundled but the call paths are never reached.

Validation:
1. `npm run build` passes cleanly ✓
2. `node_modules/turbo-stream` resolves to 3.0.0 ✓
3. No SSR runtime means turbo-stream's streaming API is dead code ✓

**Re: upgrading @remix-run***  — `@remix-run` 2.17.5 is the current latest and still pins turbo-stream 2.4.1; there's no upstream release to upgrade to. The clean long-term path is migrating to React Router v7, which drops turbo-stream entirely. Added a `//turbo-stream` comment in `overrides` to document this so the next person doesn't have to reverse-engineer the rationale.

PR #1622, opened 2026-08-10
fix(DATAGO-145440): bump GitPython to 3.1.59

moodiRealist:

    Fixes DATAGO-145440: 18 High/Medium GitHub advisories (mostly GHSA-only, no assigned CVEs) against GitPython 3.1.50, flagged by Prisma. All are variations of the same underlying issue — git command/option injection via subprocess calls — where each patch release closed one bypass and a subsequent advisory found the next one. Full GHSA list is in the Jira ticket's attached CSV.

    GitPython is a direct, exact-pinned dependency (`GitPython==3.1.50`) in `pyproject.toml`. Bumped the pin to `3.1.59`, the current latest release on PyPI with no open advisories, and refreshed `uv.lock` via `uv lock`. No other dependencies were affected.

- [x] Manual testing: `uv lock` resolved cleanly (`Updated gitpython v3.1.50 -> v3.1.59`); confirmed `import git` works at the new version.
- [x] Unit tests: ran the existing suites covering the only in-repo GitPython usage (`config_portal/backend/plugin_catalog/scraper.py`) — `tests/unit/config_portal/backend/plugin_catalog/test_registry_readd.py` and `tests/unit/cli/commands/plugin_cmd/test_catalog_cmd.py`, 25/25 passed.
- [ ] Integration tests: not applicable.
- [x] Known limitations: none — the in-repo usage (`git.Repo`, `clone_from`, `GitCommandError`) is a stable, long-standing API untouched by these security patches.

    9 patch versions in one jump looks large, but every intermediate release (3.1.51-3.1.59) was a security-only patch closing a specific injection bypass, not a feature release. Diff is limited to `pyproject.toml` and `uv.lock`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ dependencies = [
     "jsonpath-ng==1.7.0",
     "pydub==0.25.1",
     "toml==0.10.2",
-    "GitPython==3.1.50",
+    "GitPython==3.1.59",  # Security fixes (see DATAGO-145440 for the full GHSA list)
     "Flask==3.1.3",
     "flask-cors==6.0.1",
     "werkzeug==3.1.6",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ dependencies = [
     "jsonpath-ng==1.7.0",
     "pydub==0.25.1",
     "toml==0.10.2",
-    "GitPython==3.1.59",  # Security fixes (see DATAGO-145440 for the full GHSA list)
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -57,7 +57,7 @@ dependencies = [
     "jsonpath-ng==1.7.0",
     "pydub==0.25.1",
     "toml==0.10.2",
-    "GitPython==3.1.50",
+    "GitPython==3.1.59",  # Security fixes (see DATAGO-145440 for the full GHSA list)
```
rudraneel-chakraborty: as its open source, should we keep jira reference?

rudraneel-chakraborty (approved):
see the comment please

pyproject.toml
moodiRealist: Good catch, thanks. Swapped the internal Jira reference for the actual public GHSA IDs, matching the convention other multi-CVE pins in this file already use (e.g. the `pillow` pin above). Fixed in 782d2da63.

moodiRealist: Fair — dropped the trailing comment entirely instead. The GHSA list lives in the PR description; the pin itself doesn't need it. Fixed in e598290ff.

PR #1653, opened 2026-09-09
feat(evaluation): Export results to the EvalPort open format

### LINKED ISSUE #1635: Bridge evaluation/ test cases and results.json to the EvalPort open format
EvalPort (https://github.com/adhabnr-ux/evalport) is an open JSON-Schema spec for portable LLM eval documents — test suites, test cases, and result sets — meant so eval data isn't locked to one framework's format.

I read through `evaluation/shared/test_case_loader.py` and `evaluation/evaluator.py`, and the `evaluation/` framework here is honestly one of the more complete homegrown eval setups I've seen in a multi-agent repo — Pydantic-validated `TestCase` (`test_case_id`, `query`, `target_agent`, `category`, `artifacts`, and an `Evaluation` block with `expected_tools` / `expected_response` / `criterion`), three real evaluation strategies (`ToolMatchEvaluator`, `ResponseMatchEvaluator` using ROUGE, `LLMEvaluator` with a 0.0-1.0 rubric), and aggregated `EvaluationResult` → `TestCaseResults` → `ModelResults` written out as `results.json`, plus a full HTML report generator. This isn't a stub — it's a real system, which is exactly why it's worth being able to export from.

That schema maps onto EvalPort's `TestCase`/`ResultSet` (`spec/schemas/testcase.json`, `spec/schemas/resultset.json`) almost field-for-field:

```python
# SAM TestCase -> EvalPort TestCase
def sam_test_case_to_evalport(tc: dict) -> dict:
    return {
        "id": tc["test_case_id"],
        "input": tc["query"],
        "expected_output": tc["evaluation"]["expected_response"] or None,
        "expected_tools": tc["evaluation"]["expected_tools"] or None,
        "metadata": {"target_agent": tc["target_agent"], "category": tc["category"]},
        "graders": ["tool_match", "response_match", "llm_judge"],  # matches your 3 evaluators
    }

# SAM EvaluationResult (per run) -> EvalPort grader_results
def sam_result_to_grader_results(r) -> list[dict]:
    out = []
    if r.tool_match_score is not None:
        out.append({"grader_id": "tool_match", "type": "custom",
                     "score": r.tool_match_score, "passed": r.tool_match_score >= 1.0})
    if r.response_match_score is not None:
        out.append({"grader_id": "response_match", "type": "semantic_similarity",
                     "score": r.response_match_score, "passed": r.response_match_score >= 0.5})
    if r.llm_eval_score is not None:
        out.append({"grader_id": "llm_judge", "type": "llm_judge",
                     "score": r.llm_eval_score, "passed": r.llm_eval_score >= 0.6,
                     "reason": r.llm_eval_reasoning})
    return out
```

`ModelResults.model_name` maps to `ResultSet.provider.model`, and `TestCaseResults` per test case maps to one `Result` with those `grader_results`. This would let `sam eval` output be diffed against runs from other EvalPort-aware tools (there are existing converters for CrewAI, DeepEval, Inspect AI and the OpenAI Agents SDK in `sdk/python/openeval/` for reference — same `from_<x>()` / `<x>_result_to_result_set()` shape I'm sketching above), without touching your existing HTML report pipeline at all — this would just be an additional export alongside `results.json`.

Spec: https://github.com/adhabnr-ux/evalport/blob/main/SPEC.md
Schemas referenced above: https://github.com/adhabnr-ux/evalport/tree/main/spec/schemas

No pressure at all — your evaluator is already solid on its own. Happy to sketch a PR for the converter if it'd be useful, or happy to just leave this here. Thanks for solace-agent-mesh.

plthiyagu:
## What

Bridges the `evaluation/` framework to [EvalPort](https://github.com/adhabnr-ux/evalport), an open JSON-Schema specification for portable LLM evaluation documents, so SAM suites and results can be consumed by other evaluation tools without bespoke glue code.

Closes #1635

## How

- New `evaluation/evalport_bridge.py`:
  - Maps each test case to an EvalPort TestCase document (query -> `input`, expected_response -> `expected_output`, expected_tools, wait_time -> `timeout_ms`, category -> `tags`, SAM-specific fields under `sam.*` metadata keys).
  - Maps each model's `results.json` to an EvalPort ResultSet — one result per run, with the run number carried as the spec's `attempt` field.
  - The three evaluators map to framework-specific grader types (`tool_match`, `response_match`, `llm_eval`) declaring `params.handler`, so consumers that cannot execute them skip gracefully per the spec.
  - SAM scores are continuous with no pass/fail notion, while EvalPort requires a boolean per grader result: the bridge derives it from a configurable `--pass-threshold` (default 0.5) and records the threshold in ResultSet metadata.
- CLI: `sam eval <config> --export-evalport [--pass-threshold 0.5]` exports after a run; `python -m evaluation.evalport_bridge <config>` converts an existing results directory without re-running anything.
- Docs: "Exporting to EvalPort" section in `evaluation/README.md`.

## Testing

- `tests/unit/evaluation/test_evalport_bridge.py` (new) + extended `tests/unit/cli/commands/test_eval_cmd.py`: 36 tests, all passing.
- Both generated document types validate `True` against the official `evalport-sdk` validators (`validate_suite`, `validate_result_set`).
- New files are ruff-clean.

### DIFF
--- a/cli/commands/eval_cmd.py
+++ b/cli/commands/eval_cmd.py
@@ -17,7 +17,22 @@ from cli.utils import error_exit
     is_flag=True,
     help="Enable verbose output.",
 )
-def eval_cmd(test_suite_config_path, verbose):
+@click.option(
+    "--export-evalport",
+    is_flag=True,
+    help="After the run, export the test cases and results to the EvalPort "
+    "open format (https://github.com/adhabnr-ux/evalport) under "
+    "<results dir>/evalport.",
+)
+@click.option(
+    "--pass-threshold",
+    type=click.FloatRange(0.0, 1.0),
+    default=0.5,
+    show_default=True,
+    help="Score at or above which an exported grader result counts as "
+    "passed (only used with --export-evalport).",
+)
+def eval_cmd(test_suite_config_path, verbose, export_evalport, pass_threshold):
     """
     Run an evaluation suite using a specified configuration file. Such as path/to/file.yaml.
 
@@ -43,3 +58,15 @@ def eval_cmd(test_suite_config_path, verbose):
         click.echo(click.style("Evaluation completed successfully.", fg="green"))
     except Exception as e:
         error_exit(f"An error occurred during evaluation: {e}")
+
+    if export_evalport:
+        from evaluation.evalport_bridge import export_results
+
+        try:
+            written = export_results(
+                test_suite_config_path, pass_threshold=pass_threshold
+            )
+            for path in written:
+                click.echo(click.style(f"Wrote {path}", fg="green"))
+        except Exception as e:
+            error_exit(f"An error occurred during the EvalPort export: {e}")
--- a/evaluation/README.md
+++ b/evaluation/README.md
@@ -64,3 +64,34 @@ For evaluations that use an LLM to judge the response, the following variables a
 export LLM_SERVICE_ENDPOINT=<enter the endpoint for the LLM service>
 export LLM_SERVICE_API_KEY=<enter the API key for the LLM service>
 ```
+
+## Exporting to EvalPort
+
+The evaluation results can be exported to [EvalPort](https://github.com/adhabnr-ux/evalport),
+an open JSON-Schema specification for portable LLM evaluation documents, so
+suites and results can be consumed by other evaluation tools without bespoke
+glue code.
+
+Export as part of a run:
+
+```bash
+sam eval <test_suite_config.json> --export-evalport
+```
+
+Or convert the results of a previous run without re-executing anything:
+
+```bash
+python -m evaluation.evalport_bridge <test_suite_config.json>
+```
+
+This writes one `<suite>.suite.json` (the test cases plus grader
+definitions) and one `<model>.resultset.json` per evaluated model into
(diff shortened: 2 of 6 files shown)


plthiyagu:
Addressed all four Copilot findings in `225b69f53`:

1. **Suite filename traversal**: `results_dir_name` is now validated as a single safe path component (`_safe_filename_component`) before it names the default results directory, and the suite file takes its name from the results directory on disk.
2. **Model filename traversal**: each ResultSet file is named after the model directory found by the glob (`results_file.parent.name`), never the `model_name` field inside `results.json`, so crafted JSON content cannot choose where the exporter writes. Test included with a crafted `model_name: \"../../escape\"`.
3. **`pass_threshold` range**: `export_results` validates `[0, 1]`, covering the `python -m evaluation.evalport_bridge` path (the `sam eval` flag already used `click.FloatRange(0, 1)`).
4. **Schema-validation claim**: now reproducible in-repo — `tests/unit/evaluation/test_evalport_schema_validation.py` validates the generated suite and ResultSet documents with the official `evalport-sdk` validators, and skips cleanly when the package is not installed (it is deliberately not added as a project dependency).

Suite: 42 passed (bridge + CLI + schema validation with `evalport-sdk` installed), ruff clean.

Issue #1486, opened 2026-04-30
feat: expose root_path / sub-path prefix for Web UI when deployed behind a reverse proxy


jessemenning:
## Problem

When deploying the SAM Web UI behind a reverse proxy (e.g. AWS ALB, nginx) at a sub-path such as `/agent-mesh/`, the UI is unreachable because:

1. **uvicorn** is started without a `root_path` parameter in `gateway/http_sse/component.py`, so FastAPI is unaware it is mounted at a sub-path.
2. **The bundled SPA frontend** has hardcoded absolute asset paths (e.g. `src="/assets/main-B0vPuUCC.js"`) with no `<base href>` tag or runtime injection mechanism.

As a result, a browser request to `https://myhost/agent-mesh/` loads the initial HTML, but immediately requests `/assets/main-*.js` from the root — which the reverse proxy routes elsewhere — causing a blank/broken UI.

## Expected behaviour

An operator should be able to configure a URL prefix so the Web UI is accessible at `https://myhost/<prefix>/` without dedicating a separate port or hostname to SAM.

### 1 — Add `fastapi_root_path` to the app schema (`gateway/http_sse/app.py`)

```python
{
    "name": "fastapi_root_path",
    "required": False,
    "default": "",
    "type": "string",
    "description": "URL path prefix when SAM is mounted behind a reverse proxy (e.g. '/agent-mesh'). Sets uvicorn root_path so FastAPI generates correct redirect/OpenAPI URLs.",
},
```

### 2 — Pass `root_path` to uvicorn (`gateway/http_sse/component.py`)

```python
config = uvicorn.Config(
    app=self.fastapi_app,
    host=self.fastapi_host,
    port=port,
    root_path=self.fastapi_root_path,   # <-- add this
    log_level="warning",
    lifespan="on",
    ...
)
```

### 3 — Rebuild the SPA frontend with a configurable `base`

The Vite build should read `VITE_BASE_PATH` (or equivalent) at build time, and the Docker image build should inject it. Alternatively, inject a `<base href>` tag at container startup via an entrypoint script so the base path is configurable without a full frontend rebuild.

## Workarounds (current)

- Deploy SAM on a dedicated port (e.g. `:8000`) with its own ALB listener — works but requires a separate port, SG rule, and Cognito callback URL.
- Patch `component.py` locally in a forked image build — invasive and ties the deployment to a non-upstream binary.

## Environment

[...]

PR #1654, opened 2026-09-09
feat(webui): expose fastapi_root_path for reverse proxy sub-paths

plthiyagu:
## What

Lets an operator mount the Web UI gateway behind a reverse proxy at a sub-path (e.g. `https://myhost/agent-mesh/`) by exposing a `fastapi_root_path` app config, passed to uvicorn as `root_path`.

Addresses the backend half of #1486 (items 1 and 2 of the proposal there).

## How

- `gateway/http_sse/app.py`: new `fastapi_root_path` schema entry (optional, default `""`)
- `gateway/http_sse/component.py`: read the config and pass `root_path=self.fastapi_root_path` into `uvicorn.Config` — Starlette propagates the ASGI `root_path`, so redirects, `/docs`, `/openapi.json`, `request.base_url` and share-link URLs honor the prefix
- Config surfaces: `templates/webui.yaml`, `examples/gateways/webui_gateway_example.yaml` and `preset/env-template` gain `FASTAPI_ROOT_PATH` (unset resolves to `""`, so existing deployments are unaffected)

## Scope note

Item 3 of the issue (the SPA's hardcoded `/assets/...` base path) needs a frontend build change (`VITE` base or an injected `<base href>`), so it's deliberately left for a follow-up. With `root_path` in place the API and docs are fully functional behind the proxy, which also unblocks API-only deployments.

## Testing

- New `tests/unit/gateway/http_sse/test_fastapi_root_path.py`: schema default, component runtime fallback, and uvicorn kwargs for both a configured prefix and the default — mirrors the existing `test_uvicorn_config.py` / `test_sse_max_queue_size_defaults.py` patterns
- Full `tests/unit/gateway/http_sse` suite: 1353 passed, 23 skipped
- Ruff: zero new findings on touched files

### DIFF
--- a/examples/gateways/webui_gateway_example.yaml
+++ b/examples/gateways/webui_gateway_example.yaml
@@ -57,7 +57,7 @@ apps:
       gateway_id: ${WEBUI_GATEWAY_ID} # Optional: Unique ID for this instance. If omitted, one will be generated.
       fastapi_host: ${FASTAPI_HOST, localhost} # Host for the FastAPI server
       fastapi_port: ${FASTAPI_PORT, 8000} # Port for the FastAPI server
-      fastapi_root_path: ${FASTAPI_ROOT_PATH} # URL prefix behind a reverse proxy sub-path, e.g. "/agent-mesh"
+      fastapi_root_path: ${FASTAPI_ROOT_PATH, ""} # URL prefix behind a reverse proxy sub-path, e.g. "/agent-mesh"
       cors_allowed_origins: # List of allowed origins for CORS
         - "http://localhost:3000" # Example: Allow local React dev server
         - "http://127.0.0.1:3000"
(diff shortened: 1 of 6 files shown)


plthiyagu:
Addressed both Copilot findings in `8d3793496`, with one correction to their premise:

The connector's `expandvars_with_defaults` never leaves a literal `${FASTAPI_ROOT_PATH}` placeholder — an unset variable without a default substitutes an empty string (`os.environ.get(var, "")`). The real subtlety is one step later: the substitution happens on raw YAML text, so an unset variable leaves `fastapi_root_path:` with a blank value, which YAML parses as **None**, not `""`.

Fixed at both layers:
- `templates/webui.yaml` and the gateway example now carry an explicit default: `${FASTAPI_ROOT_PATH, ""}` — the quotes survive into the YAML text, so it parses as an empty string, and the intent is visible to readers
- The component read is now `self.get_config("fastapi_root_path", "") or ""`, so even a config that carries the key with a None value hands uvicorn a string; the runtime-fallback test pins this

Verified both resolutions end to end through `expandvars_with_defaults` + `yaml.safe_load`: unset → `''`, set → `'/agent-mesh'`. Tests: 5/5, ruff: zero new findings.

